PR #26, opened 2021-03-22
Print output to cli once the task is finished

poudels14:


pkg/print/table.go
```diff
@@ -76,3 +81,95 @@ func (t Table) runs(runs []api.Run) {
...
+		return strconv.Itoa(t)
+	case float64:
+		return strconv.FormatFloat(t, 'f', -1, 64)
+	case string:
+		return t
+	case nil:
+		return ""
+	default:
+		v, _ := json.Marshal(t)
+		return string(v)
```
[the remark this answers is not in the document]
poudels14: done

pkg/print/json.go
```diff
@@ -40,3 +41,11 @@ func (j *JSON) runs(runs []api.Run) {
 func (j *JSON) run(run api.Run) {
 	j.enc.Encode(run)
 }
+
+func (j *JSON) outputs(outputs api.Outputs) {
+	for key, values := range outputs {
+		for _, value := range values {
+			fmt.Fprintln(os.Stdout, fmt.Sprintf(`{"output_name": "%s", "value": %s}`, key, string(value)))
```
joshma, line 48: output_name -> outputName in the interest of JSON / javascript style?

joshma, line 48: I'd prefer you used json.Marshal here, so that we can know for sure that it's valid output. Make a new struct, e.g.

```
type OutputRow struct {
  OutputName string `json:"outputName"`
  Value json.RawMessage `json:"value"`
}
```

(double check that it serializes correctly...I'm not totally sure)

and then use `j.enc.Encode` like the other functions do. note that for YAML below, you can probably add a `yaml` tag to each field as well (see how the other structs above do it)

pkg/print/yaml.go
```diff
@@ -31,3 +31,7 @@ func (YAML) runs(runs []api.Run) {
 func (YAML) run(run api.Run) {
 	yaml.NewEncoder(os.Stderr).Encode(run)
 }
+
+func (YAML) outputs(outputs api.Outputs) {
+	// TODO(sagar): implement this
```
joshma, line 36: To handle this + the JSON case above, I think you should do the following:

- Ensure api.Outputs is a list of json+yaml serializeable data type. For json, include `json:"..."` tags, and for yaml you want `yaml:"..."`.
- For JSON, print out one object per line (ie newline json https://jsonlines.org/)
- For YAML, I'm not sure, @colinking do you have opinions? I suppose you could either do a YAML list of outputs, or one YAML doc per output, separated by `---` (https://stackoverflow.com/questions/53092786/use-of-in-yaml)

joshma (changes requested):

pkg/print/print.go
```diff
@@ -37,3 +43,12 @@ func Runs(runs []api.Run) {
 func Run(run api.Run) {
 	DefaultFormatter.run(run)
 }
+
+func Outputs(outputs api.Outputs) {
+	if isatty.IsTerminal(os.Stdout.Fd()) {
```
joshma, line 48: ah forgot to comment on this - I think this is the wrong place to determine the default output. let's do DefaultFormatter here. in pkg/cmd/root/root.go:64, the default is set to `table`. how about you read `isatty.IsTerminal` there, and set it to `"table"` if is terminal and "json" if not?

pkg/print/json.go
```diff
@@ -40,3 +41,7 @@ func (j *JSON) runs(runs []api.Run) {
 func (j *JSON) run(run api.Run) {
 	j.enc.Encode(run)
 }
+
+func (j *JSON) outputs(outputs api.Outputs) {
+	errors.New("Not implemented")
+}
```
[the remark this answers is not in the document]
colinking: If we can keep this API error-less, similar to the `fmt` API, that'd be great! One approach we could take is for `Formatter` to expose an `Err() error` function that returns the last error, if any. That way, at the root, we can do one check of `formatter.Err()` before exiting to see if we hit any errors along the way.

pkg/print/yaml.go
colinking: Hmm, interesting, I've never thought about how YAML doesn't have a "stringified" variant like JSON until now 😄 

Using `---` delimiters seems reasonable to me!

pkg/print/table.go
```diff
@@ -76,3 +81,98 @@ func (t Table) runs(runs []api.Run) {
...
+	case nil:
+		return ""
+	default:
+		v, err := json.Marshal(t)
+		if err != nil {
+			return fmt.Sprintf("%v", value)
+		}
+		return string(v)
+	}
+	return ""
```
colinking, line 177: nit: don't need this, b/c of the default case! :)

pkg/print/json.go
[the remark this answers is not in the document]
joshma: that's pretty nifty! I'm in agreement then - maybe you can keep this errorless but printing to stderr and aborting while printing, and we can add in additional error handling later

pkg/print/json.go
colinking: Noting that we don't currently use YAML tags, since it defaults to using the field name, which has been reasonable so far. Would be nice to start adding yaml tags though. Or, instead switch to a YAML serialization library that lets you re-use JSON tags: https://github.com/ghodss/yaml

pkg/print/table.go
poudels14: removed

poudels14:
changes:
  - deserializing output to interface{} instead of json.RawMessage. it's easier to work
  - printing each line for JSON output and at once for YAML
  - tty working

pkg/print/json.go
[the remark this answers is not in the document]
poudels14: no need for error handling in print anymore. it no longer throws error

pkg/api/watcher_test.go
```diff
@@ -37,6 +38,12 @@ func TestWatcher(t *testing.T) {
 			{Logs: []LogItem{a, b}},
 		}
 
+		var outputs = GetOutputsResponse{
```
joshma, line 41:
```suggestion
		outputs := GetOutputsResponse{
```

pkg/print/json.go
joshma, line 48: Not addressed? Or is GH caching things weirdly? This should not be manually constructing JSON strings.

pkg/print/print.go
joshma, line 48: Also not addressed?

joshma: Also not addressed?

joshma (changes requested):

pkg/print/print.go
poudels14: already fixed this

pkg/print/json.go
poudels14: this is outdated

pkg/print/print.go
poudels14: also outdated. not sure why you are seeing that

pkg/print/json.go
```diff
@@ -9,34 +9,50 @@ import (
...
 
 // NewJSONFormatter returns a new json formatter.
 func NewJSONFormatter() *JSON {
-	enc := json.NewEncoder(os.Stderr)
-	enc.SetIndent("", "  ")
+	encErr := json.NewEncoder(os.Stderr)
+	encErr.SetIndent("", "  ")
+
+	encOut := json.NewEncoder(os.Stdout)
+	encOut.SetIndent("", "  ")
```
joshma, line 22: Sorry, one more thing - this should output 1 JSON per line ("newline-separate JSON" is a distinct datatype from JSON...). So I think you want to get rid of SetIndent.

You should be able to run `airplane tasks execute ... | jq .value` (once we move the other stdout prints to stderr...)

pkg/print/yaml.go
```diff
@@ -31,3 +31,16 @@ func (YAML) runs(runs []api.Run) {
...
+	var rows []api.OutputRow
+	for key, values := range outputs {
+		for _, value := range values {
+			rows = append(rows, api.OutputRow{
+				OutputName: key,
+				Value:      value,
+			})
+		}
+	}
+	yaml.NewEncoder(os.Stderr).Encode(rows)
```
joshma, line 45: this should be os.Stdout

joshma (approved):

PR #28, opened 2021-03-24
Add Linux and Windows (and non-homebrew Mac) installation scripts

colinking:
You can test this PR by running:

```sh
# Mac/Linux
curl -L https://github.com/airplanedev/cli/releases/download/v0.0.2-alpha.7/install.sh | bash -s v0.0.2-alpha.7

# Windows
# .. well something like this, but you need to pass in "v0.0.2-alpha.7" as an argument for now
iwr https://github.com/airplanedev/cli/releases/download/v0.0.2-alpha.7/install.ps1 -useb | iex
```

Once this merges, the "latest" release will get updated and you'll be able to run the above commands without specifying a specific version. Then these commands will be:

```sh
# Mac/Linux
curl -fsSL https://github.com/airplanedev/cli/releases/latest/download/install.sh | sh

# Windows
iwr https://github.com/airplanedev/cli/releases/latest/download/install.ps1 -useb | iex
```


.goreleaser.yml
```diff
@@ -12,24 +12,27 @@ builds:
       - linux
       - windows
       - darwin
+    ignore:
+      - goos: darwin
+        goarch: 386
+      - goos: linux
+        goarch: 386
+      - goos: windows
+        goarch: 386
```
joshma, line 21: This is just to not support 32-bit? Does it break stuff?

joshma (approved):

.goreleaser.yml
colinking: Fewer binaries, and we don't need to detect them in our installation script

PR #30, opened 2021-03-24
Fix brew install issue

colinking:
This previously produced a brew config with a newline: https://github.com/airplanedev/homebrew-tap/blob/4a472454b0def2aa7a11470322afe98fe9d89074/airplane.rb#L12-L13

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -25,8 +25,7 @@ archives:
     wrap_in_directory: false
     # This is the default name_template, but without the version. This allows install.sh to compute the
     # latest artifact name without knowing the version.
-    name_template: |
-      {{ .ProjectName }}_{{ .Os }}_{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}{{ if .Mips }}_{{ .Mips }}{{ end }}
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -25,8 +25,7 @@ archives:
     wrap_in_directory: false
     # This is the default name_template, but without the version. This allows install.sh to compute the
     # latest artifact name without knowing the version.
-    name_template: |
-      {{ .ProjectName }}_{{ .Os }}_{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}{{ if .Mips }}_{{ .Mips }}{{ end }}
+    name_template: "{{ .ProjectName }}_{{ .Os }}_{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}{{ if .Mips }}_{{ .Mips }}{{ end }}"
```
joshma, line 28: NB I think `|-` would have also worked

joshma (approved):

.goreleaser.yml
colinking: YAML is way too complicated lol 😭

PR #31, opened 2021-03-24
Fix conf-based login error

colinking:
This fixes an issue with `airplane login` that would lead to a "conf missing" error.

While here, I also noticed that `airplane login` was slow to shutdown when logging in through Chrome. Fixed by adding a `ReadHeaderTimeout`.


pkg/token/server.go
```diff
@@ -74,10 +77,13 @@ func (srv *Server) Token() <-chan string {
 
 // ServeHTTP implementation.
 func (srv *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	fmt.Printf("received connection...\n")
```
joshma, line 80: remove the Printfs

joshma (approved):

PR #32, opened 2021-03-24
Recommend `brew update` in installation instructions

colinking:
To upgrade the airplane CLI, you'll have to run `brew update` or else the latest version will not be installed.


joshma (approved):
Hm didn't know that haha - I'm sort of surprised.

colinking:
Yeah, brew will run this automatically when you do a `brew install X`, but not for updates!

PR #33, opened 2021-03-24
Merge `create` and `push` into `deploy` command

colinking:
We now expect a slug to be set in task definitions. We then do an upsert operation based on that slug.

```
# Creating
❯ lairlocal tasks deploy -f ./_examples/deno/task.yml
  Creating...
  Building...
  Updating...
  Done!

To execute Deno Echo:
- From the CLI: airplane tasks execute deno-echo -- [parameters]
- From the UI: https://app.airplane.so:5000/tasks/1qDMF83RMkHWlQNIyx3f6UPAN9u

# Updating
❯ lairlocal tasks deploy -f ./_examples/deno/task.yml
  Building...
  Updating...
  Done!

To execute Deno Echo:
- From the CLI: airplane tasks execute deno-echo -- [parameters]
- From the UI: https://app.airplane.so:5000/tasks/1qDMF83RMkHWlQNIyx3f6UPAN9u
```


joshma (commented):
Hm I'm not sure how I feel about modifying the yml file. User would have to remember to commit it for future use.

Or is that more for a migratory case? By default we init / define the slug?

pkg/api/types.go
```diff
@@ -7,20 +7,21 @@ import (
...
-	Command        []string          `json:"command" yaml:"command"`
-	Arguments      []string          `json:"arguments" yaml:"arguments"`
-	Parameters     Parameters        `json:"parameters" yaml:"parameters"`
-	Constraints    Constraints       `json:"constraints" yaml:"constraints"`
-	Env            map[string]string `json:"env" yaml:"env"`
-	ResourceLimits map[string]string `json:"resourceLimits" yaml:"resourceLimits"`
-	Builder        string            `json:"builder" yaml:"builder"`
-	BuilderConfig  map[string]string `json:"builderConfig" yaml:"builderConfig"`
-	Repo           string            `json:"repo" yaml:"repo"`
+	Slug           string            `json:"slug"`
```
joshma, line 10: for my info - why is the yaml tag no longer necessary? or rather...was it every necessary?

pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -0,0 +1,157 @@
...
+		return err
+	}
+
+	var taskID string
+	task, err := client.GetTask(ctx, def.Slug)
+	if err == nil {
+		taskID = task.ID
+	} else if aerr, ok := err.(api.Error); ok && aerr.Code == 404 {
+		// A task with this slug does not exist, so we should create one.
+		fmt.Println("  Creating...")
```
joshma, line 63: hm I sort of think we should prompt the user before creating, but maybe we should try w/o it at first

pkg/taskdef/taskdef.go
```diff
@@ -0,0 +1,77 @@
...
+
+// WriteSlug inserts a task definition into a file and attempts to
+// preserve the files existing format as much as possible.
+func WriteSlug(path, slug string) error {
+	b, err := ioutil.ReadFile(path)
+	if err != nil {
+		return errors.Wrap(err, "reading task definition")
+	}
+
+	// Find the first line without a document break. This is where
```
joshma, line 56: ooh man haha - there's pathological ways to break this, i think...what about finding `name:` and putting this above it? not foolproof either but could be better?

joshma (approved):
:D exciting

joshma:
make sure you update the PR description

pkg/taskdef/taskdef.go
colinking: Ah yeah, forgot to remove this! Don't need it anymore

pkg/cmd/tasks/deploy/deploy.go
colinking: Seems reasonable to me to create if the slug doesn't exist, but yeah let's revisit if it becomes a gotcha!

PR #37, opened 2021-03-25
Auto-login users

colinking:
This PR adds support for automatically logging in a user, via `airplane login`, if they are not yet authenticated.

For testing, I added an `airplane logout` command. I've also moved these both under the `auth` namespace, which matches our namespacing convention and the choice of namespace in similar CLIs.

You can test it like so:

```sh
airplane auth logout
airplane tasks list
```


pkg/cmd/auth/logout/logout.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/airplanedev/cli/pkg/cli"
+	"github.com/airplanedev/cli/pkg/conf"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+func New(c *cli.Config) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "logout",
+		Short: "Logout of Airplane",
```
joshma, line 15:
```suggestion
		Short: "Log out of Airplane",
```

pkg/cmd/auth/login/login.go
```diff
@@ -27,45 +24,45 @@ func New(c *cli.Config) *cobra.Command {
...
 
-// Open attempts to open the URL in the browser.
-//
-// As a special case, if `AP_BROWSER` env var is set to `none`
-// the command will always print the URL.
-func open(cmd *cobra.Command, url string) {
-	if os.Getenv("AP_BROWSER") != "none" {
-		if err := browser.Open(runtime.GOOS, url); err == nil {
-			return
+func EnsureLoggedIn(ctx context.Context, cmd *cobra.Command, c *cli.Config) error {
```
joshma, line 35: I think the normal `login` experience should do this, but a separate EnsureLoggedIn function should ideally do something like:
- if TTY, print "You are not logged in - log in first?" and open the browser then - it feels a bit aggressive/unexpected to open this URL for other commands
- if not TTY, just error - "You are not logged in - run the login command:\n    airplane auth login"

joshma (approved):

pkg/cmd/auth/login/login.go
colinking: Sounds good. Made some tweaks so that if you need to login, we confirm with you before redirecting. The confirmation goes through stderr, so you can still pipe stdout elsewhere.

PR #38, opened 2021-03-26
Support Docker builder

colinking:
This PR adds support for the Docker builder. While there, I tested the CLI on all of our new examples and fixed an edge case with our Go builder and with publishing multiple images to the same team.


pkg/build/builder.go
```diff
@@ -218,10 +225,6 @@ func (b *Builder) Push(ctx context.Context, tag string) error {
 
 // Dockerfile returns the dockerfile for the builder.
 func (b *Builder) dockerfile() (string, error) {
-	if b.args["entrypoint"] == "" {
-		return "", fmt.Errorf("build: .entrypoint is required")
-	}
```
colinking, line 223: This will live on in a future task definition validation library. Removing for now!

pkg/build/golang.go
```diff
@@ -18,14 +15,14 @@ func golang(root string, args Args) (string, error) {
...
-	if err := exist(gomod, gosum, main); err != nil {
+	if err := exist(gomod, main); err != nil {
 		return "", err
 	}
 
 	t, err := template.New("golang").Parse(`
 FROM golang:1.16.0-alpine3.13 as builder
 WORKDIR /airplane
-COPY go.mod go.sum ./
+COPY go.mod {{ if .HasGoSum -}} go.sum {{ end -}} ./
```
colinking, line 25: If you have no deps, like in our Hello World example, then you don't need (or even get) a `go.sum`.

pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -120,13 +120,13 @@ func run(ctx context.Context, cfg config) error {
...
 
 		fmt.Println("  Building...")
-		img, err := b.Build(ctx, taskID, "latest")
+		bo, err := b.Build(ctx, taskID, "latest")
 		if err != nil {
 			return errors.Wrap(err, "build")
 		}
 
 		fmt.Println("  Updating...")
-		if err := b.Push(ctx, img.RepoTags[0]); err != nil {
```
colinking, line 129: `img.RepoTags` includes all tags published to a given docker repo (f.e. everything matching `us-central1-docker.pkg.dev/airplane-prod/team-1jq9ndetctmyqqxajesk7v0wxav/*`). Fixed by returning just the built tag.

pkg/build/golang.go
joshma, line 25: lol minor but maybe just avoid the `-` business

```suggestion
COPY go.mod {{ if .HasGoSum }}go.sum {{ end }}./
```

pkg/build/golang.go
```diff
@@ -35,13 +32,11 @@ ENTRYPOINT ["go", "run", "/airplane/{{ .Main }}"]
 	}
 
 	var data struct {
-		Root  string
-		Main  string
-		Gomod string
-		Gosum string
```
joshma, line 41: were some/most of these fields just never used?

pkg/cmd/tasks/deploy/deploy.go
joshma: hm I can see why we want to return the right tag, but OOC why did the previous thing break with the push? wouldn't pushing one of the images have worked at least, even if incorrect?

joshma (approved):

pkg/cmd/tasks/deploy/deploy.go
colinking: It's a problem since it ends up pushing the image fforTask A as the image for Task B

pkg/build/golang.go
colinking: Nah

PR #40, opened 2021-03-29
Add full-featured init command

colinking:
This PR expands the functionality of the `airplane tasks init` command to support three ways of creating a local task definition -- either from an Airplane-provided sample, from scratch, or from an existing task (what we added support for last week).

You can try it out with:

```
airplane tasks init
```

You can still use `airplane tasks init --from hello_world`, however you can now access this same functionality by choosing `Create from an existing task` then selecting your task from the select menu.

Creating from scratch will likely need some TLC over time. The version in this PR is fairly minimal. Open for feedback on how we could improve it, but don't want to spend too much time on it right now.


pkg/cmd/tasks/initcmd/sample.go
```diff
@@ -0,0 +1,139 @@
...
+	} else {
+		// By default, store the sample in a directory with the same name
+		// as the containing directory in GitHub.
+		outputdir = path.Base(path.Dir(dir.DefinitionPath()))
+	}
+
+	// Rename the sample definition to match the filename in the `-f` argument.
+	// This is done to maintain semantic consistency with the other kinds of
+	// init, but is not strictly necessary.
+	defname := path.Base(cfg.file)
```
joshma, line 50: Hm I don't think this is being honored - I tried init-ing a Node sample app w/ `-f` and it still ended up being airplane.yml

pkg/cmd/tasks/initcmd/scratch.go
```diff
@@ -0,0 +1,150 @@
...
+			return err
+		}
+	}
+
+	if err := dir.WriteDefinition(def); err != nil {
+		return err
+	}
+
+	cmd.Printf(`
+An Airplane task definition for '%s' has been created in %s!
```
joshma, line 63: Maybe:

```suggestion
An Airplane task definition skeleton for '%s' has been created in %s! Fill it out with the rest of your task details.

When you're ready, deploy it to Airplane with:
```

pkg/cmd/tasks/initcmd/scratch.go
```diff
@@ -0,0 +1,150 @@
...
+		// TODO: let folks enter an image
+		def.Image = "alpine:3"
+		def.Command = []string{"echo", `"Hello World"`}
+	} else {
+		if def.Builder, def.BuilderConfig, err = defaultRuntimeConfig(runtime); err != nil {
+			return err
+		}
+	}
+
+	if err := dir.WriteDefinition(def); err != nil {
```
joshma, line 58: This doesn't write the other files like requirements.txt (python) right? I think we'll need to do some sort of code scaffolding support for the various langauges...maybe we can get rid of "from scratch" and just use example repos for now?

pkg/cmd/tasks/initcmd/init.go
```diff
@@ -36,50 +37,75 @@ func New(c *cli.Config) *cobra.Command {
...
+}
 
-	cmd.Printf("Created an Airplane task definition for %s in %s\n", res.Name, cfg.file)
+type initKind string
 
-	return nil
+var (
+	initKindSample  initKind = "Create from an Airplane-provided sample"
+	initKindScratch initKind = "Create from scratch"
+	initKindTask    initKind = "Create from an existing Airplane task"
```
joshma, line 87: Maybe we should make this read more like an import? It's not quite the same as the others...

```suggestion
	initKindTask    initKind = "Import definition from an existing Airplane task"
```

joshma (changes requested):

pkg/cmd/tasks/initcmd/scratch.go
colinking: Yep, so adding in the scaffolding brings us to what we talked about yesterday around how it's the same thing as creating a Hello World. Feels like we can do more here with helpful comments (descriptions of every airplane.yml fields for the chosen builder), etc. So I'm now thinking of this as the fast-path for creating your N+1-th task. Not sure how relevant it'll be, but figure we can experiment a bit with it and see what's helpful.

pkg/cmd/tasks/initcmd/init.go
colinking: Went with "clone"!

pkg/cmd/tasks/initcmd/sample.go
colinking: Ah, we weren't updating the file path (`file`) that we print, but the file itself is being renamed. (or at least, it is for me).

pkg/cmd/tasks/initcmd/init.go
```diff
@@ -36,50 +37,75 @@ func New(c *cli.Config) *cobra.Command {
...
 
 func run(ctx context.Context, cmd *cobra.Command, cfg config) error {
-	var client = cfg.cli.Client
+	var kind initKind
+	var err error
+	// If --from is provided, we already know the user wants to create
+	// from an existing task, so we don't need to prompt the user here.
+	if cfg.from == "" {
+		cmd.Printf("Airplane is the code-first solution for engineers building internal tools.\n\n")
+		cmd.Printf("This command will configure a task definition which specifies how to deploy your task to Airplane.\n\n")
```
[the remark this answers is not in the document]
colinking: maybe `... which Airplane uses to deploy your task`?

pkg/cmd/tasks/initcmd/init.go
colinking:
```sh
❯ lairlocal tasks init --from echo_loop
X Sorry, your reply was invalid: You must pick a new slug.
? Pick a unique identifier (slug) for this task echo_loop_2

An Airplane task definition for 'Echo Loop' has been created!

To deploy it to Airplane, run:
  airplane tasks deploy -f airplane.yml

❯ cat airplane.yml
slug: echo_loop_2
# ...
```

PR #43, opened 2021-03-30
Add debug logging

colinking:
I ran into a 500 error that I couldn't debug very easily. To make this easier, I've pulled the `--debug` flag up to the root from just the `tasks deploy` command and created a `logger` package that we can use for debug logging. This PR replaces all calls to `fmt.Print*` or `cmd.Print*` with `logger.[Log|Debug]` calls. In doing so, this re-routes a number of existing logs to stderr instead of stdout. It also allows us to avoid passing around a `*cobra.Command` just for logging.

The `logger` package does not do much right now, but I imagine we may add logic for colorizing logs down the road. We could also consider folding it in elsewhere, since it's such a small package.

The main functional change of this PR is that we now get full stack traces if there's an error and a command is run with `--debug`.


pkg/logger/logger.go
```diff
@@ -0,0 +1,20 @@
...
+	"fmt"
+	"os"
+)
+
+var (
+	EnableDebug bool
+)
+
+func Log(msg string, args ...interface{}) {
+	fmt.Fprintf(os.Stderr, msg, args...)
```
joshma, line 13: 😍

joshma (approved):

PR #45, opened 2021-03-30
Unpin major versions of node in builder

joshma:
Supports e.g. "15" for version instead of the full version, and resolves
that version to the pinned minor version.

### DIFF
--- a/pkg/build/node.go
+++ b/pkg/build/node.go
@@ -23,7 +23,7 @@ func node(root string, args Args) (string, error) {
 
 	// Default version.
 	if version == "" {
-		version = "15.8"
+		version = "15"
 	}
 
 	// Make sure that entrypoint and `package.json` exist.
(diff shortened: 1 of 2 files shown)


pkg/build/node.go
```diff
@@ -86,3 +86,18 @@ ENTRYPOINT ["node", "{{ .Main }}"]
 
 	return buf.String(), nil
 }
+
+// expandVersion returns a pinned minor version of Node to use
+func expandVersion(version string) string {
```
colinking, line 91: Since it's all in the same package, maybe `expandNodeVersion`?

pkg/build/node.go
```diff
@@ -86,3 +86,18 @@ ENTRYPOINT ["node", "{{ .Main }}"]
...
+	switch version {
+	case "15":
```
colinking, line 93: Maybe the default version should live here?

```suggestion
	switch version {
	case "":
	  fallthrough
	case "15":
```

colinking (approved):

PR #46, opened 2021-03-30
Prompt users for a slug when creating from a sample

colinking:
This will now prompt users for a slug when creating a task from a sample. In a future PR, I'll add logic for validating that a slug is unique, and at that point I imagine we can make this prompt conditional on the slug not being unique already.

```
❯ lairlocal tasks init
Airplane is a development platform for engineers building internal tools.

This command will configure a task definition which Airplane uses to deploy your task.

? How do you want to get started? Create from an Airplane-provided sample
? Pick a runtime: Node.js
? Pick a sample: Hello World
? Pick a unique identifier (slug) for this task: hello_world

An Airplane task definition for 'Hello World' has been created!

To deploy it to Airplane, run:
  airplane tasks deploy -f hello-world-javascript/airplane.yml
```

For this PR, I brought back some YAML parsing logic that I had used (and removed) in a previous PR. We can use this to update fields in a YAML document without losing comments/formatting/etc. that is lost when (un)marshaling.

### DIFF
--- a/pkg/taskdir/taskdef.go
+++ b/pkg/taskdir/taskdef.go
@@ -30,7 +30,7 @@ type Definition struct {
 	Builder        string             `yaml:"builder,omitempty"`
 	BuilderConfig  api.BuilderConfig  `yaml:"builderConfig,omitempty"`
 	Repo           string             `yaml:"repo,omitempty"`
-	Timeout        int                `yaml:"timeout"`
+	Timeout        int                `yaml:"timeout,omitempty"`
 }
 
 func (this Definition) Validate() (Definition, error) {
@@ -64,6 +64,10 @@ func (this Definition) Validate() (Definition, error) {
 	// TODO: validate the rest of the fields!
 
 	return this, nil
+}
+
+func (this TaskDirectory) DefinitionPath() string {
+	return this.path
 }
 
 func (this TaskDirectory) ReadDefinition() (Definition, error) {
--- a/pkg/taskdir/taskdef.go
+++ b/pkg/taskdir/taskdef.go
@@ -29,6 +29,15 @@ type Definition struct {
 	BuilderConfig  api.BuilderConfig  `yaml:"builderConfig,omitempty"`
 	Repo           string             `yaml:"repo,omitempty"`
 	Timeout        int                `yaml:"timeout,omitempty"`
+
+	// Root is a directory path relative to the parent directory of this
+	// task definition which defines what directory should be included
+	// in the task's Docker image.
+	//
+	// If not set, defaults to "." (in other words, the parent directory of this task definition).
+	//
+	// This field is ignored when using the pre-built image builder (aka "manual").
+	Root string `yaml:"root,omitempty"`
 }
 
 func (this Definition) Validate() (Definition, error) {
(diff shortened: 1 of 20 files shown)


pkg/taskdir/taskdef.go
```diff
@@ -35,28 +33,16 @@ type Definition struct {
...
-			survey.WithValidator(func(val interface{}) error {
-				if str, ok := val.(string); !ok || !utils.IsSlug(str) {
-					return errors.New("Slugs can only contain lowercase letters, underscores, and numbers.")
-				}
-
-				return nil
-			}),
-		); err != nil {
-			return this, errors.Wrap(err, "prompting for slug")
+		defaultSlug := utils.MakeSlug(this.Name)
```
joshma, line 43: Just so I understand - if we init from sample, we prompt for a slug. When we deploy, we expect there to be a slug, but as a backup we'll prompt for a slug too? Should this also WriteSlug? Otherwise if you pick a new slug the 2nd time you'll end up with a second task.

IMO we could just not prompt and error.

pkg/taskdir/taskdef.go
```diff
@@ -84,6 +70,17 @@ func (this TaskDirectory) ReadDefinition() (Definition, error) {
 	return def, nil
 }
 
+// WriteSlug updates the slug of a task definition and persists this to disk.
+//
+// It attempts to retain the existing file's formatting (comments, etc.) where possible.
+func (this TaskDirectory) WriteSlug(slug string) error {
```
joshma, line 76: Can we log to stderr that we modified `X.yaml`? (Maybe even encourage them to commit the change if we detect .git or whatnot, but that may be overkill.)

pkg/utils/yaml.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	if len(root.Content) == 0 {
+		return errors.Errorf("cannot insert %s: yaml document empty", field)
+	}
+	// Find the root map, which may not be the first element due to comments.
+	var mapnode *yaml.Node
+	for _, subnode := range root.Content {
+		if subnode.Kind == yaml.MappingNode {
+			mapnode = subnode
```
joshma, line 49: break after this?

pkg/taskdir/taskdef.go
```diff
@@ -84,6 +70,17 @@ func (this TaskDirectory) ReadDefinition() (Definition, error) {
 	return def, nil
 }
 
+// WriteSlug updates the slug of a task definition and persists this to disk.
+//
+// It attempts to retain the existing file's formatting (comments, etc.) where possible.
```
joshma, line 75: I'm OK w/ what you have below, but this _does_ rewrite a lot of the file right? e.g. it reformats the indenting?

pkg/taskdir/taskdef.go
colinking: Agreed, we don't need this anymore now that we're setting it via `init`.

pkg/utils/yaml.go
colinking: good catch

pkg/taskdir/taskdef.go
colinking: Yep, unfortunately :( It's the best approach I know of, short of doing a regexp replace on `slug: <prev>` -> `slug: <new>`. But this approach seems more reliable!

joshma (approved):

PR #47, opened 2021-03-31
Support per-host tokens

colinking:
This is useful for our internal development, where we test the CLI against a local/staging/production host. Previously tokens were global, but now they are per-host so there will be a token persisted for each host you've used.

### DIFF
--- a/pkg/cmd/auth/logout/logout.go
+++ b/pkg/cmd/auth/logout/logout.go
@@ -23,12 +23,14 @@ func New(c *cli.Config) *cobra.Command {
 
 func run(ctx context.Context, c *cli.Config) error {
 	cfg, err := conf.ReadDefault()
-	if !errors.Is(err, conf.ErrMissing) {
-		cfg.Tokens[c.Client.Host] = ""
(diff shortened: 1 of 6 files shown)


pkg/cmd/auth/logout/logout.go
```diff
@@ -23,12 +23,14 @@ func New(c *cli.Config) *cobra.Command {
...
 	cfg, err := conf.ReadDefault()
-	if !errors.Is(err, conf.ErrMissing) {
-		cfg.Token = ""
+	if err == nil || !errors.Is(err, conf.ErrMissing) {
+		delete(cfg.Tokens, c.Client.Host)
 
 		if err := conf.WriteDefault(cfg); err != nil {
 			return err
 		}
+	} else if err != nil {
```
joshma, line 32: so this only returns the err if it IS `conf.ErrMissing`? why this change?

colinking: Ah yeah, the logic here for ErrMissing was wrong. Thanks for catching! Made the check simpler.

PR #48, opened 2021-03-31
Support root directory override

colinking:
This allows you to add an optional `root` field to your task definition with a relative directory path (from the parent directory of your task definition) that determines which directory to build your image from.

For example, if your code was laid out like:

```
utils/db.go

taskA/main.go
taskA/airplane.yml

taskB/main.go
taskB/airplane.yml
```

You would be able to set `taskA/airplane.yml` to have `root: "../"` and it would build a Docker image that contains `utils` along with `taskA` and `taskB`, which would allow you to share the `utils` package across both tasks.

Note that other file paths, such as entrypoints, are relative to the root so if you adjust root then make sure to also adjust those filepaths.

### DIFF
--- a/pkg/taskdir/directory.go
+++ b/pkg/taskdir/directory.go
@@ -43,6 +43,10 @@ func Open(file string) (TaskDirectory, error) {
 	}
 	td.rootPath = path.Join(filepath.Dir(td.defPath), def.Root)
 
+	if !strings.HasPrefix(td.defPath, td.rootPath+string(filepath.Separator)) {
+		return TaskDirectory{}, errors.Errorf("%s must be inside of the task's root directory: %s", path.Base(td.defPath), td.rootPath)
+	}
+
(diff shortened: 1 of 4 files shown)


pkg/taskdir/directory.go
```diff
@@ -26,22 +26,34 @@ func Open(file string) (TaskDirectory, error) {
...
+		}
 	}
 
-	td.Dir, err = filepath.Abs(filepath.Dir(td.path))
+	def, err := td.ReadDefinition()
 	if err != nil {
-		return TaskDirectory{}, errors.Wrap(err, "parsing file directory")
+		return TaskDirectory{}, err
 	}
+	td.rootPath = path.Join(filepath.Dir(td.defPath), def.Root)
```
joshma, line 44: how about we validate defPath is inside rootPath?

joshma (approved):

PR #49, opened 2021-04-01
Export func to convert config -> Dockerfile

poudels14:
- this function will be used by cloud builder in airport pkg

### DIFF
--- a/pkg/build/builder.go
+++ b/pkg/build/builder.go
@@ -71,6 +71,12 @@ type Config struct {
 	Writer io.Writer
 }
 
+type DockerfileConfig struct {
+	Builder string
+	Root    string
+	Args    Args
+}
+
 // Builder implements an image builder.
 type Builder struct {
 	root   string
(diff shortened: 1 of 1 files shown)


pkg/build/builder.go
```diff
@@ -218,22 +218,11 @@ func (b *Builder) Push(ctx context.Context, tag string) error {
...
-	switch b.name {
-	case "go":
-		return golang(b.root, b.args)
-	case "deno":
-		return deno(b.root, b.args)
-	case "python":
-		return python(b.root, b.args)
-	case "node":
-		return node(b.root, b.args)
-	default:
```
colinking, line 234: There have been some changes on `main`, would recommend merging in first: https://github.com/airplanedev/cli/blob/1b636a30c091b1ad44af3b67e73c877d7f0d0c21/pkg/build/builder.go#L227

pkg/build/builder.go
```diff
@@ -218,22 +218,11 @@ func (b *Builder) Push(ctx context.Context, tag string) error {
 
 // Dockerfile returns the dockerfile for the builder.
 func (b *Builder) dockerfile() (string, error) {
```
colinking, line 220: Let's just call `BuildDockerfile` directly, instead of `dockerfile()`, since `dockerfile()` is only used once.

pkg/build/builder.go
```diff
@@ -268,3 +257,22 @@ func sanitizeTaskID(s string) string {
 	}
 	return s
 }
+
+func BuildDockerfile(c Config) (string, error) {
```
colinking, line 261: Let's use a separate config struct just for `BuildDockerfile`, maybe:

```go
type DockerfileConfig struct {
 // ...
}
```

colinking (approved):

PR #53, opened 2021-04-02
Fix logger.Log to not use Fprintf if no args

joshma:
This was causing MISSING / NOVERB to be interpolated into messages that
had percent signs, since Go was treating them as format strings.

### DIFF
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -13,7 +13,12 @@ var (
 // Log writes a log message to stderr, followed by a newline. Printf-style
 // formatting is applied to msg using args.
 func Log(msg string, args ...interface{}) {
-	fmt.Fprintf(os.Stderr, msg+"\n", args...)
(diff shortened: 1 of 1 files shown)


colinking (approved):

pkg/logger/logger.go
```diff
@@ -13,7 +13,12 @@ var (
 // Log writes a log message to stderr, followed by a newline. Printf-style
 // formatting is applied to msg using args.
 func Log(msg string, args ...interface{}) {
-	fmt.Fprintf(os.Stderr, msg+"\n", args...)
+	if len(args) == 0 {
+		// Use Fprint if no args - avoids treating msg like a format string
+		fmt.Fprint(os.Stderr, msg+"\n")
+	} else {
+		fmt.Fprintf(os.Stderr, msg+"\n", args...)
+	}
```
colinking, line 21: Actually, can you add this for Debug too?

PR #54, opened 2021-04-06
Build Go binaries instead of directly running

colinking:
Previously, we would `go run` the Go entrypoint. If we do that, we have to use the same image containing all of the dependencies from `go mod download` which from internal testing up to 1.5GB. Instead, we move towards building and pulling the binary into a separate base image without the dependencies. Since we're using Alpine, this gets us down to ~300MB.


joshma (approved):
🎉

PR #56, opened 2021-04-06
Remove entrypoint check in BuildDocker

poudels14:
This was added back accidentally while merging commits

### DIFF
--- a/pkg/build/builder.go
+++ b/pkg/build/builder.go
@@ -277,10 +277,6 @@ func exist(paths ...string) error {
 }
 
 func BuildDockerfile(c DockerfileConfig) (string, error) {
-	if c.Args["entrypoint"] == "" {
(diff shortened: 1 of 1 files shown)


pkg/build/builder.go
```diff
@@ -277,10 +277,6 @@ func exist(paths ...string) error {
 }
 
 func BuildDockerfile(c DockerfileConfig) (string, error) {
-	if c.Args["entrypoint"] == "" {
-		return "", fmt.Errorf("build: .entrypoint is required")
-	}
-
```
colinking, line 283: Can you bring `docker` back too? 🙏🏻 
 https://github.com/airplanedev/cli/commit/e98015481812df895b3b0a99cb4b4ae6494934af

poudels14: done. didn't realize docker was removed too

colinking (approved):

PR #57, opened 2021-04-06
Add support for uploading to the remote builder

colinking:
This PR adds a new `--builder` flag to the `deploy` command which allows you to toggle between the local builder (via the local Docker daemon) or the remote builder (via the Airplane hosted Docker builder). For now, we default to `--builder=local` since the remote builder isn't live yet.

As for remote building functionality, this PR adds logic for archiving and uploading source code into GCS where it'll be available to our hosted builder. Once it does that, it temporarily errors until the remote builder is ready.

I've left a number of TODO comments that will be addressed in future passes, but let me know if any aren't clear.

### DIFF
--- a/pkg/build/remote.go
+++ b/pkg/build/remote.go
@@ -2,7 +2,9 @@ package build
 
 import (
 	"context"
+	"fmt"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"path"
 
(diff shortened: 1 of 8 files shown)


pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -94,39 +102,46 @@ func run(ctx context.Context, cfg config) error {
...
+			logger.Log("  Building...")
+			bo, err := b.Build(ctx, taskID, "latest")
+			if err != nil {
+				return errors.Wrap(err, "build")
+			}
+
+			logger.Log("  Updating...")
+			if err := b.Push(ctx, bo.Tag); err != nil {
+				return errors.Wrap(err, "push")
+			}
```
colinking, line 142: As it currently stands, there isn't much shared CLI logic between the local and remote builders. That might change as we start handling build logs and progress/status indicators, but TBD.

joshma: Would be nice to pull this into `build.Local(...)` at least?

pkg/build/remote.go
```diff
@@ -0,0 +1,88 @@
...
+
+	// Archive the root task directory:
+	// TODO: filter out files/directories that match .dockerignore
+	archiveName := "airplane-build.tar.gz"
+	archivePath := path.Join(tmpdir, archiveName)
+	// We want to produce an archive where the contents of the archive
+	// are the files inside of `dir.DefinitionRootPath()`, rather than
+	// a directory containing those files. Therefore, we need to produce
+	// a list of files/directories within the root directory instead of
+	// directly providing mholt/archiver with `dir.DefinitionRootPath()`.
```
joshma, line 34: Can you try...rephrasing this comment? I'm not sure I understand - I'm guessing passing the root path directly ends up with some undesired folder hierarchy? Does it do `/workdir/rootpath/x` instead of `/workdir/x`?

joshma (approved):

pkg/cmd/tasks/deploy/deploy.go
colinking: Sounds good. Will probably end up refactoring this at some point as we improve `deploy`'s output :)

pkg/build/remote.go
colinking: Done

colinking: But no, the other way around! LMK if this is more clear.

joshma: Yep, LGTM!

PR #58, opened 2021-04-06
Add `configs get` and `configs set` commands

joshma:
To set a config:

    airplane configs set --secret db some_db_url

To get a config:

    airplane configs get db

### DIFF
--- a/pkg/cmd/configs/set/set.go
+++ b/pkg/cmd/configs/set/set.go
@@ -12,8 +12,8 @@ import (
 	"github.com/airplanedev/cli/pkg/cli"
 	"github.com/airplanedev/cli/pkg/configs"
 	"github.com/airplanedev/cli/pkg/logger"
+	"github.com/airplanedev/cli/pkg/utils"
 	"github.com/fatih/color"
-	"github.com/mattn/go-isatty"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
@@ -97,8 +97,7 @@ func run(ctx context.Context, c *cli.Config, name string, argValue *string, secr
 
 func readValue(secret bool) (string, error) {
 	var value string
-	if isatty.IsTerminal(os.Stdin.Fd()) {
-		// Prompt
+	if utils.CanPrompt() {
 		msg := "Config value:"
 		var prompt survey.Prompt
 		if secret {
--- a/pkg/cmd/configs/set/set.go
+++ b/pkg/cmd/configs/set/set.go
@@ -1,0 +1,64 @@
+package set
+
+import (
+	"context"
+
+	"github.com/airplanedev/cli/pkg/api"
+	"github.com/airplanedev/cli/pkg/cli"
+	"github.com/airplanedev/cli/pkg/configs"
+	"github.com/airplanedev/cli/pkg/logger"
+	"github.com/fatih/color"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+var (
+	blue  = color.New(color.FgHiBlue).SprintFunc()
+	green = color.New(color.FgGreen).SprintFunc()
+)
+
+// New returns a new set command.
+func New(c *cli.Config) *cobra.Command {
+	var secret bool
+	cmd := &cobra.Command{
+		Use:   "set [--secret] <name> <value>",
+		Short: "Set a new or existing config variable",
+		Args:  cobra.ExactArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return run(cmd.Context(), c, args[0], args[1], secret)
+		},
+	}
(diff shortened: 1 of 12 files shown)


pkg/print/table.go
```diff
@@ -171,3 +176,15 @@ func getCellValue(value interface{}) string {
 		return string(v)
 	}
 }
+
+// print config as table
+func (t Table) config(config api.Config) {
```
joshma, line 181: We might want to rename `Table` here, as it's not always a "table" output? It's just...the default human friendly one.

I could _not_ have this go through the `print` module, but I wanted the JSON/YAML variants to work.

pkg/cmd/configs/set/set.go
```diff
@@ -0,0 +1,64 @@
...
+var (
+	blue  = color.New(color.FgHiBlue).SprintFunc()
+	green = color.New(color.FgGreen).SprintFunc()
+)
+
+// New returns a new set command.
+func New(c *cli.Config) *cobra.Command {
+	var secret bool
+	cmd := &cobra.Command{
+		Use:   "set [--secret] <name> <value>",
```
colinking, line 24: For secrets, we should disallow you from providing the value on the command line since it ends up in your shell history.

Maybe we read it from stdin, and if nothing is in stdin, then we open an editor to set the value and then we save/trim the value from the editor (something like what git does for commit messages)?

colinking: So you could do `echo "postgres://" | airplane configs set --secret db/dsn` if you want this behavior, but at least the default would encourage you to be more safe about it.

pkg/cmd/configs/set/set.go
```diff
@@ -0,0 +1,64 @@
...
+		return errors.Wrap(err, "parsing config name")
+	}
+	// Avoid printing back secrets
+	var valueStr string
+	if secret {
+		valueStr = "<secret value>"
+	} else {
+		valueStr = value
+	}
+	logger.Log("  Setting %s to %s...", blue(name), green(valueStr))
```
colinking, line 52: 😍

pkg/print/json.go
```diff
@@ -49,3 +50,8 @@ func (j *JSON) outputs(outputs api.Outputs) {
 		}
 	}
 }
+
+// Config implementation.
+func (j *JSON) config(config api.Config) {
+	j.enc.Encode(config)
```
colinking, line 56: So is the goal to mirror k8s, where you can get the underlying value by outputting to JSON?

colinking (approved):

pkg/print/json.go
joshma: Yeah exactly - figured it's nice to be consistent here and output machine-readable stuff in general for things like `airplane configs get X | jq -r.value`.

Although IDK, just returning the value alone could be fine, that's pretty scriptable as well.

pkg/cmd/configs/set/set.go
joshma: Hrm yeah, I noticed heroku lets you do this on the CLI but vercel prompts instead. (My half-justification is you can put a space before your command to not log it to history, but that's easy to forget.)

I don't like having two diff ways of doing this for secret/non-secret, so how about for all configs you have to do this? Either paste it in at the prompt (don't think we need to do an editor thing, can just prompt them without displaying the value) or accept it from stdin?

joshma: @colinking I updated this to accept all 3 😇  (stdin, prompt, argument). The examples discourage arguments, but all 3 works for secrets as well as non-secrets. WDYT?

pkg/cmd/configs/set/set.go
```diff
@@ -62,3 +94,26 @@ func run(ctx context.Context, c *cli.Config, name, value string, secret bool) er
 	logger.Log("  Done!")
 	return nil
 }
+
+func readValue() (string, error) {
+	var value string
+	if isatty.IsTerminal(os.Stdin.Fd()) {
```
colinking, line 100: `s/utils.CanPrompt()/`!

pkg/cmd/configs/set/set.go
```diff
@@ -62,3 +94,26 @@ func run(ctx context.Context, c *cli.Config, name, value string, secret bool) er
...
+		// Prompt
+		if err := survey.AskOne(
+			&survey.Input{Message: "Config value:"},
+			&value,
```
colinking, line 104: Should we use https://github.com/AlecAivazis/survey#password if --secret?

colinking (approved):

pkg/cmd/configs/set/set.go
joshma: Yep, beat me to it! Just noticed that / fixing.

pkg/cmd/configs/set/set.go
joshma: NB made CanPrompt check both stdin and stderr

PR #59, opened 2021-04-07
Wait for hosted build to complete before returning

colinking:
This PR updates the CLI's remote builder logic to:
1. Create a `build` using the exact task revision that was just created/updated.
2. Attach the upload to that build.
3. Long-poll that build until it has stopped.

I also split the `Remote` function up into a few helper functions to make it more readable.

In order to get the exact task revision for `#1` above, we now update your task before building/pushing.

--

Once this PR merges, remote building from the CLI will work once the backend components launch. We will want do some further polish here down the road though, specifically around surfacing progress and build logs (esp in the case of an error).

### DIFF
--- a/pkg/api/client.go
+++ b/pkg/api/client.go
@@ -96,8 +96,8 @@ func (c Client) CreateTask(ctx context.Context, req CreateTaskRequest) (res Crea
 }
 
 // UpdateTask updates a task with the given req.
-func (c Client) UpdateTask(ctx context.Context, req UpdateTaskRequest) (err error) {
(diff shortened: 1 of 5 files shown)


joshma (approved):
> In order to get the exact task revision for #1 above, we now update your task before building/pushing.

HMM can we create a revision without making it the latest version of the task? Ideally you can do that and then link the revision afterwards?

colinking:
> HMM can we create a revision without making it the latest version of the task? Ideally you can do that and then link the revision afterwards?

Yeah that would be great, I'll create a separate Linear ticket for that.

PR #61, opened 2021-04-07
Add aliases for common commands

colinking:
This PR adds aliases for the following commands:

- `airplane init` == `airplane tasks init`
- `airplane deploy` == `airplane tasks deploy`
- `airplane execute` == `airplane tasks execute`
- `airplane login` == `airplane auth login`
- `airplane logout` == `airplane auth logout`

It also adds singular aliases for each plural namespace (`configs`, `tasks`, `runs`) so that you can run the following:

```
airplane task deploy ...
```

### DIFF
--- a/pkg/cmd/root/root.go
+++ b/pkg/cmd/root/root.go
@@ -8,6 +8,8 @@ import (
 	"github.com/airplanedev/cli/pkg/api"
 	"github.com/airplanedev/cli/pkg/cli"
 	"github.com/airplanedev/cli/pkg/cmd/auth"
+	"github.com/airplanedev/cli/pkg/cmd/auth/login"
+	"github.com/airplanedev/cli/pkg/cmd/auth/logout"
 	"github.com/airplanedev/cli/pkg/cmd/configs"
 	"github.com/airplanedev/cli/pkg/cmd/runs"
 	"github.com/airplanedev/cli/pkg/cmd/tasks"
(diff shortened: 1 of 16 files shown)


pkg/cmd/root/root.go
```diff
@@ -81,7 +84,12 @@ func New() *cobra.Command {
 	cmd.PersistentFlags().BoolVar(&cfg.DebugMode, "debug", false, "Whether to produce debugging output.")
 	cmd.PersistentFlags().BoolVarP(&cfg.Version, "version", "v", false, "Print the CLI version.")
 
-	// Sub-commands.
+	// Aliases for popular namespaced commands:
+	cmd.AddCommand(initcmd.New(cfg))
+	cmd.AddCommand(deploy.New(cfg))
+	cmd.AddCommand(execute.New(cfg))
```
joshma, line 90: Oh nice - does this work always? Does this assume the parent subcommand isn't setting up any persistent flags, etc.?

joshma (approved):

pkg/cmd/root/root.go
colinking: Great catch. This was missing the auto-login behavior that comes from a `PersistentPreRunE` on the `tasks` namespace. Wish Cobra had a way to alias commands, but that doesn't seem to be supported. Instead, since I hooked up each of these commands to also run the `PersistentPreRunE` since it is idempotent.

colinking: Also I swapped `context.TODO()` for `cmd.Root().Context()`!

colinking: Oh, and I also added aliases for `login` and `logout` haha

joshma (approved):

joshma:
I'm not super happy with the repeated wrapping, it's going to be easy to forget that (or remember why we added that). OK for now.

PR #64, opened 2021-04-08
Add commands to create, list, and delete API keys

joshma:
This adds the following commands:

    $ airplane apikeys create agent
    Creating API key named agent...
    Done!
    Save this somewhere safe, as you won't be able to retrieve it later:

      tkn_5XohzCpAWTCDhqCOLBg0YvhgRTRwpstCoRqE

      Key ID: 1qrx6ri533i0FoW1KWLepyMmCto
      Team ID: 1jq9NDeTcTmyqqXAjEsk7V0wXaV

    $ airplane apikeys delete 1qrx6ri533i0FoW1KWLepyMmCto
        Deleting key 1qrx6ri533i0FoW1KWLepyMmCto...
        Done.

    $ airplane apikeys list
                      ID              |        CREATED AT
    ------------------------------+----------------------------
      1qrb8R6IoEABqqVErMYQP5gEHa2 | 2021-04-07T17:08:15-07:00
      1qrbEkSTeRoxNA8pmrDf8yUIgXH | 2021-04-07T17:09:06-07:00
      1qrc0oe29gAwud5reQueY8k805S | 2021-04-07T17:15:28-07:00
      1qrw0adF2fdObxKnQ7cBhNMfehD | 2021-04-07T19:59:54-07:00
      1qrw1gt7ne2km9Fjo78d1mz5Ijf | 2021-04-07T20:00:02-07:00
      1qrw6VxV1rbAn9xDhQlsv7p3cNG | 2021-04-07T20:00:41-07:00
      1qrwwYVTnrR0dadmYVGaWYcSKnN | 2021-04-07T20:07:35-07:00

### DIFF
--- a/pkg/cmd/apikeys/apikeys.go
+++ b/pkg/cmd/apikeys/apikeys.go
@@ -21,7 +21,7 @@ func New(c *cli.Config) *cobra.Command {
 		Long:    "Manage API keys",
 		Aliases: []string{"apikey"},
 		Example: heredoc.Doc(`
-			airplane apikeys create
+			airplane apikeys create "Agent Key"
 			airplane apikeys list
 			airplane apikeys delete <key_id>
 		`),
--- a/pkg/cmd/apikeys/apikeys.go
+++ b/pkg/cmd/apikeys/apikeys.go
@@ -1,0 +1,38 @@
+package apikeys
+
+import (
+	"context"
+
+	"github.com/MakeNowJust/heredoc"
+	"github.com/airplanedev/cli/pkg/cli"
+	"github.com/airplanedev/cli/pkg/cmd/apikeys/create"
+	"github.com/airplanedev/cli/pkg/cmd/apikeys/delete"
+	"github.com/airplanedev/cli/pkg/cmd/apikeys/list"
+	"github.com/airplanedev/cli/pkg/cmd/auth/login"
+	"github.com/airplanedev/cli/pkg/utils"
+	"github.com/spf13/cobra"
+)
+
+// New returns a new cobra command.
+func New(c *cli.Config) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "apikeys",
+		Short:   "Manage API keys",
+		Long:    "Manage API keys",
+		Aliases: []string{"apikey"},
+		Example: heredoc.Doc(`
+			airplane apikeys create
+			airplane apikeys list
+			airplane apikeys delete <key_id>
+		`),
+		PersistentPreRunE: utils.WithParentPersistentPreRunE(func(cmd *cobra.Command, args []string) error {
+			return login.EnsureLoggedIn(context.TODO(), c)
+		}),
+	}
+
+	cmd.AddCommand(create.New(c))
+	cmd.AddCommand(delete.New(c))
+	cmd.AddCommand(list.New(c))
+
+	return cmd
+}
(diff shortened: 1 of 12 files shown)


pkg/cmd/apikeys/apikeys.go
```diff
@@ -0,0 +1,38 @@
...
+		Short:   "Manage API keys",
+		Long:    "Manage API keys",
+		Aliases: []string{"apikey"},
+		Example: heredoc.Doc(`
+			airplane apikeys create "Agent Key"
+			airplane apikeys list
+			airplane apikeys delete <key_id>
+		`),
+		PersistentPreRunE: utils.WithParentPersistentPreRunE(func(cmd *cobra.Command, args []string) error {
+			return login.EnsureLoggedIn(context.TODO(), c)
```
colinking, line 29: Just recently changed this in another PR:

```suggestion
			return login.EnsureLoggedIn(cmd.Root().Context(), c)
```

pkg/cmd/apikeys/delete/delete.go
```diff
@@ -0,0 +1,46 @@
...
+func run(ctx context.Context, c *cli.Config, apiKeyIDs []string) error {
+	var client = c.Client
+
+	for _, apiKeyID := range apiKeyIDs {
+		req := api.DeleteAPIKeyRequest{
+			KeyID: apiKeyID,
+		}
+		logger.Log("  Deleting key %s...", red(apiKeyID))
+		if err := client.DeleteAPIKey(ctx, req); err != nil {
+			return errors.Wrap(err, "creating API key")
```
colinking, line 41:
```suggestion
			return errors.Wrap(err, "deleting API key")
```

pkg/cmd/apikeys/create/create.go
```diff
@@ -0,0 +1,50 @@
...
+	cmd := &cobra.Command{
+		Use:   "create [<key_name>]",
+		Short: "Generates a new API key for self-hosting agents and building custom integrations",
+		Args:  cobra.RangeArgs(0, 1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			var name string
+			if len(args) > 0 {
+				name = args[0]
+			}
+			return run(cmd.Context(), c, name)
```
colinking, line 26: Would recommend using `cmd.Root().Context()` since Cobra doesn't appear to pass the `ctx` down to child commands :( It just ends up as `nil` instead.

pkg/cmd/apikeys/create/create.go
```diff
@@ -0,0 +1,50 @@
...
+
+	req := api.CreateAPIKeyRequest{
+		Name: name,
+	}
+	resp, err := client.CreateAPIKey(ctx, req)
+	if err != nil {
+		return errors.Wrap(err, "creating API key")
+	}
+
+	print.APIKeyCreated(resp.APIKey)
```
colinking, line 48: This kind of output -- that is specific to a single command -- should probably not leak into the `print` package -- maybe we can just pass in an object with struct tags (json, yaml, table) that then gets printed through a generic `print` function?

colinking (approved):

pkg/cmd/apikeys/apikeys.go
colinking: Noting that we'll need to update this in `configs` too

pkg/cmd/apikeys/create/create.go
joshma: Yeah I was going crazy writing these print functions, I'm going to make that change thx

pkg/cmd/apikeys/apikeys.go
joshma: Ah yep I can take care of that later

PR #65, opened 2021-04-08
Add `auth info` command to retrieve current user and team

joshma:
$ airplane auth info
  Signed in as josh@airplane.dev
  Using team Airplane (ID: 1jq9NDeTcTmyqqXAjEsk7V0wXaV)

### DIFF
--- a/pkg/cmd/auth/info/info.go
+++ b/pkg/cmd/auth/info/info.go
@@ -11,6 +11,7 @@ import (
 
 var (
 	blue = color.New(color.FgHiBlue).SprintFunc()
+	gray = color.New(color.FgHiBlack).SprintFunc()
 )
 
 func New(c *cli.Config) *cobra.Command {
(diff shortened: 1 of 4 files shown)


pkg/cmd/auth/info/info.go
```diff
@@ -0,0 +1,46 @@
...
+	var client = c.Client
+
+	res, err := client.AuthInfo(ctx)
+	if err != nil {
+		return err
+	}
+
+	var userStr string
+	if res.User == nil {
+		userStr = gray("<no user>")
```
colinking, line 38: Maybe print `Not logged in. To log in, run: airplane auth login`?

colinking (approved):

pkg/cmd/auth/info/info.go
joshma: Welll I'm trying to have a distinction between "not logged in" and "no user" - the latter case _may_ be possible if we allow logging in via other means? Like TokenID right now.

PR #71, opened 2021-04-09
Stream remote build logs

poudels14:
This PR updates the `deploy --builder=remote` command to render build logs. Here's what it looks like:

image

This PR doesn't update the `--builder=local` command since that will require extra logic to parse Docker's output. We'll tackle that in a future PR.

While here, I pulled the color parsing logic into the `logging` package so it's easier to import new colors.


pkg/logger/colors.go
```diff
@@ -0,0 +1,12 @@
+package logger
+
+import "github.com/fatih/color"
+
+var (
```
joshma, line 5: ⭐

joshma (approved):

PR #72, opened 2021-04-10
Lock base image versions to digest SHAs

colinking:
We now have a `versions.json` file that is embedded in our CLI and contains a lookup table for identifying the image, tag, and digest to use for any given builder and user-supplied builder version (currently that's just Node, the rest are hardcoded since the major versions are stable enough that there is no relevant choice for customers to make).

We store this information in JSON in a centralized file (rather than split across our builders) since this gives us a format we can easily automate down the road to make upgrading versions easier.

While here, I also made the decision to a) standardize on Debian (buster) b) standardize on specifying minor versions. The former impacted Go and Deno, which previously were on Alpine. The latter is a no-op, considering we're locking on the digest, but it helps clarify what underlying version the digest maps to.


pkg/build/versions.go
```diff
@@ -0,0 +1,57 @@
...
+// images built for the linux/amd64 architecture.
+//
+// This lookup table is used to construct Dockerfiles that always
+// pull from the most-up-date version of the underlying base image
+// based on what we have cached in the Airplane registry for customers.
+type Versions map[string]map[string]Version
+
+type Version struct {
+	Image  string `json:"image"`
+	Tag    string `json:"tag"`
```
joshma, line 24: Assuming `Tag` isn't used but just kept here for future usage? We'll need it to update the digest later I suppose?

joshma (approved):
>  the rest are hardcoded since the major versions are stable enough that there is no relevant choice for customers to make

I think Python may want to add back `2` for some legacy customers, but I'm happy to not offer that until needed!

pkg/build/versions.go
colinking: Yep! The tag will be used to a) identify if there is a newer SHA for the same tag and b) to identify if there is a new minor semver release to upgrade to.

colinking: It's also just nice for readability 😄

PR #73, opened 2021-04-14
Add prompts to execute when missing params

joshma:
The new behavior:

- If you pass in `-- [params]`, the old behavior remains the same
- If you don't, we try to prompt - if `!CanPrompt()` (e.g. you're running this in CI), we print an error message unless there are no parameters (in which case we proceed)
- If CanPrompt(), we'll ask you for parameters and ask you to confirm before executing.

Note that I didn't include:
- Default values - TODO at some point, I would ideally like to share code from pkg/parameters here
- Upload support - for now, we let you skip uploads, so if a task has optional uploads you can skip it and still execute

image

Resolves AIR-613

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -269,7 +269,7 @@ type Values = map[string]interface{}
 // RunTaskRequest represents a run task request.
 type RunTaskRequest struct {
 	TaskID      string            `json:"taskID"`
-	Parameters  Values            `json:"params"`
+	ParamValues Values            `json:"params"`
 	Env         map[string]string `json:"env"`
 	Constraints Constraints       `json:"constraints"`
 }
--- a/pkg/cmd/tasks/execute/cli_inputs.go
+++ b/pkg/cmd/tasks/execute/cli_inputs.go
@@ -31,10 +31,8 @@ func promptForParamValues(client *api.Client, task api.Task, paramValues map[str
 			if !param.Constraints.Optional {
 				req = "*"
 			}
-			logger.Log("  %s%s %s (%s)", param.Name, req, param.Type, param.Slug)
-			if param.Desc != "" {
-				logger.Log("    %s", param.Desc)
-			}
+			logger.Log("  %s%s (%s)", param.Slug, req, param.Name)
+			logger.Log("    %s %s", gray(string(param.Type)), gray(param.Desc))
 		}
 		return errors.New("missing parameters")
 	}
--- a/pkg/cmd/tasks/execute/cli_inputs.go
+++ b/pkg/cmd/tasks/execute/cli_inputs.go
@@ -3,57 +3,98 @@ package execute
 
 import (
 	"fmt"
+	"os"
 	"reflect"
 	"strconv"
 	"time"
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/airplanedev/cli/pkg/api"
+	"github.com/airplanedev/cli/pkg/logger"
+	"github.com/airplanedev/cli/pkg/utils"
 	"github.com/pkg/errors"
 )
 
-func promptFromParam(param api.Parameter) (survey.Prompt, error) {
+// promptForParamValues attempts to prompt user for param values, setting them on `params`
+// If no TTY, errors unless there are no parameters
+// If TTY, prompts for parameters (if any) and asks user to confirm
+func promptForParamValues(client *api.Client, task api.Task, paramValues map[string]interface{}) error {
+	if !utils.CanPrompt() {
+		// Don't error if there are no params
+		if len(task.Parameters) == 0 {
+			return nil
+		}
+		// Otherwise, error since we have no params and no way to prompt for it
+		logger.Log("Parameters were not specified! Task has %d parameter(s):\n", len(task.Parameters))
+		for _, param := range task.Parameters {
+			var req string
+			if !param.Constraints.Optional {
+				req = "*"
+			}
+			logger.Log("  %s%s %s (%s)", param.Name, req, param.Type, param.Slug)
+			if param.Desc != "" {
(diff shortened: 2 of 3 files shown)


pkg/api/types.go
```diff
@@ -275,7 +275,7 @@ type Values = map[string]interface{}
 // RunTaskRequest represents a run task request.
 type RunTaskRequest struct {
 	TaskID      string            `json:"taskID"`
-	Parameters  Values            `json:"params"`
+	ParamValues Values            `json:"params"`
```
colinking, line 278: :awyis:

pkg/cmd/tasks/execute/cli_inputs.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+	if !confirmed {
+		return errors.New("user cancelled")
+	}
+	return nil
+}
+
+// promptForParam returns a survey.Prompt matching the param type
+func promptForParam(param api.Parameter) (survey.Prompt, error) {
+	message := fmt.Sprintf("%s (%s):", param.Name, param.Slug)
```
colinking, line 82: Should we hide the slug here? Feels only relevant to folks that are writing the task, not folks consuming it.

pkg/cmd/tasks/execute/cli_inputs.go
```diff
@@ -0,0 +1,198 @@
...
+	// TODO: support default values
```
colinking, line 83: Could we do this in this PR? 😇  Think we just need to convert the default value -> string yeah?

pkg/cmd/tasks/execute/cli_inputs.go
```diff
@@ -0,0 +1,198 @@
...
+	case api.TypeBoolean:
+		return parseBool(v)
+
+	case api.TypeInteger:
+		return strconv.Atoi(v)
+
+	case api.TypeFloat:
+		return strconv.ParseFloat(v, 64)
+
+	case api.TypeUpload:
```
colinking, line 119: Should we just show a warning (when optional) instead of even showing the prompt?

pkg/cmd/tasks/execute/cli_inputs.go
```diff
@@ -0,0 +1,198 @@
...
+			return nil
+		}
+		return nil
+	}
+}
+
+// Light wrapper around strconv.ParseBool with support for yes and no
+func parseBool(v string) (bool, error) {
+	switch v {
+	case "Yes", "yes":
```
colinking, line 191: maybe y/n too?

pkg/cmd/tasks/execute/cli_inputs.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/airplanedev/cli/pkg/api"
+	"github.com/airplanedev/cli/pkg/logger"
+	"github.com/airplanedev/cli/pkg/utils"
+	"github.com/pkg/errors"
+)
+
+// promptForParamValues attempts to prompt user for param values, setting them on `params`
+// If no TTY, errors unless there are no parameters
+// If TTY, prompts for parameters (if any) and asks user to confirm
+func promptForParamValues(client *api.Client, task api.Task, paramValues map[string]interface{}) error {
```
colinking, line 21: What do you think of supporting partial args? I imagine we could just pass in the provided args via `paramValues` and then `promptForParamValues` could just skip prompting for values already specified in `paramValues`.

For example, in this example I try to run a task with the `--name` flag that I know it needs, but the CLI realizes I also need a `--occupation` flag so instead of erroring, it just hops right into the relevant prompt:

```
airplane execute -f airplane.yml -- --name=Colin

? Name: Colin
? Occupation: ...
```

In general, over the long-term, I'd like the CLI to do this for normal flags too. If you are missing a required flag, we just prompt you for it instead of erroring.

colinking (approved):

pkg/cmd/tasks/execute/cli_inputs.go
joshma: I thought the slug was more important because these are devs executing the task, and so it should ideally match up to the `--flags` they'd otherwise pass in

pkg/cmd/tasks/execute/cli_inputs.go
joshma: I wanted to do this in a separate PR either way, in the interest of shipping things more quickly.

But also, this is a bit complicated to do right because I want to share the code w/ airport - I'll do a separate PR for this.

pkg/cmd/tasks/execute/cli_inputs.go
joshma: Yeah makes sense

pkg/cmd/tasks/execute/cli_inputs.go
joshma: Yeah sg

pkg/cmd/tasks/execute/cli_inputs.go
joshma: I thought about it but decided not to because it felt hard to handle optional args correctly. If you don't specify optional args, do we prompt? And if we don't, if you specify partial required args, do we prompt for both required and optional args?

I wasn't convinced there is an intuitive yet magical soln here.

pkg/cmd/tasks/execute/cli_inputs.go
colinking: Ah, if the goal is to associate slugs with CLI flags, maybe add the `--` prefix?

pkg/cmd/tasks/execute/cli_inputs.go
colinking: Oof, good point. Yeah, doesn't seem to be a good solution here.

PR #74, opened 2021-04-14
Show parameter table instead of task arguments

francescaleung:
Inserts a parameter table instead of showing task arguments.

```
$ airplane tasks list
     NAME     |    SLUG     | BUILDER |                          PARAMETERS                            
--------------+-------------+---------+----------------------------------------------------------------
  Echo Loop   | echo_loop   | manual  |     NAME    |   SLUG    |   TYPE   | REQUIRED |   DEFAULT      
              |             |         | ------------+-----------+----------+----------+--------------  
              |             |         |   Count     | count     | integer  | true     |          25    
              |             |         |   text      | text      | string   | true     | hello world    
              |             |         |   long text | long_text | string   | false    |                
              |             |         |   dry run   | dry_run   | boolean  | true     |                
              |             |         |   since     | since     | datetime | true     |                
              |             |         |                                                                
  Hello World | hello_world | manual  |
```

### DIFF
--- a/pkg/print/table.go
+++ b/pkg/print/table.go
@@ -3,6 +3,7 @@ package print
 import (
 	"encoding/json"
 	"strconv"
+	"strings"
 
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


pkg/print/table.go
```diff
@@ -41,20 +42,43 @@ func (t Table) apiKeys(apiKeys []api.APIKey) {
...
 	for _, t := range tasks {
 		var builder = t.Builder
-
 		if builder == "" {
 			builder = "manual"
 		}
 
+		var parametersStr string
+		if len(t.Parameters) > 0 {
+			parametersTableString := &strings.Builder{}
```
joshma, line 55: minor but for go, when there's no great variable name i'd prefer to use something short, like `b` for the builder or `pts` or `ptb` or `ptBuilder`

pkg/print/table.go
```diff
@@ -41,20 +42,43 @@ func (t Table) apiKeys(apiKeys []api.APIKey) {
...
+			ptw := tablewriter.NewWriter(parametersTableString)
+			ptw.SetBorder(false)
+			ptw.SetHeader([]string{"name", "slug", "type", "required", "default"})
+			for _, p := range t.Parameters {
+				var defaultStr = fmt.Sprintf("%v", p.Default)
```
joshma, line 60: NB this won't always work when Default is something more exotic like an upload ID, and it's OK but not great for dates, etc. Filed AIR-760 to track for later, `%v` seems OK for now.

And also, minor nit:

```suggestion
				defaultStr := fmt.Sprintf("%v", p.Default)
```

(@colinking would be nice to have lint enforce consistency on short decl over vars :) )

joshma (approved):

PR #76, opened 2021-04-14
Fix logger colors to support format strings

joshma:
You can now call `logger.Gray("hello %s", "world")` properly.


colinking (approved):

joshma:
@colinking I couldn't get the linting thing to work w/o creating a function per color, it seems sort of overkill - going to merge this w/o it for now but could be convinced to fix it more. :)

PR #77, opened 2021-04-14
Add prompt to create missing configs when running `deploy`

joshma:
This refactors out and re-uses some of `configs set` code to ask the
user to enter in a config value if they try to deploy a task that is
missing configs.

This helps with creating new tasks from e.g. examples (or just when you
want to create a new config). You can run `deploy` and create the
missing configs then.

Resolves AIR-696

### DIFF
--- a/pkg/cmd/tasks/deploy/deploy.go
+++ b/pkg/cmd/tasks/deploy/deploy.go
@@ -86,7 +86,7 @@ func run(ctx context.Context, cfg config) error {
 	task, err := client.GetTask(ctx, def.Slug)
 	if err == nil {
 		// This task already exists, so we update it:
-		logger.Log("  Updating task...")
+		logger.Log("Updating task...")
 		res, err := client.UpdateTask(ctx, api.UpdateTaskRequest{
 			Slug:           def.Slug,
 			Name:           def.Name,
@@ -111,7 +111,7 @@ func run(ctx context.Context, cfg config) error {
 		taskRevisionID = res.TaskRevisionID
 	} else if aerr, ok := err.(api.Error); ok && aerr.Code == 404 {
 		// A task with this slug does not exist, so we should create one.
-		logger.Log("  Creating task...")
+		logger.Log("Creating task...")
 		res, err := client.CreateTask(ctx, api.CreateTaskRequest{
 			Slug:           def.Slug,
 			Name:           def.Name,
@@ -155,7 +155,7 @@ func run(ctx context.Context, cfg config) error {
 		}
 	}
 
-	logger.Log("  Done!")
+	logger.Log("Done!")
 	cmd := fmt.Sprintf("airplane execute %s", def.Slug)
 	if len(def.Parameters) > 0 {
 		cmd += " -- [parameters]"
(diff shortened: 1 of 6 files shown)


pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -77,12 +77,16 @@ func run(ctx context.Context, cfg config) error {
...
+		return err
+	}
+
 	var taskID string
 	var taskRevisionID string
 	task, err := client.GetTask(ctx, def.Slug)
 	if err == nil {
 		// This task already exists, so we update it:
-		logger.Log("Updating...")
+		logger.Log("  Updating task...")
```
colinking, line 89: Are we bringing back the spaces? I just removed them the other day lmao

pkg/cmd/tasks/deploy/predeploy.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/airplanedev/cli/pkg/api"
+	"github.com/airplanedev/cli/pkg/configs"
+	"github.com/airplanedev/cli/pkg/logger"
+	"github.com/airplanedev/cli/pkg/taskdir"
+	"github.com/airplanedev/cli/pkg/utils"
+	"github.com/pkg/errors"
+)
+
+// predeployHooks runs before a task is deployed and can e.g. prompt the user to specify or fix things
+func predeployHooks(ctx context.Context, client *api.Client, def taskdir.Definition) error {
```
colinking, line 18: nit: but hooks imply to me that this is user-specified behavior (f.e. git hooks) -- maybe there's a different name we could use? Maybe just  `ensureConfigsExist` since n=1?

pkg/cmd/tasks/deploy/predeploy.go
```diff
@@ -0,0 +1,79 @@
...
+	switch err := errors.Cause(err).(type) {
+	case api.Error:
+		if err.Code != 404 {
+			return err
+		}
+		if !utils.CanPrompt() {
+			return errors.Errorf("config %s does not exist", configName)
+		}
+		logger.Log("Your task definition references config %s, which does not exist", logger.Bold(configName))
+		confirmed, errc := utils.Confirm("Create it now?")
```
colinking, line 48: Any reason not to use `err` here?

```suggestion
		confirmed, err := utils.Confirm("Create it now?")
```

colinking (approved):
This looks great 😍

pkg/cmd/tasks/deploy/deploy.go
joshma: Ah I just did this to be consistent with the local builder output tbh, I should've probably just gone the other way. I'll do that instead.

pkg/cmd/tasks/deploy/predeploy.go
joshma: Didn't want to shadow `err` above...

pkg/cmd/tasks/deploy/predeploy.go
joshma: Sounds good

pkg/cmd/tasks/deploy/predeploy.go
colinking: Does this shadow err from the `return` below? Thought this would be inside the case statement's scope.

joshma: `err` at this point is `api.Error` from the switch statement, so you get this error when comparing to `nil`:

image

PR #78, opened 2021-04-15
Add improvements to execute prompts

joshma:
This is a follow up of #73:

- Render slugs in prompts with `--` in front and in gray
- Refactor out pkg/params/ for input to API value conversions, and use that to
  pre-fill default param values + when listing task parameter defaults
- Instead of prompting for file inputs, print a yellow warning message and skip
  them


pkg/print/table.go
```diff
@@ -58,9 +59,9 @@ func (t Table) tasks(tasks []api.Task) {
 			ptw.SetBorder(false)
 			ptw.SetHeader([]string{"name", "slug", "type", "required", "default"})
 			for _, p := range t.Parameters {
-				defaultStr := fmt.Sprintf("%v", p.Default)
-				if p.Default == nil {
-					defaultStr = ""
+				defaultStr, err := params.APIValueToInput(p, p.Default)
+				if err != nil {
+					defaultStr = "?"
```
joshma, line 64: LMK if you have better ideas for this :)

pkg/params/params.go
```diff
@@ -0,0 +1,168 @@
...
+		return nil
+	}
+
+	switch param.Type {
+	case api.TypeString:
+		return nil
+
+	case api.TypeBoolean:
+		if _, err := ParseBool(in); err != nil {
+			return errors.New("expected yes, no, true, false, 1 or 0")
```
colinking, line 33: Should we use sentence case + periods here, since these errors are shown to users?

pkg/print/table.go
colinking: Maybe something like `<unknown>`? So it doesn't look like the string `"?"`

colinking (approved):

pkg/params/params.go
joshma: Hm so it shows up w/ a prefix, e.g. here's an invalid number:

> X Sorry, your reply was invalid: invalid number

So I didn't want to capitalize it, but then I didn't want to add a period either.

PR #80, opened 2021-04-15
Add support for `.dockerignore` files when constructing uploads

colinking:
If you have a `.dockerignore` in the root of your task directory, we will now respect it when creating an upload of your task directory.

Note this change is only relevant to how we upload your task directory, not the docker build process itself. When performing a local build, the docker CLI will use the `.dockerignore` like normal and when performing a remote build, the files ignored by the `.dockerignore` simply will not exist since they were not included in the upload.

This PR also adds defaults `.dockerignore` files for each builder we support so that customers only need to add their own `.dockerignore` if they want custom behavior.

--

This PR uses a fork of `mholt/archiver` to add support for filtering archives: https://github.com/airplanedev/archiver/pull/1

We might want to upstream this at some point, but we would likely need to expand support across all other archive formats first.


pkg/build/remote.go
```diff
@@ -62,14 +67,121 @@ func archiveTaskDir(dir taskdir.TaskDirectory, archivePath string) error {
...
+		return nil, errors.Wrap(err, "reading .dockerignore")
+	}
+
+	if len(excludes) > 0 {
+		return excludes, nil
+	}
+
+	// If a .dockerignore was not provided, use a default based on the builder.
+	defaultExcludes := []string{
+		".git",
```
joshma, line 151: 🤔  thoughts on always excluding .git? could lead to unexpected behavior I suppose

it's a bit weird if you add a .dockerignore you could end up including more files than expected, so we'd want to make that clear...

joshma (approved):

pkg/build/remote.go
colinking: Ah, if you're trying to do `git` operations on the repo? That's fair, though I'd be happy with documenting this behavior and explaining that you can override it by adding your own `.dockerignore`.

As an example, the `.git` folder is 50MB in our main mono repo, so feels like we should default to the 90%+ case here and see if folks don't like that

joshma: Yeah I'm down to try this route and see how it feels

PR #82, opened 2021-04-15
Query API for a unique slug when init'ing tasks

francescaleung:


joshma:
LGTM but @colinking can I get your eyes on this one?

pkg/cmd/tasks/initcmd/sample.go
```diff
@@ -34,17 +36,12 @@ func initFromSample(cfg config) error {
 		return err
 	}
 
-	// TODO: Only prompt the user for a slug if the default slug is not unique.
-	defaultSlug := def.Slug
-	if defaultSlug == "" {
-		defaultSlug = utils.MakeSlug(def.Name)
-	}
-	slug, err := utils.PickSlug(defaultSlug)
+	r, err := client.GetUniqueSlug(ctx, def.Name)
```
colinking, line 39: wdyt of passing `def.Slug` in here, too? And the API would prefer `def.Slug` if unique, but otherwise would generate a slug from `name` and then fallback on a suffix if that still isn't unique.

That way we increase the probability of not having to change the task definition below (which causes some formatting issues unfortunately :( )

colinking (approved):

colinking (approved):
awyis

PR #90, opened 2021-04-18
Add support to `auth info` for `-o` flag

joshma:
This enables `airplane auth info -o json | jq .team.id`:

```
$ airplane auth info -o json
{"user":{"id":"prof_01EQV9VY8D2NEK6X2BFRXC2BM8","email":"josh@airplane.dev"},"team":{"id":"1jq9NDeTcTmyqqXAjEsk7V0wXaV","name":"Airplane"}}

$ airplane auth info -o json | jq -r .team.id
1jq9NDeTcTmyqqXAjEsk7V0wXaV
```

### DIFF
--- a/pkg/cmd/auth/info/info.go
+++ b/pkg/cmd/auth/info/info.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/airplanedev/cli/pkg/cli"
 	"github.com/airplanedev/cli/pkg/logger"
+	"github.com/airplanedev/cli/pkg/print"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/cmd/auth/info/info.go
```diff
@@ -27,14 +28,21 @@ func run(ctx context.Context, c *cli.Config) error {
 		return err
 	}
 
-	var userStr string
-	if res.User == nil {
-		userStr = logger.Gray("<no user>")
-	} else {
-		userStr = res.User.Email
+	switch f := print.DefaultFormatter.(type) {
```
colinking, line 31: What do you think of a `print` helper like so, where if the encoder is JSON/YAML then it'll just handle that itself. But if the encoder is tabular, then it would call your func?

```go
print.Print(res, func() {
  var userStr string
  if res.User == nil {
    userStr = logger.Gray("<no user>")
  } else {
    userStr = res.User.Email
  }
  logger.Log("  Signed in as %s", logger.Blue(userStr))
  logger.Log("  Using team %s (ID: %s)", logger.Blue(res.Team.Name), res.Team.ID)
})
```

colinking (approved):

pkg/cmd/auth/info/info.go
joshma: Yeah I like it, should be useful. I'll do that.

PR #92, opened 2021-04-20
Add buildCommand and buildDir to Node builderConfig

joshma:
- `buildCommand` is something like `npm run build` that you want run
  after dependencies are installed
- `buildDir` is the output directory of build - this is where we'll cd
  into and wher we'll find `entrypoint`

So if your build outputs to `.airplane/main.js`, you'll want
`buildDir=.airplane` and `entrypoint=main.js`.

### DIFF
--- a/pkg/build/node.go
+++ b/pkg/build/node.go
@@ -75,7 +75,7 @@ func node(root string, args Args) (string, error) {
 FROM {{ .Base }}
 
 WORKDIR {{ .Workdir }}
-COPY . /airplane
+COPY . {{ .Workdir }}
 {{ range .Commands }}
 RUN {{ . }}
 {{ end }}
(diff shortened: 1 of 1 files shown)


pkg/build/node.go
```diff
@@ -36,29 +45,42 @@ func node(root string, args Args) (string, error) {
...
 	}
+	entrypoint = path.Join(buildWorkdir, entrypoint)
 
 	// Dockerfile template.
 	t, err := template.New("node").Parse(`
 FROM {{ .Base }}
 
-WORKDIR /airplane
+WORKDIR {{ .Workdir }}
 COPY . /airplane
```
colinking, line 78: Should this be `COPY . .` or `COPY . {{ .Workdir }}`?

colinking (approved):

pkg/build/node.go
joshma: Oof thanks

PR #94, opened 2021-04-21
Add support for buildEnv and NPM auth in local builder

joshma:
This is a first iteration that adds support for hardcoded env values (i.e. not config values) - setting BUILD_NPM_AUTH or BUILD_NPM_TOKEN will result in a `.npmrc` file created.

A future change will add support for fetching a full buildEnv from API instead.

You can test this with a task using `@airplane/npm-private-pkg` and env like:

```
env:
  BUILD_NPM_TOKEN: <get this from npm website>
```

### DIFF
--- a/pkg/build/local.go
+++ b/pkg/build/local.go
@@ -2,7 +2,6 @@ package build
 
 import (
 	"context"
-	"strings"
 
 	"github.com/airplanedev/cli/pkg/api"
 	"github.com/airplanedev/cli/pkg/logger"
@@ -19,7 +18,7 @@ func Local(ctx context.Context, client *api.Client, dir taskdir.TaskDirectory, d
 	buildEnv := make(map[string]string)
 	// TODO: currently, we just read non-config values from env. Should ask API for full BuildEnv instead.
 	for k, v := range def.Env {
-		if strings.HasPrefix(k, "BUILD_") && v.Value != nil {
+		if v.Value != nil {
 			buildEnv[k] = *v.Value
 		}
 	}
--- a/pkg/build/local.go
+++ b/pkg/build/local.go
@@ -2,6 +2,7 @@ package build
 
 import (
 	"context"
+	"strings"
 
 	"github.com/airplanedev/cli/pkg/api"
 	"github.com/airplanedev/cli/pkg/logger"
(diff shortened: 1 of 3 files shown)


joshma:
Follow up PR for configs: https://github.com/airplanedev/cli/pull/95 (I'll put in review once the first set of PRs merge)

pkg/build/node.go
```diff
@@ -75,6 +75,13 @@ func node(root string, args Args) (string, error) {
 FROM {{ .Base }}
 
 WORKDIR {{ .Workdir }}
+
+# Support setting BUILD_NPM_RC or BUILD_NPM_TOKEN to configure private registry auth
+ARG BUILD_NPM_RC
+ARG BUILD_NPM_TOKEN
```
colinking, line 81: So does this just default to an empty string if the build-arg is not set?

pkg/build/node.go
```diff
@@ -75,6 +75,13 @@ func node(root string, args Args) (string, error) {
...
+RUN [ -z "${BUILD_NPM_RC}" ] || "${BUILD_NPM_RC}" > .npmrc
```
colinking, line 82: Does this work with `double quotes` inside of `BUILD_NPM_RC`?

pkg/build/local.go
```diff
@@ -15,14 +16,23 @@ func Local(ctx context.Context, client *api.Client, dir taskdir.TaskDirectory, d
 		return errors.Wrap(err, "getting registry token")
 	}
 
-	b, err := New(Config{
+	buildEnv := make(map[string]string)
+	// TODO: currently, we just read non-config values from env. Should ask API for full BuildEnv instead.
+	for k, v := range def.Env {
+		if strings.HasPrefix(k, "BUILD_") && v.Value != nil {
```
colinking, line 22: A `BUILD_` prefix sounds good to me -- it makes it very clear which variables are available. If we do that, should we exclude `BUILD_` variables at runtime? Alternatively, what do you think of having a `build: true` variable in env vars? f.e. if you need an env var without the `BUILD_` prefix.

OTOH, curious if there's a case to be made to share all env vars during build-time instead of a subset. Is that hard from a technical perspective?

colinking (approved):

pkg/build/node.go
joshma: Yeah exactly - hence `-z` checking for it

pkg/build/node.go
joshma: Yep! `"${x}"` AFAIK is the safest way to expand a var - it'll be one argument written to e.g. `.npmrc`

But now that I'm staring at this line, it's missing `echo` 🤦 will fix.

pkg/build/local.go
joshma: Hm I want to avoid more config when possible. I'm not even sure if we should strip values during build actually, so I'm leaning towards your latter suggestion. It's not hard to do right, we already have all the values? Let's...do that? (and not filter to BUILD_*?)

joshma: Note for posterity - going to support all env vars, not just BUILD_

PR #97, opened 2021-04-22
Perform remote builds by default

colinking:
We've made a number of improvements to our hosted builder, so we're now ready for that to replace the local builder as the default.

This PR won't merge until the relevant backend changes are live.


joshma (approved):
yesssssss

PR #98, opened 2021-04-22
Improve remote builder logs

colinking:
This PR updates the logs we produce during a remote build to provide more detailed context on what is happening.

An example remote build now looks like this:

image


joshma (approved):
Just hot off the feedback from Air on the logs, thinking about making these logs less detailed - could you look and see if maybe half of these lines can be deleted instead? Or moved to debug?

colinking:
@joshma Cut out a bunch of logs. Also marked the Dockerfile commands as builder / gray:

image

joshma:
🤩

PR #105, opened 2021-04-27
Change YAML file format

francescaleung:
Before:
```
builder: python
builderConfig:
  entrypoint: main.py
```

After:
```
python:
  entrypoint: main.py
```

This PR adds logic for understanding multiple yaml file versions; the latest version will be tried first, & then previous versions. Only errors from the newest version will be shown to the user. This doesn't change the output of `tasks get`; communications with the API server are not changing, either. When initializing from a sample, if the sample's task definition is an older version, the new task definition will be upgraded in the newest version.

Bonus: `ResourceLimits` -> `ResourceRequests`.

### DIFF
--- a/pkg/cmd/tasks/initcmd/sample.go
+++ b/pkg/cmd/tasks/initcmd/sample.go
@@ -57,27 +57,30 @@ func initFromSample(ctx context.Context, cfg config) error {
 		outputdir = path.Base(path.Dir(dir.DefinitionPath()))
 	}
 
-	// Rename the sample definition to match the filename in the `-f` argument.
-	// This is done to maintain semantic consistency with the other kinds of
-	// init, but is not strictly necessary.
-	defname := path.Base(dir.DefinitionPath())
-	if cfg.file != "" && defname != path.Base(cfg.file) {
-		defname = path.Base(cfg.file)
-		if err := os.Rename(
-			dir.DefinitionPath(),
-			path.Join(path.Dir(dir.DefinitionPath()), defname),
-		); err != nil {
-			return errors.Wrap(err, "renaming task definitino")
-		}
-	}
-
 	// Copy the sample code from the temporary directory into the user's
 	// local directory.
 	if err := copy.Copy(dir.DefinitionRootPath(), outputdir); err != nil {
 		return errors.Wrap(err, "copying sample directory")
 	}
 
+	// Remove the original def file.
+	if err := os.Remove(path.Join(outputdir, path.Base(dir.DefinitionPath()))); err != nil {
+		return errors.Wrap(err, "removing original def")
+	}
+
+	// Write the new definition to match the filename in the `-f` argument.
+	// This is done to maintain semantic consistency with the other kinds of
+	// init, but is not strictly necessary.
+	defname := path.Base(dir.DefinitionPath())
+	if cfg.file != "" && defname != path.Base(cfg.file) {
+		defname = path.Base(cfg.file)
+	}
 	file := path.Join(outputdir, defname)
+	oDir, err := taskdir.New(file)
+	if oDir.WriteDefinition(def); err != nil {
+		return errors.Wrap(err, "rewriting task definition")
+	}
+
 	logger.Log(`
 An Airplane task definition for '%s' has been created!
 
--- a/pkg/cmd/tasks/initcmd/sample.go
+++ b/pkg/cmd/tasks/initcmd/sample.go
@@ -41,9 +41,7 @@ func initFromSample(ctx context.Context, cfg config) error {
 		return errors.Wrap(err, "getting unique slug")
 	}
 	if r.Slug != def.Slug {
-		if err := dir.WriteSlug(r.Slug); err != nil {
-			return err
-		}
+		def.Slug = r.Slug
 	}
 
 	var outputdir string
--- a/pkg/cmd/tasks/initcmd/task.go
+++ b/pkg/cmd/tasks/initcmd/task.go
@@ -53,11 +53,11 @@ func initFromTask(ctx context.Context, cfg config) error {
 		Constraints:    task.Constraints,
 		Env:            task.Env,
 		ResourceLimits: task.ResourceLimits,
-		Kind:           task.Kind,
-		KindOptions:    task.KindOptions,
 		Repo:           task.Repo,
 		Timeout:        task.Timeout,
 	}
+	def.FromKindAndOptions(task.Kind, task.KindOptions)
+
 	// Only show the image field if this is a manual builder
 	if task.Kind == "manual" {
 		def.Image = task.Image
--- a/pkg/cmd/tasks/initcmd/task.go
+++ b/pkg/cmd/tasks/initcmd/task.go
@@ -9,6 +9,7 @@ import (
 	"github.com/airplanedev/cli/pkg/api"
 	"github.com/airplanedev/cli/pkg/logger"
 	"github.com/airplanedev/cli/pkg/taskdir"
+	"github.com/airplanedev/cli/pkg/taskdir/definitions"
 	"github.com/pkg/errors"
 )
 
@@ -42,7 +43,7 @@ func initFromTask(ctx context.Context, cfg config) error {
 		return errors.Wrap(err, "getting unique slug")
 	}
 
-	def := taskdir.Definition{
+	def := definitions.Definition{
 		Slug:           r.Slug,
 		Name:           task.Name,
 		Description:    task.Description,
--- a/pkg/cmd/tasks/initcmd/task.go
+++ b/pkg/cmd/tasks/initcmd/task.go
@@ -43,25 +43,13 @@ func initFromTask(ctx context.Context, cfg config) error {
 		return errors.Wrap(err, "getting unique slug")
 	}
 
-	def := definitions.Definition{
-		Slug:           r.Slug,
-		Name:           task.Name,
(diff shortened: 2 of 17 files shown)


francescaleung:
Open to a different naming scheme for task definitions (e.g., datebased).

pkg/taskdir/definitions/def_0_2.go
```diff
@@ -0,0 +1,66 @@
...
+	Slug           string             `yaml:"slug"`
+	Name           string             `yaml:"name"`
+	Description    string             `yaml:"description,omitempty"`
+	Image          string             `yaml:"image,omitempty"`
+	Command        []string           `yaml:"command,omitempty"`
+	Arguments      []string           `yaml:"arguments,omitempty"`
+	Parameters     api.Parameters     `yaml:"parameters,omitempty"`
+	Constraints    api.RunConstraints `yaml:"constraints,omitempty"`
+	Env            api.TaskEnv        `yaml:"env,omitempty"`
+	ResourceLimits api.ResourceLimits `yaml:"resourceLimits,omitempty"`
```
colinking, line 17: Ah looks like we had a bug here -- s/ResourceLimits/ResourceRequests

https://github.com/airplanedev/airport/blob/be66dd653fea554300a65df5d2385f4b6f7d223d/pkg/runs/runs.go#L564

pkg/taskdir/definitions/def_0_2.go
```diff
@@ -0,0 +1,66 @@
...
+	// in the task's Docker image.
+	//
+	// If not set, defaults to "." (in other words, the parent directory of this task definition).
+	//
+	// This field is ignored when using the pre-built image builder (aka "manual").
+	Root string `yaml:"root,omitempty"`
+}
+
+type ManualDefinition struct {
+	Config map[string]string
```
colinking, line 39: The config we would need for the manual "builder" mirrors what we have in the UI:

image


```suggestion
	Image string `yaml:"image"`
	Command []string `yaml:"command,omitempty"`
```

pkg/taskdir/definitions/def_0_2.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"github.com/airplanedev/cli/pkg/api"
+)
+
+type Definition_0_2 struct {
+	Slug           string             `yaml:"slug"`
+	Name           string             `yaml:"name"`
+	Description    string             `yaml:"description,omitempty"`
+	Image          string             `yaml:"image,omitempty"`
```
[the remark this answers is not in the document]
colinking: We can also remove `command`, too

pkg/taskdir/definitions/def_0_2.go
colinking: Also we should probably rename `Entrypoint` in the UI here to `Command` to mirror Docker terminology.

colinking: For the manual builder, we don't actually set anything in `builder` or `builderConfig`. Manual is what all tasks looked like before we introduced builders -- you would bring your own pre-built Docker image that we would run. So the config instead lives in the `image`/`command` top-level fields, but I like how your PR will hide that distinction in the file format :)

pkg/taskdir/definitions/definitions.go
```diff
@@ -0,0 +1,145 @@
...
+	if err := yaml.Unmarshal(buf, &def); err != nil {
+		return Definition{}, errors.Wrap(err, "unmarshalling task definition")
+	}
+
+	return def, nil
+}
+
+func tryOlderDefinitions(buf []byte) (Definition, error) {
+	var err error
+	if err = validateYAML(buf, Definition_0_1{}); err == nil {
```
colinking, line 137: nit: would recommend handling `err != nil` here to flatten this block

pkg/cmd/tasks/initcmd/sample.go
```diff
@@ -57,27 +57,30 @@ func initFromSample(ctx context.Context, cfg config) error {
...
+	// Copy the sample code from the temporary directory into the user's
+	// local directory.
+	if err := copy.Copy(dir.DefinitionRootPath(), outputdir); err != nil {
+		return errors.Wrap(err, "copying sample directory")
+	}
+
+	// Remove the original def file.
+	if err := os.Remove(path.Join(outputdir, path.Base(dir.DefinitionPath()))); err != nil {
+		return errors.Wrap(err, "removing original def")
+	}
```
colinking, line 69: What's the context on these changes? Is this just to rewrite the definition files from the examples into the new format?

pkg/taskdir/definitions/definitions.go
```diff
@@ -0,0 +1,145 @@
...
+)
+
+// Definition represents a YAML-based task definition that can be used to create
+// or update Airplane tasks.
+//
+// Note this is the subset of fields that can be represented with a revision,
+// and therefore isolated to a specific environment.
+type Definition Definition_0_2
+
+func (this Definition) FromKindAndOptions(kind string, options api.KindOptions) error {
```
colinking, line 18: What do you think of adding a constructor instead?

Something like:

```go
func NewDefinitionFromTask(task *api.Task) (Definition, error) {
// ...
}
```

pkg/taskdir/definitions/def_0_2.go
colinking: We would then need to pull out `image/command` from the manual config here: https://github.com/airplanedev/cli/pull/105/files#diff-ee891759dd8b9f7107c066f2b0545cd4e07fc73403cbc252f482a0c5c466b53e

Noting that `image` and `command` can/should be empty strings for the other builders (the former defaults to the Airplane Registry and the latter defaults at runtime to the command from the Docker image).

colinking (changes requested):
🔥 This looks great. Just some comments about idiosyncrasies with the manual builder!

pkg/taskdir/definitions/def_0_2.go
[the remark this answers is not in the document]
francescaleung: What about `arguments`?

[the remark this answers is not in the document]
colinking: Arguments can stay, those apply to all builders! The command tends to be builder specific (f.e. run this compiled Go file) so we defer to the builder for that, except when a user is bringing their own image (via the manual builder)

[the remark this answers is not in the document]
francescaleung: nope, that belongs at the toplevel.

pkg/cmd/tasks/initcmd/sample.go
francescaleung: I think so. I have a vague recollection of there being other reasons for this, but the main thing is to rewrite the def files. Not entirely sure this is wanted/needed, though, now that I think about it.

pkg/taskdir/definitions/definitions.go
francescaleung: yea, this was written so as to accommodate multiple older defs. Once we have a 0.3 def, you can copy the the entire if block, paste it above this one, s/Definition_0_1/Definition_0_2/, & you're good to go. Thinking ahead. Maybe too much.

pkg/taskdir/definitions/definitions.go
francescaleung: fancy, & I like it. Done.

pkg/cmd/tasks/initcmd/task.go
```diff
@@ -43,25 +43,13 @@ func initFromTask(ctx context.Context, cfg config) error {
...
-		Description:    task.Description,
-		Command:        task.Command,
-		Arguments:      task.Arguments,
-		Parameters:     task.Parameters,
-		Constraints:    task.Constraints,
-		Env:            task.Env,
-		ResourceLimits: task.ResourceLimits,
-		Repo:           task.Repo,
-		Timeout:        task.Timeout,
+	def, err := definitions.NewDefinitionFromTask(task)
```
colinking, line 46: 🔥

pkg/taskdir/definitions/definitions.go
```diff
@@ -15,60 +15,74 @@ import (
...
 func (this Definition) GetKindAndOptions() (string, api.KindOptions, error) {
 	if this.Deno != nil {
 		return "deno", api.KindOptions{
 			"entrypoint": this.Deno.Entrypoint,
 		}, nil
-	} else if this.Docker != nil {
+	} else if this.Dockerfile != nil {
 		return "docker", api.KindOptions{
-			"dockerfile": this.Docker.Dockerfile,
+			"dockerfile": this.Dockerfile.Dockerfile,
```
colinking, line 85: ooh yeah this is odd hmm, maybe we rename to `dockerfile.path` in the task definition format?

colinking (approved):
🙌🏻

joshma:
@francescaleung I didn't follow this PR closely and am going to trust you two on this. However, can you summarize the important changes in the PR description, but for my reference and for posterity?

pkg/taskdir/definitions/definitions.go
francescaleung: It does feel a little odd, but it's a convention that works with everything else. I thought about having something like `def.PythonDef.Entrypoint`, in which case this would be `def.DockerfileDef.Dockerfile`, but didn't think it was totally necessary. I would prefer a convention applied evenly instead of having random exceptions.

PR #110, opened 2021-04-28
Hide exec instructions on remote build failure

colinking:
We now bubble up an error that will cause the CLI to exit with a non-zero exit code.


joshma (approved):
Just noting that the error message is duplicated, but idk if that's worth fixing right now.

```
Updating task...
[build] Uploading 2.8 kB build archive...
[build] Waiting for builder...
[build] Starting build...

Build failed

Error: Build failed
```

Should error messages be capitalized? Will keep in mind.

PR #111, opened 2021-04-29
Only allow one definition on a task

francescaleung:
Error out if a task definition has multiple task types or no task types.

### DIFF
--- a/pkg/taskdir/definitions/definitions.go
+++ b/pkg/taskdir/definitions/definitions.go
@@ -135,7 +135,7 @@ func (this Definition) Validate() (Definition, error) {
 		return this, errors.New("No task type defined")
 	}
 	if len(defs) > 1 {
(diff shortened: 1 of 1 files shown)


colinking (approved):

pkg/taskdir/definitions/definitions.go
```diff
@@ -110,6 +110,33 @@ func (this Definition) Validate() (Definition, error) {
...
+	}
+	if this.Python != nil {
+		numDefs++
+	}
+
+	if numDefs == 0 {
+		return this, errors.New("No task type defined")
+	}
+	if numDefs > 1 {
+		return this, errors.New("Too many task types defined")
```
colinking, line 137: Maybe we could indicate which types were defined here?

```
Too many task kinds defined: only one of node, or deno is expected
```

PR #112, opened 2021-04-29
Check task kind against an explicit list for determining what needs building

francescaleung:
This pulls out the `NeedsBuilding` function as an explicit check & checks it against a list of task kinds that require building (rather than task kinds that _don't_ require building). Also narrows down typing of `task.Kind`. Some overlap with `BuilderName`, but they are definitely two separate concepts.

### DIFF
--- a/pkg/build/builder.go
+++ b/pkg/build/builder.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"unicode"
 
+	"github.com/airplanedev/cli/pkg/api"
 	"github.com/airplanedev/cli/pkg/logger"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/client"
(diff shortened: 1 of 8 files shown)


pkg/build/builder.go
```diff
@@ -313,6 +314,15 @@ const (
...
 )
 
+func NeedsBuilding(kind api.TaskKind) bool {
+	switch BuilderName(kind) {
+	case BuilderNameGo, BuilderNameDeno, BuilderNamePython, BuilderNameNode, BuilderNameDocker:
+		return true
+	default:
+		return false
+	}
+}
```
colinking, line 324: 🔥

pkg/taskdir/definitions/definitions.go
```diff
@@ -29,39 +29,39 @@ func NewDefinitionFromTask(task api.Task) (Definition, error) {
...
+	} else if task.Kind == api.TaskKindNode {
 		node := NodeDefinition{
 			Entrypoint:  task.KindOptions["entrypoint"],
 			Language:    task.KindOptions["language"],
 			NodeVersion: task.KindOptions["nodeVersion"],
 		}
 		def.Node = &node
 
-	} else if task.Kind == "python" {
+	} else if task.Kind == api.TaskKindPython {
```
colinking, line 58: 🔥

colinking (approved):
nice cleanup 🔥

PR #113, opened 2021-04-30
Add --debug flag to deploy & execute for more logs

francescaleung:
Pulls down more logs from the api if run with `--debug`. Defaults to showing info-level logs.

### DIFF
--- a/pkg/api/client.go
+++ b/pkg/api/client.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/airplanedev/cli/pkg/logger"
 	"github.com/airplanedev/cli/pkg/version"
 
 	"github.com/pkg/errors"
@@ -144,12 +145,12 @@ func (c Client) RunTask(ctx context.Context, req RunTaskRequest) (res RunTaskRes
 }
 
 // Watcher runs a task with the given arguments and returns a run watcher.
-func (c Client) Watcher(ctx context.Context, req RunTaskRequest, debug bool) (*Watcher, error) {
+func (c Client) Watcher(ctx context.Context, req RunTaskRequest) (*Watcher, error) {
 	resp, err := c.RunTask(ctx, req)
 	if err != nil {
 		return nil, err
 	}
-	return newWatcher(ctx, c, resp.RunID, debug), nil
+	return newWatcher(ctx, c, resp.RunID), nil
 }
 
 // GetRun returns a run by id.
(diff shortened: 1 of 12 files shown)


pkg/build/remote.go
```diff
@@ -21,7 +21,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-func Remote(ctx context.Context, dir taskdir.TaskDirectory, client *api.Client, taskRevisionID string) error {
+func Remote(ctx context.Context, debug bool, dir taskdir.TaskDirectory, client *api.Client, taskRevisionID string) error {
```
joshma, line 24: can we put debug last in the arg list? it seems less important

pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -42,6 +43,7 @@ func New(c *cli.Config) *cobra.Command {
 
 	cmd.Flags().StringVarP(&cfg.file, "file", "f", "", "Path to a task definition file.")
 	cmd.Flags().StringVar(&cfg.builder, "builder", string(build.BuilderKindRemote), "Where to build the task's Docker image. Accepts: [local, remote]")
+	cmd.Flags().BoolVarP(&cfg.debug, "debug", "d", false, "True to output more logs.")
```
joshma, line 46: In root.go, we already have a `PersistentFlags` for `cfg.DebugMode`, I think this should re-use that instead of defining a new flag here.

pkg/api/client.go
```diff
@@ -160,13 +160,17 @@ func (c Client) GetRun(ctx context.Context, id string) (res GetRunResponse, err
 }
 
 // GetLogs returns the logs by runID and since timestamp.
-func (c Client) GetLogs(ctx context.Context, runID string, since time.Time) (res GetLogsResponse, err error) {
+func (c Client) GetLogs(ctx context.Context, runID string, since time.Time, debug bool) (res GetLogsResponse, err error) {
```
joshma, line 163: maybe merge `since` and `debug` into a `LotOpts` interface? same for GetBuildLogs

pkg/api/client.go
```diff
@@ -160,13 +160,17 @@ func (c Client) GetRun(ctx context.Context, id string) (res GetRunResponse, err
...
 	q := url.Values{"runID": []string{runID}}
 
 	if !since.IsZero() {
 		q.Set("since", since.Format(time.RFC3339))
 	}
 
+	if debug {
+		q.Set("debug", "true")
```
joshma, line 171: maybe the API should be level=info|debug? if debug is true, set level=debug?

pkg/build/remote.go
```diff
@@ -232,7 +232,7 @@ func uploadArchive(ctx context.Context, client *api.Client, archivePath string)
 	return upload.Upload.ID, nil
 }
 
-func waitForBuild(ctx context.Context, client *api.Client, buildID string) error {
+func waitForBuild(ctx context.Context, debug bool, client *api.Client, buildID string) error {
```
joshma, line 235: same here, move to end

pkg/cmd/tasks/deploy/deploy.go
joshma: I'm struggling to come up with the right recommendation here, though, on how to thread that through. @colinking do you have any ideas?

f.e. in root.go we set `logger.EnableDebug = cfg.DebugMode`, which is what `logger.Debug` looks for. maybe we do something similar, set a `LogLevel` variable for watcher.go?

joshma (changes requested):

pkg/api/client.go
colinking: ooh +1, love how this leaves the door open for new level types

pkg/cmd/tasks/deploy/deploy.go
colinking: This is still log-level related, so I wouldn't be against just reading `logger.EnableDebug`!

Just briefly skimming Cobra docs, and they sure seem confusing, but seems like you might be able to access persistent flags in subcommands? There might be a cobra-y way to do it.

Otherwise, could also just have a global `var Debug bool` in `cmd/airplane/main.go` but that might lead to circular deps, so eh.

joshma (approved):

PR #116, opened 2021-05-01
Update install.sh to install to /usr/local/bin

joshma:
It's more aggressive than the current strategy, but it means most users
won't have to add to their rc files to get airplane on their paths.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -3,28 +3,27 @@
 # Inspired by:
 # - https://github.com/denoland/deno_install/blob/master/install.sh
 #   Copyright 2019 the Deno authors. All rights reserved. MIT license.
-# - https://github.com/superfly/flyctl/blob/master/installers/install.sh
 
 set -e
 
 case $(uname -sm) in
-  "Darwin x86_64") target="darwin_x86_64" ;;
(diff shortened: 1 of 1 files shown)


joshma:
putting review to @colinking given he added it originally

joshma:
heroku also installs to /usr/local/bin https://cli-assets.heroku.com/install.sh
so does railway https://raw.githubusercontent.com/railwayapp/cli/master/install.sh

scripts/install.sh
```diff
@@ -3,28 +3,27 @@
 # Inspired by:
 # - https://github.com/denoland/deno_install/blob/master/install.sh
 #   Copyright 2019 the Deno authors. All rights reserved. MIT license.
-# - https://github.com/superfly/flyctl/blob/master/installers/install.sh
```
colinking, line 6: 🤔 lmao

colinking (approved):

scripts/install.sh
joshma: Ooh lol I actually thought it was a typo. The Copyright was left for Deno but not fly

joshma: Will add it back 😊

PR #117, opened 2021-05-04
Add SQL task definition

francescaleung:
Currently requires passing in driver + DSN via options.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -54,6 +54,8 @@ const (
 	TaskKindManual TaskKind = ""
 	TaskKindNode   TaskKind = "node"
 	TaskKindPython TaskKind = "python"
+
+	TaskKindSQL TaskKind = "sql"
 )
 
 type UpdateTaskResponse struct {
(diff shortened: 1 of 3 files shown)


pkg/api/types.go
```diff
@@ -54,6 +54,8 @@ const (
 	TaskKindManual TaskKind = ""
 	TaskKindNode   TaskKind = "node"
 	TaskKindPython TaskKind = "python"
+
```
yields, line 57:
```suggestion
```

pkg/taskdir/definitions/def_0_2.go
```diff
@@ -23,6 +23,8 @@ type Definition_0_2 struct {
 	Node       *NodeDefinition   `yaml:"node,omitempty"`
 	Python     *PythonDefinition `yaml:"python,omitempty"`
```
yields, line 25:
```suggestion
```

Just to match the general code style ;)

yields (approved):
lgtm!

pkg/taskdir/definitions/def_0_2.go
francescaleung: This does actually represent a new category of task type, so I'm inclined to keep them separate for now.

PR #119, opened 2021-05-07
Fix boolean default value

francescaleung:
On boolean value prompt, if you just press enter (without moving the selector), the selector _looks_ like it'll default to "Yes", but it doesn't actually default to yes. It will actually end up sending a null value, which is a terrible idea.

Before:
```
? Is suspended (--is_suspended): 
? Execute? Yes
```

After:
```
? Is suspended (--is_suspended): Yes
? Execute? Yes
```

### DIFF
--- a/pkg/cmd/tasks/execute/cli_inputs.go
+++ b/pkg/cmd/tasks/execute/cli_inputs.go
@@ -94,11 +94,17 @@ func promptForParam(param api.Parameter) (survey.Prompt, error) {
 	}
 	switch param.Type {
 	case api.TypeBoolean:
+		var dv interface{}
+		if defaultValue == "" {
+			dv = nil
+		} else {
+			dv = defaultValue
+		}
 		return &survey.Select{
 			Message: message,
 			Help:    param.Desc,
 			Options: []string{params.YesString, params.NoString},
-			Default: defaultValue,
+			Default: dv,
 		}, nil
 	default:
 		return &survey.Input{


pkg/cmd/tasks/execute/cli_inputs.go
```diff
@@ -94,11 +94,17 @@ func promptForParam(param api.Parameter) (survey.Prompt, error) {
 	}
 	switch param.Type {
 	case api.TypeBoolean:
+		var dv interface{}
+		if defaultValue == "" {
+			dv = nil
```
joshma, line 99: if there's no default value, can we default this to `false` to match UI?

joshma (changes requested):

pkg/cmd/tasks/execute/cli_inputs.go
joshma: Just to make sure I understand - this PR is really "Fix boolean prompt default when no defaultValue is set"? A bit complicated haha but just to confirm.

On second thought, I think what you have here is reasonable.

joshma (approved):

pkg/cmd/tasks/execute/cli_inputs.go
```diff
@@ -94,11 +94,17 @@ func promptForParam(param api.Parameter) (survey.Prompt, error) {
 	}
 	switch param.Type {
 	case api.TypeBoolean:
+		var dv interface{}
```
yields, line 97: could use a ptr to a boolean?

PR #122, opened 2021-05-07
Fix initializing default runtime config in init from scratch

francescaleung:


pkg/cmd/tasks/initcmd/scratch.go
```diff
@@ -93,21 +93,36 @@ func defaultRuntimeConfig(runtime runtimeKind, def *definitions.Definition) (sca
 	// TODO: let folks configure the following configuration
 	switch runtime {
 	case runtimeKindDeno:
-		def.Deno.Entrypoint = "main.ts"
+		deno := definitions.DenoDefinition{
+			Entrypoint: "main.ts",
+		}
+		def.Deno = &deno
```
colinking, line 99: Good catch, can you inline these?

```
def.Deno = &definitions.DenoDefinition{...}
```

colinking (approved):

PR #126, opened 2021-05-18
Create a new revision on build

yields:
Changed the deploy command to create a new task revision with the image URL
once the build succeeds.

I had to refactor the `build` package a bit to unify the remote vs local builders
so we can treat both the same way from the `deploy` command.

### DIFF
--- a/pkg/build/remote.go
+++ b/pkg/build/remote.go
@@ -21,38 +21,51 @@ import (
 	"github.com/pkg/errors"
 )
 
-func Remote(ctx context.Context, dir taskdir.TaskDirectory, client *api.Client, taskID string, env api.TaskEnv) error {
+func remote(ctx context.Context, req Request) (*Response, error) {
+	registry, err := req.Client.GetRegistryToken(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "getting registry token")
+	}
+
 	tmpdir, err := ioutil.TempDir("", "airplane-builds-")
 	if err != nil {
-		return errors.Wrap(err, "creating temporary directory for remote build")
+		return nil, errors.Wrap(err, "creating temporary directory for remote build")
 	}
 	defer os.RemoveAll(tmpdir)
 
 	archivePath := path.Join(tmpdir, "archive.tar.gz")
-	if err := archiveTaskDir(dir, archivePath); err != nil {
-		return err
+	if err := archiveTaskDir(req.Dir, archivePath); err != nil {
+		return nil, err
 	}
 
-	uploadID, err := uploadArchive(ctx, client, archivePath)
+	uploadID, err := uploadArchive(ctx, req.Client, archivePath)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
-	build, err := client.CreateBuild(ctx, api.CreateBuildRequest{
-		TaskID:         taskID,
+	build, err := req.Client.CreateBuild(ctx, api.CreateBuildRequest{
+		TaskID:         req.TaskID,
 		SourceUploadID: uploadID,
-		Env:            env,
+		Env:            req.TaskEnv,
 	})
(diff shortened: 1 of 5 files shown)


pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -144,15 +144,36 @@ func run(ctx context.Context, cfg config) error {
...
+		})
+		if err != nil {
+			return err
+		}
+
+		_, err = client.UpdateTask(ctx, api.UpdateTaskRequest{
+			Slug:             def.Slug,
+			Name:             def.Name,
+			Description:      def.Description,
+			Image:            resp.ImageURL,
```
colinking, line 162: Cool, so when you run `airplane deploy`, it'll update the task to `:latest` while performing the build so the task ends up resolving to the previous version of the task, so it keeps working while building?

pkg/build/remote.go
```diff
@@ -21,38 +21,51 @@ import (
...
-		return err
+	if err := waitForBuild(ctx, req.Client, build.Build.ID); err != nil {
+		return nil, err
 	}
 
-	return nil
+	imageURL := fmt.Sprintf("%s/task-%s:bld_%s",
+		registry.Repo,
+		sanitizeTaskID(req.TaskID),
+		build.Build.ID,
```
colinking, line 63: build ID should already have `bld_` as a prefix I believe

pkg/build/remote.go
```diff
@@ -21,38 +21,51 @@ import (
 	"github.com/pkg/errors"
 )
 
-func Remote(ctx context.Context, dir taskdir.TaskDirectory, client *api.Client, taskID string, env api.TaskEnv) error {
+func remote(ctx context.Context, req Request) (*Response, error) {
+	registry, err := req.Client.GetRegistryToken(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "getting registry token")
+	}
```
colinking, line 28: Do we create a token just to access `registry.Repo`?  If so, I'm not against just copying that logic over to `cli` since we're already doing that for sanitizing task IDs.

colinking (approved):

pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -144,15 +144,36 @@ func run(ctx context.Context, cfg config) error {
...
-		switch builder {
-		case build.BuilderKindLocal:
-			if err := build.Local(ctx, client, dir, def, taskID); err != nil {
-				return err
-			}
-		case build.BuilderKindRemote:
-			if err := build.Remote(ctx, dir, client, taskID, def.Env); err != nil {
-				return err
-			}
+		resp, err := build.Run(ctx, build.Request{
```
colinking, line 147: 😍

pkg/cmd/tasks/deploy/deploy.go
joshma: Wait hm, should the `UpdateTask` call above be deleted? The task shouldn't get updated at all until after the build is done?

yields: We should call the `UpdateTask` above as well since the builder may be `manual`, the user may use their own image.

pkg/build/remote.go
yields: Yeah, but I want all this data to come from the API, updating the CLI every time we make a change to our registry will get annoying.

yields: Ideally the full image URL will be decided by the API or something.

pkg/cmd/tasks/deploy/deploy.go
joshma: In that case, though, shouldn't it just skip the build step and call UpdateTask here? Why call it twice in the manual case?

PR #127, opened 2021-05-18
Add basic REST definition

francescaleung:
Also moves `KindOptions` to a generic `map[string]interface{}` type.

### DIFF
--- a/pkg/build/local.go
+++ b/pkg/build/local.go
@@ -28,7 +28,7 @@ func local(ctx context.Context, req Request) (*Response, error) {
 	args := make(map[string]string, len(options))
 	for k, v := range options {
 		if sv, ok := v.(string); !ok {
-			return errors.New("unexpected non-string option for builder arg")
+			return nil, errors.New("unexpected non-string option for builder arg")
 		} else {
 			args[k] = sv
 		}
--- a/pkg/build/local.go
+++ b/pkg/build/local.go
@@ -26,10 +26,18 @@ func Local(ctx context.Context, client *api.Client, dir taskdir.TaskDirectory, d
 	if err != nil {
 		return err
 	}
+	args := make(map[string]string, len(options))
+	for k, v := range options {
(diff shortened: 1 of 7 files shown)


pkg/taskdir/definitions/definitions.go
```diff
@@ -31,49 +32,52 @@ func NewDefinitionFromTask(task api.Task) (Definition, error) {
...
-			Query: task.KindOptions["query"],
+		def.SQL = &SQLDefinition{}
+		if err := mapstructure.Decode(task.KindOptions, &def.SQL); err != nil {
+			return Definition{}, errors.Wrap(err, "decoding SQL options")
+		}
+
+	} else if task.Kind == api.TaskKindREST {
+		def.REST = &RESTDefinition{}
+		if err := mapstructure.Decode(task.KindOptions, &def.REST); err != nil {
+			return Definition{}, errors.Wrap(err, "decoding REST options")
```
colinking, line 79: Haven't tested this, but could you share the `mapstructure.Decode` step via something like this?

```go
var def interface{}

if task.Kind == api.TaskKindDeno {
  task.Deno = &DenoDefinition{}
  def = &task.Deno
} else ... {
  // ...
}

// f.e. manual
if def != nil {
  if err := mapstructure.Decode(task.KindOptions, def); err != nil {
    // ...
  }
}
```

pkg/build/local.go
```diff
@@ -25,10 +25,18 @@ func local(ctx context.Context, req Request) (*Response, error) {
 	if err != nil {
 		return nil, err
 	}
+	args := make(map[string]string, len(options))
+	for k, v := range options {
+		if sv, ok := v.(string); !ok {
+			return nil, errors.New("unexpected non-string option for builder arg")
```
colinking, line 31: wdyt of passing `KindOptions` through instead of converting to a `map[string]string`?

colinking (approved):

PR #129, opened 2021-05-24
Add exec alias

yields:
Adds exec alias:

```bash
$ airplane exec <slug> [flags]
$ airplane execute <slug> [flags]
```


colinking (approved):

joshma:
(Which one is going to be standard moving forward? Not sure I like having too many ways to do the same thing, but it's just a minor feeling.)

PR #130, opened 2021-05-24
Refactor init command

yields:
The command used to accept a slug and provide an interactive tutorial
on how to generate the task.

The changes here accept a slug and a filepath and the command will
generate some startup code based on the filepath extension.

```bash
  λ cli (refactor/init) go run cmd/airplane/main.go init --host api.airplane.so:5000 --slug echo_loop /tmp/task.ts
  λ cli (refactor/init) cat /tmp/task.ts                                                                          
```

```ts
// airplane: echo_loop

type Params = {
  count: number
}

export default async function(args: Params){
  console.log('arguments: ', args);
}
```


yields:
@colinking lmk if the we want to merge into `v2` instead of `main` and I'll change it.

cmd/airplane/main.go
```diff
@@ -37,7 +37,7 @@ func main() {
 			logger.Log("")
 			logger.Log(exerr.ExplainError())
 		} else {
-			logger.Log(logger.Red("Error: %s", errors.Cause(err).Error()))
+			logger.Error(errors.Cause(err).Error())
```
colinking, line 40: Darn, I liked the red 😄

pkg/runtime/javascript/javascript.go
```diff
@@ -0,0 +1,75 @@
...
+
+	if err := code.Execute(&buf, args); err != nil {
+		return nil, fmt.Errorf("typescript: template execute - %w", err)
+	}
+
+	return buf.Bytes(), nil
+}
+
+// URL implementation.
+func (r Runtime) URL(code []byte) (string, bool) {
```
colinking, line 49: Could we share this logic across interfaces? Seems all we would need is for the interface to declare what comment character to use (`//` vs `#`) but the rest of the logic is the same.

pkg/runtime/javascript/javascript.go
```diff
@@ -0,0 +1,75 @@
...
+// Runtime implementaton.
+type Runtime struct{}
+
+// Generate implementation.
+func (r Runtime) Generate(t api.Task) ([]byte, error) {
+	var args = data{URL: t.URL}
+	var buf bytes.Buffer
+
+	if err := code.Execute(&buf, args); err != nil {
+		return nil, fmt.Errorf("typescript: template execute - %w", err)
```
colinking, line 42:
```suggestion
		return nil, fmt.Errorf("javascript: template execute - %w", err)
```

pkg/runtime/typescript/typescript.go
```diff
@@ -0,0 +1,84 @@
...
+
+// Param represents the parameter.
+type param struct {
+	Name string
+	Type string
+}
+
+// Runtime implementaton.
+type Runtime struct {
+	javascript.Runtime
```
colinking, line 46: nice

pkg/runtime/typescript/typescript.go
```diff
@@ -0,0 +1,84 @@
...
+	case api.TypeInteger, api.TypeFloat:
+		return "number"
+	case api.TypeDate, api.TypeDatetime:
+		return "string"
+	case api.TypeBoolean:
+		return "boolean"
+	case api.TypeString:
+		return "string"
+	default:
+		// TODO(amir): ideally how do we convert upload this type in typescript.
```
colinking, line 81: An upload would come in as a URL string, so maybe just `string` for now?

pkg/cmd/tasks/initcmd/init.go
```diff
@@ -6,112 +6,150 @@ package initcmd
...
+	var cfg = config{client: c.Client}
 
 	cmd := &cobra.Command{
 		Use:   "init",
 		Short: "Initialize a task definition",
 		Example: heredoc.Doc(`
 			$ airplane tasks init
+			$ airplane tasks init --slug task-slug ./my/task.js
+			$ airplane tasks init --slug task-slug ./my/task.ts
 			$ airplane tasks init -f ./airplane.yml
```
colinking: Let's remove the docs on the old `init` behavior!

pkg/cmd/tasks/initcmd/init.go
```diff
@@ -6,112 +6,150 @@ package initcmd
...
+	var client = cfg.client
+
+	task, err := client.GetTask(ctx, cfg.slug)
+	if err != nil {
+		return err
+	}
+
+	r, ok := runtime.Lookup(cfg.file)
+	if !ok {
+		return fmt.Errorf("cannot find a runtime that matches %s", cfg.file)
```
colinking, line 68: Maybe something like this?

```suggestion
		return fmt.Errorf("Unable to deploy tasks with a %s file ending.", fileEnding)
```

pkg/cmd/tasks/initcmd/init.go
```diff
@@ -6,112 +6,150 @@ package initcmd
...
+		buf, err := ioutil.ReadFile(cfg.file)
+		if err != nil {
 			return err
 		}
-	case initKindScratch:
-		if err := initFromScratch(ctx, cfg); err != nil {
+
+		if u, ok := r.URL(buf); ok {
+			logger.Log("%s is already linked to %s", cfg.file, cfg.slug)
+			logger.Log("URL: %s", u)
```
colinking, line 79: If users end up here, bet they are actually trying to deploy. Maybe we drop advice on how to do that?

```
{file} is already linked to {u}

You can deploy this task with:
  airplane deploy -f {file}
```

Also it's possible `url.slug != cfg.slug`, we might want to use the slug from the URL instead!

pkg/cmd/tasks/initcmd/init.go
```diff
@@ -6,112 +6,150 @@ package initcmd
...
+		code = append(code, '\n', '\n')
+		code = append(code, buf...)
+
+		if err := ioutil.WriteFile(cfg.file, code, 0644); err != nil {
 			return err
 		}
-	default:
-		return errors.Errorf("Unexpected unknown initKind choice: %s", kind)
+
+		logger.Log("Linked %s to %s", cfg.file, cfg.slug)
```
colinking, line 101: Maybe add a deploy comment here, too

pkg/api/client.go
```diff
@@ -113,6 +113,12 @@ func (c Client) UpdateTask(ctx context.Context, req UpdateTaskRequest) (res Upda
 // ListTasks lists all tasks.
 func (c Client) ListTasks(ctx context.Context) (res ListTasksResponse, err error) {
 	err = c.do(ctx, "GET", "/tasks/list", nil, &res)
+	if err != nil {
+		return
+	}
+	for j, t := range res.Tasks {
+		res.Tasks[j].URL = c.TaskURL(t.ID)
```
colinking, line 120: This URL will include an ID not a slug. Maybe we should add some slug-based routes for tasks?

Maybe this? (single-letter routes would signify we should load by slug not ID)

```
airplane.dev/t/<slug>
```

colinking (approved):

pkg/api/client.go
colinking: Though alternatively, we could use IDs here... hmm

cmd/airplane/main.go
yields: ah the red is still there I think?, `logger.Error()` should color it :)

pkg/runtime/javascript/javascript.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/airplanedev/cli/pkg/runtime"
+)
+
+// Init register the runtime.
+func init() {
+	runtime.Register(".js", Runtime{})
+}
+
+// Code template.
+var code = template.Must(template.New("js").Parse(`// airplane: {{ .URL }}
```
colinking, line 21: wdyt of removing the `airplane:` prefix?

cmd/airplane/main.go
colinking: Ah, I missed that! :awyis:

pkg/api/types.go
```diff
@@ -267,6 +267,7 @@ func (this *EnvVarValue) UnmarshalYAML(node *yaml.Node) error {
 
 // Task represents a task.
 type Task struct {
+	URL              string           `json:"-" yaml:"-"`
```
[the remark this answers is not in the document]
yields: The issue is that we initialize the client at the root of the command tree, and only it knows what the URL should be. I contemplated just hardcoding `app.airplane.dev` but not sure. wdyt?

pkg/runtime/javascript/javascript.go
yields: Yeah, typescript re-uses those funcs, I think we should just add some `code` package that implements those, wdyt?

pkg/cmd/tasks/initcmd/init.go
yields: What if you do `init --slug /tmp/foo` then there's no file ext 😅

pkg/cmd/tasks/initcmd/init.go
yields: Done, just compared the URL to the task's URL.

pkg/api/client.go
yields: Yeah, since slugs are unique I think it'd be nicer anyway. btw, I really like the single letter there, makes it easy to just type out the tasks :)

pkg/runtime/javascript/javascript.go
yields: I don't mind either way, it kinda makes sense tho, you can tell immediately who added that URL. Hopefully no one deletes it 😂

pkg/api/types.go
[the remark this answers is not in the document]
colinking: Ah, supporting different hosts would be great. sgtm as-is then :)

pkg/cmd/tasks/initcmd/init.go
colinking: That could be a separate error! `Expected a file ending` etc.

pkg/api/client.go
colinking: I could go either way, but maybe let's go with IDs for now since they are already supported in the app. They also lend themselves to a "generated" feel and align with the URL folks already see when looking at their task in the UI!

colinking: If we use an ID in the comment, would it make sense of the CLI command to also use IDs?

```sh
airplane init --id t_1234 ./path.js
```

yields: What's the benefit in doing so? seems too verbose 😅

PR #131, opened 2021-05-24
Add shim-based Node builder

colinking:
This PR adds a new version of the Node builder that uses a shim to execute Node tasks. This will be opt-in by using the new version of the CLI, so old versions of the CLI will continue to work.

### DIFF
--- a/cmd/shimtest/main.go
+++ b/cmd/shimtest/main.go
@@ -1,0 +1,57 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/airplanedev/cli/pkg/api"
+	"github.com/airplanedev/cli/pkg/build"
+	"github.com/airplanedev/cli/pkg/conf"
+	"github.com/airplanedev/cli/pkg/taskdir"
+	"github.com/airplanedev/cli/pkg/taskdir/definitions"
+)
+
+// This is used to test the Node.js shim. It will be removed
+// before merging and replaced by a future PR adding
+// support for deploying via the standard `deploy` command.
+func main() {
+	if len(os.Args) != 3 {
+		fmt.Printf("shimtest [filepath] [host]")
+		os.Exit(1)
+	}
+	path := os.Args[1]
+	host := os.Args[2]
+
+	client := &api.Client{
+		Host: host,
+	}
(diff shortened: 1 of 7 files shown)


cmd/shimtest/main.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/airplanedev/cli/pkg/build"
+	"github.com/airplanedev/cli/pkg/conf"
+	"github.com/airplanedev/cli/pkg/taskdir"
+	"github.com/airplanedev/cli/pkg/taskdir/definitions"
+)
+
+// This is used to test the Node.js shim. It will be removed
+// before merging and replaced by a future PR adding
+// support for deploying via the standard `deploy` command.
+func main() {
```
colinking, line 18: This is just for testing, since we can't test the shim via `airplane deploy` yet. @yields feel free to remove this in the PR that adds `deploy` support for Node

pkg/build/node.go
```diff
@@ -10,17 +10,165 @@ import (
...
+	try {
+		await task(JSON.parse(process.argv[2]))
+	} catch (err) {
+		console.error(err)
+		console.log("airplane_output:error " + JSON.stringify({ "error": String(err) }))
+		process.exit(1)
+	}
+}
+
+main()`
```
colinking, line 85: We should start shipping our own base images at some point, and we can move this shim into that base image at that point. For now, I don't think it's worth the effort and just inlining this file works just fine for now.

However, we may also want to pull some of this out for local execution. Ideally local execution (presumably happening outside of Docker) is as similar to the deployed version as possible.

pkg/build/node.go
```diff
@@ -10,17 +10,165 @@ import (
...
+		{{else}}
+		RUN echo '{}' > package.json
+		{{end}}
+
+		{{if .HasPackageLock}}
+		RUN npm install package-lock.json && npm install --save-dev @types/node
+		{{else if .HasYarnLock}}
+		RUN yarn install && yarn add -D @types/node
+		{{else}}
+		npm install @types/node
```
colinking, line 126: We need `@types/node` for the shim file, so this makes sure it is installed.

yields (approved):
lgtm! 💥 

should we merge all the things into some sort of `v2` branch? I'm fine either way tbh

PR #132, opened 2021-05-26
Add Node 16

colinking:
This PR adds a digest for Node 16. I've separately deployed this digest to the public registry cache.


yields (approved):
lgtm

PR #133, opened 2021-05-26
Init changes

yields:
Small init changes to make `init` command a bit better.

Here's how the file looks like after those changes:

```javascript
// Generated by Airplane, do not edit:
// https://app.airplane.so:5000/t/echo_loop

export default async function(args){
  console.log('arguments: ', args);
}
```

---

- Add task.Kind validation to init
- Change init comment
- Change TaskURL to t/<slug> instead of tasks/<id>
- Adjust runtime parsing of task URL
- Change runtime to extract slug instead of full URL


pkg/cmd/tasks/initcmd/init.go
```diff
@@ -74,13 +74,17 @@ func run(ctx context.Context, cfg config) error {
 		return err
 	}
 
+	if task.Kind != api.TaskKindNode {
+		return fmt.Errorf("cannot link %q to a non node.js task", cfg.file)
+	}
```
colinking, line 79: wdyt of a `.TaskKind() TaskKind` func on `runtime.Interface`? This would also capture the case where you try to link a `.js` file to a Python task, etc.

colinking (approved):

pkg/cmd/tasks/initcmd/init.go
yields: Yeah definitely, intending to add it once we add more runtimes that are not node.js

PR #135, opened 2021-05-27
Default newly created tasks with NULL image

colinking:
This PR makes a few changes:
1. Tasks created from the CLI will have their `image` field set to `NULL` (indicating they haven't been deployed yet). This mirrors the UI now.
2. The `image` field is now only changed after a build completes, at which point it is set to the `:bld_...` tag from the Airplane Registry. It previously was also changed before the build started b/c it was set to `""` which gets sanitized into `:latest` in the Airplane Registry via our API. We now just leave the `image` field as-is before the deploy.


colinking:
cc @yields since I know you're working on unrelated `deploy` changes

pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -93,21 +93,27 @@ func run(ctx context.Context, cfg config) error {
...
-			Repo:             def.Repo,
-			Timeout:          def.Timeout,
+			// Only update kind/kindOptions since the remote build will
+			// fetch this from the task.
+			Kind:        kind,
+			KindOptions: kindOptions,
+
+			// The following fields are not updated until after the build finishes.
+			//
+			// TODO: move kind/kindOptions to builds so we don't need to do this.
```
joshma, line 103: Hm @yields @colinking I don't think my comment here was ever addressed: https://github.com/airplanedev/cli/pull/126/files#r635370901

Why are we doing this pre-deploy update in the first place?

pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -93,21 +93,27 @@ func run(ctx context.Context, cfg config) error {
...
+			Name:             task.Name,
+			Description:      task.Description,
+			Image:            task.Image,
+			Command:          task.Command,
+			Arguments:        task.Arguments,
+			Parameters:       task.Parameters,
+			Constraints:      task.Constraints,
+			Env:              task.Env,
+			ResourceRequests: task.ResourceRequests,
+			Resources:        task.Resources,
```
joshma, line 114: > Fixes a bug where resources were disconnected by that pre-deploy UpdateTask call

just for my understanding, where was this happening? I see that it used to be `Resources: def.Resources`, is that not right?

pkg/cmd/tasks/deploy/deploy.go
colinking: We don't store `kind` and `kindOptions` on the `builds` table, those are read from `tasks`. We should move them over, but until then we do need to do this update.

pkg/cmd/tasks/deploy/deploy.go
colinking: Ah, sorry nm about that :)

pkg/cmd/tasks/deploy/deploy.go
colinking: Gonna tweak this a bit so that we properly update manual tasks

joshma: Ah I see ok - it's right there in the comment haha. Thanks.

pkg/print/types.go
```diff
@@ -11,13 +11,14 @@ type printTask struct {
...
 	Description      string               `json:"description" yaml:"description"`
-	Image            string               `json:"image" yaml:"image"`
+	Image            *string              `json:"image" yaml:"image"`
 	Command          []string             `json:"command" yaml:"command"`
 	Arguments        []string             `json:"arguments" yaml:"arguments"`
 	Parameters       api.Parameters       `json:"parameters" yaml:"parameters"`
 	Constraints      api.RunConstraints   `json:"constraints" yaml:"constraints"`
 	Env              api.TaskEnv          `json:"env" yaml:"env"`
 	ResourceRequests api.ResourceRequests `json:"resourceRequests" yaml:"resourceRequests"`
+	Resources        api.Resources        `json:"resources" yaml:"resources"`
```
joshma, line 21: Note this'll be an ID not the name...

joshma (approved):

yields (approved):
lgtm!

PR #136, opened 2021-05-27
Deploy changes

yields:
The goal of the PR is to allow users to deploy like so:

```bash
$ airplane deploy task.yml
$ airplane deploy task.yaml
$ airplane deploy task.js
$ airplane deploy task.ts
```

To do this, i had to change the builder to not depend on `taskdir`, since scripts don't
have definitions and the taskdir relies on them.

I also had to hardcode serveral things (like file extensions, kinds etc) but this is just
so we get it running and working, the next PRs will improve the quality here by relying
on `runtime.Interface` to do those things.

The PR also omits package.json detection, we still need to figure out how to do this:

- Do we do it at `init`? if so, where do we persist the state, how can someone change it?
- Do we do it at `deploy`? if so, wouldn't it get annoying to always say yes to prompts before deploys?

Currently, we just do `filepath.Dir('task.js')` and treat that as the root.

### DIFF
--- a/cmd/shimtest/main.go
+++ b/cmd/shimtest/main.go
@@ -40,7 +40,7 @@ func main() {
 	if resp, err := build.Run(context.Background(), build.Request{
 		Builder: build.BuilderKindLocal,
 		Client:  client,
-		Dir:     dir,
+		Root:    dir.DefinitionRootPath(),
 		Def: definitions.Definition(definitions.Definition_0_2{
 			Node: &definitions.NodeDefinition{
 				Entrypoint: path,
--- a/pkg/cmd/tasks/deploy/script.go
+++ b/pkg/cmd/tasks/deploy/script.go
@@ -60,7 +60,7 @@ func deployFromScript(ctx context.Context, cfg config) error {
 		return err
 	}
 
-	def.Node = &definitions.NodeDefinition{}
+	def.Node.Entrypoint = cfg.file
 
 	resp, err := build.Run(ctx, build.Request{
 		Builder: buildkind,
--- a/pkg/cmd/tasks/deploy/script.go
+++ b/pkg/cmd/tasks/deploy/script.go
@@ -60,7 +60,7 @@ func deployFromScript(ctx context.Context, cfg config) error {
 		return err
 	}
 
-	def.Node.Entrypoint = cfg.file
+	def.Node.Entrypoint = filepath.Base(abs)
 
 	resp, err := build.Run(ctx, build.Request{
 		Builder: buildkind,
@@ -84,7 +84,7 @@ func deployFromScript(ctx context.Context, cfg config) error {
 		Name:             def.Name,
 		Description:      def.Description,
 		Image:            &resp.ImageURL,
-		Command:          []string{"node", cfg.file},
+		Command:          []string{},
 		Arguments:        def.Arguments,
 		Parameters:       def.Parameters,
 		Constraints:      def.Constraints,
--- a/pkg/cmd/tasks/deploy/script.go
+++ b/pkg/cmd/tasks/deploy/script.go
@@ -1,0 +1,131 @@
+package deploy
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+
+	"github.com/airplanedev/cli/pkg/api"
+	"github.com/airplanedev/cli/pkg/build"
+	"github.com/airplanedev/cli/pkg/logger"
+	"github.com/airplanedev/cli/pkg/runtime"
+	"github.com/airplanedev/cli/pkg/taskdir/definitions"
+)
+
+// DeployFromScript deploys from the given script.
+func deployFromScript(ctx context.Context, cfg config) error {
+	var client = cfg.client
+	var ext = filepath.Ext(cfg.file)
+
+	if ext == "" {
+		return errors.New("cannot deploy a file without extension")
+	}
+
+	r, ok := runtime.Lookup(cfg.file)
+	if !ok {
+		return fmt.Errorf("cannot deploy a file with extension of %q", ext)
+	}
+
+	code, err := ioutil.ReadFile(cfg.file)
+	if err != nil {
+		return fmt.Errorf("reading %s - %w", cfg.file, err)
+	}
+
+	slug, ok := r.Slug(code)
+	if !ok {
+		return errors.New("")
+	}
+
+	task, err := client.GetTask(ctx, slug)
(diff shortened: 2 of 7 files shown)


joshma:
Prob better for @colinking to review this one

pkg/cmd/tasks/deploy/script.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	code, err := ioutil.ReadFile(cfg.file)
+	if err != nil {
+		return fmt.Errorf("reading %s - %w", cfg.file, err)
+	}
+
+	slug, ok := r.Slug(code)
+	if !ok {
+		return errors.New("")
```
colinking, line 38: ^ needs an error

pkg/cmd/tasks/deploy/script.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return err
+	}
+
+	abs, err := filepath.Abs(cfg.file)
+	if err != nil {
+		return err
+	}
+
+	def.Node = &definitions.NodeDefinition{}
```
colinking, line 61: Do we need to initialize this? It should get set via `NewDefinitionFromTask` and will include configuration like `nodeVersion`: https://github.com/airplanedev/cli/blob/82e9ee0b27c6b19fc070e5ecf06b1edaaf557b21/pkg/taskdir/definitions/definitions.go#L74

pkg/cmd/tasks/deploy/script.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return err
+	}
+
+	_, err = client.UpdateTask(ctx, api.UpdateTaskRequest{
+		Slug:             def.Slug,
+		Name:             def.Name,
+		Description:      def.Description,
+		Image:            resp.ImageURL,
+		Command:          []string{"node", cfg.file},
```
colinking, line 85: Let's leave this as-is -- it'll be `[]` which instructs the agent to default to the the entrypoint from inside the Docker image.

```suggestion
		Command:          def.Command
```

pkg/cmd/tasks/deploy/script.go
```diff
@@ -0,0 +1,131 @@
...
+
+	abs, err := filepath.Abs(cfg.file)
+	if err != nil {
+		return err
+	}
+
+	def.Node = &definitions.NodeDefinition{}
+
+	resp, err := build.Run(ctx, build.Request{
+		Builder: buildkind,
```
colinking, line 64: Could we defer to `--builder` here instead of hardcoding? So we can test this with the remote builder!

pkg/cmd/tasks/deploy/script.go
```diff
@@ -0,0 +1,131 @@
...
+	if !ok {
+		return errors.New("")
+	}
+
+	task, err := client.GetTask(ctx, slug)
+	if err != nil {
+		return err
+	}
+
+	def, err := definitions.NewDefinitionFromTask(task)
```
colinking, line 46: nice 🔥

cmd/shimtest/main.go
```diff
@@ -40,7 +40,7 @@ func main() {
 	if resp, err := build.Run(context.Background(), build.Request{
 		Builder: build.BuilderKindLocal,
 		Client:  client,
-		Dir:     dir,
+		Root:    dir.DefinitionRootPath(),
```
colinking, line 43: You can remove `shimtest` now that `deploy` works!

colinking (approved):

pkg/cmd/tasks/deploy/yaml.go
```diff
@@ -0,0 +1,157 @@
...
+
+	"github.com/airplanedev/cli/pkg/api"
+	"github.com/airplanedev/cli/pkg/build"
+	"github.com/airplanedev/cli/pkg/logger"
+	"github.com/airplanedev/cli/pkg/taskdir"
+	"github.com/pkg/errors"
+)
+
+// DeployFromYaml deploys from a yaml file.
+func deployFromYaml(ctx context.Context, cfg config) error {
```
colinking, line 15: FYI, made changes to this codepath that you'll want to pull in: https://github.com/airplanedev/cli/pull/135

pkg/cmd/tasks/deploy/script.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	abs, err := filepath.Abs(cfg.file)
+	if err != nil {
+		return err
+	}
+
+	def.Node = &definitions.NodeDefinition{}
+
+	resp, err := build.Run(ctx, build.Request{
```
colinking, line 63: We'll need to pass `cfg.file` here, too. The `node` builder currently looks at [`args["entrypoint"]`](https://github.com/airplanedev/cli/blob/998445d1fe533b2cf73f743b1886bb228a5e754c/pkg/build/node.go#L24), but we're no longer storing `entrypoint` as `kindConfig` (since you instead pass it to the CLI, rather than storing it in the YAML file in advance). The other builders can continue to pull `entrypoint` from `airplane.yml`, but the Node builder should be updated to get it from `cfg.file` instead.

colinking:
(noting for others that using this command to deploy JS tasks with the new shim builder is broken until https://github.com/airplanedev/cli/pull/137 merges)

pkg/cmd/tasks/deploy/yaml.go
yields: done :)

pkg/cmd/tasks/deploy/script.go
yields: There's no `def.Command`, changed to `[]`.

pkg/cmd/tasks/deploy/script.go
yields: Good call, removed the explicit init.

cmd/shimtest/main.go
yields: hmm, let's remove it once it works end-to-end ;)

pkg/cmd/tasks/deploy/script.go
yields: Yep, I've explicit set `def.Node.Entrypoint` to the base name of the file.

yields:
Seems like it works well! :)

```bash
$ go run cmd/airplane/main.go --host=api.airplane.so:5000 deploy --builder=local ../tasks/echo_js/index.js
$ docker run -it $(docker images -q | head -1) '{ "name": "Larry" }'
arguments:  { name: 'Larry' }
```

PR #137, opened 2021-05-27
Fix issues with shim builder

colinking:
This PR fixes two issues with the new Node.js shim builder.
1. We were using `html/template` which was escaping quotes and other characters in `shim.ts`. Moving to `text/template` disables escaping.
2. Missing `RUN` command when not using npm or yarn.

It also updates the default Node version to 16 in a place that I missed in a previous PR.

### DIFF
--- a/pkg/build/builder.go
+++ b/pkg/build/builder.go
@@ -6,11 +6,11 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
-	"html/template"
 	"os"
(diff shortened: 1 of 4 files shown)


colinking:
Assigning to @yields since he has more context

yields (approved):

yields:
@colinking gonna merge this one cause I think I saw an issue with deploy that relates to it :)

PR #138, opened 2021-05-28
Refactor execute to accept scripts

yields:
Previously execute would accept a yaml definition using the -f flag
since init and deploy removed that flag and just accept a file positional
argument we removed that --file flag.

Essentially changes execute to detect what type of position argument it was
given, it may be a slug, a yaml definition or a script. Once it detects it
it will lookup the task and execute it remotely.

There are still some open questions on how we are going to add local execution
i think it might be a larger effort to add, especially since we rely on a shim etc.

For now, this works and i suggest we merge it and focus on developer experience:

 - Better error messages
 - Nicer CLI UI
 - Edge-cases
 - Detecting root via package.json / .git

### DIFF
--- a/pkg/cmd/tasks/execute/execute.go
+++ b/pkg/cmd/tasks/execute/execute.go
@@ -3,15 +3,20 @@ package execute
 import (
 	"context"
 	"flag"
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
 	"strings"
 
 	"github.com/MakeNowJust/heredoc"
 	"github.com/airplanedev/cli/pkg/api"
 	"github.com/airplanedev/cli/pkg/cli"
 	"github.com/airplanedev/cli/pkg/cmd/auth/login"
+	"github.com/airplanedev/cli/pkg/fs"
 	"github.com/airplanedev/cli/pkg/logger"
 	"github.com/airplanedev/cli/pkg/params"
 	"github.com/airplanedev/cli/pkg/print"
+	"github.com/airplanedev/cli/pkg/runtime"
 	"github.com/airplanedev/cli/pkg/taskdir"
 	"github.com/airplanedev/cli/pkg/utils"
 	"github.com/pkg/errors"
@@ -21,9 +26,8 @@ import (
 // Config is the execute config.
 type config struct {
 	root *cli.Config
-	slug string
-	args []string
 	file string
+	args []string
 }
 
 // New returns a new execute cobra command.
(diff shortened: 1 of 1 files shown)


pkg/cmd/tasks/execute/execute.go
```diff
@@ -53,46 +60,25 @@ func New(c *cli.Config) *cobra.Command {
 				cfg.args = args[n:]
 			}
 
-			// If an arg was passed, before the --, then it is a task slug to execute.
+			// If an arg was passed, before the --, then it is a task file/slug to execute.
 			if len(args) > 0 && n != 0 {
-				cfg.slug = args[0]
+				cfg.file = args[0]
```
colinking, line 65: So this could be a file or a slug. Should we call it `fileOrSlug`? Verbose, but gets the point across 😄

pkg/cmd/tasks/execute/execute.go
```diff
@@ -208,3 +194,60 @@ func flagset(task api.Task, args api.Values) *flag.FlagSet {
...
+			return "", fmt.Errorf("the file %s must have an extension", file)
+		default:
+			return "", fmt.Errorf("the file %s has unrecognized extension", file)
+		}
+	}
+	return file, nil
+}
+
+// FromYaml attempts to extract a slug from a yaml definition.
+func fromYaml(file string) (string, error) {
```
colinking, line 216: Maybe `slugFromYAML`? 🙏🏻  same with `slugFromScript`

pkg/cmd/tasks/execute/execute.go
```diff
@@ -36,12 +40,15 @@ func New(c *cli.Config) *cobra.Command {
...
-			airplane execute -f ./airplane.yml [-- <parameters...>]
+			airplane execute ./airplane.yml [-- <parameters...>]
+			airplane execute ./task.js [-- <parameters...>]
+			airplane execute ./task.ts [-- <parameters...>]
 			airplane execute hello_world [-- <parameters...>]
 		`),
 		PersistentPreRunE: utils.WithParentPersistentPreRunE(func(cmd *cobra.Command, args []string) error {
 			return login.EnsureLoggedIn(cmd.Root().Context(), c)
 		}),
+		Args: cobra.ExactArgs(1),
```
colinking, line 51: Does this include the `--`? Does `airplane execute --`  err from a missing positional arg?

pkg/cmd/tasks/execute/execute.go
```diff
@@ -53,46 +60,25 @@ func New(c *cli.Config) *cobra.Command {
...
-		if err != nil {
-			return err
-		}
-
-		if def.Slug == "" {
-			return errors.Errorf("no task slug found in task definition at %s", cfg.file)
-		}
-
-		slug = def.Slug
+	slug, err := slugFrom(cfg.file)
```
colinking, line 79: So there'll be a few options here:

```sh
airplane execute slug # 1
airplane execute ./file.ts # 2
airplane execute ./file.yml # 3
airplane execute slug --remote # 4
airplane execute ./file.ts --remote # 5
airplane execute ./file.yml --remote # 6
```

The current behavior feels good for 1/3/4/6. 2 should end up running the local copy of the code, AD, so that you can make changes and re-run without re-deploying.  It's 5 that is trickier IMO. Feels like `airplane execute ./file.ts --remote` should be equivalent to `airplane execute ./file.ts` except that it runs remotely. Does that mean we should automatically do a deploy first for you? If you've already done a deploy and haven't made any changes since, is there a way we can detect that and skip the deploy? Maybe a prompt is the best we can do. Or maybe we calculate a checksum of the filesystem under the root?

colinking (approved):

pkg/cmd/tasks/execute/execute.go
yields: yeah hmm, we'll talk about local execution once you're back.

pkg/cmd/tasks/execute/execute.go
yields: whoops fixed by changing to minargs and changing how we collect the file/slug and the arguments.

pkg/cmd/tasks/execute/execute.go
yields: Change to `task`, added docs too.

PR #139, opened 2021-05-28
Add --local to deploy command

yields:
Noticed that `BuildKind` adds some redundant funcs, checks etc where we only really
need a boolean there.


joshma (approved):
Hm sure, LGTM. I had half a thought that we'd eventually support N builders but that's likely premature.

PR #140, opened 2021-05-28
Run staticcheck

yields:
I ran staticcheck on the repo, it follows:
https://github.com/golang/go/wiki/CodeReviewComments

Let me know if anyone feels strongly about keeping the "this" receiver.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -240,13 +240,13 @@ var _ yaml.Unmarshaler = &EnvVarValue{}
 //   AIRPLANE_DSN:
 //     value: "foobar"
 //
-func (this *EnvVarValue) UnmarshalYAML(node *yaml.Node) error {
+func (ev *EnvVarValue) UnmarshalYAML(node *yaml.Node) error {
 	// First, try to unmarshal as a string.
 	// This would be the first case above.
 	var value string
 	if err := node.Decode(&value); err == nil {
 		// Success!
-		this.Value = &value
+		ev.Value = &value
 		return nil
 	}
 
@@ -260,7 +260,7 @@ func (this *EnvVarValue) UnmarshalYAML(node *yaml.Node) error {
 	if err := node.Decode(&v); err != nil {
 		return err
 	}
-	*this = EnvVarValue(v)
+	*ev = EnvVarValue(v)
 
 	return nil
 }
@@ -414,8 +414,8 @@ const (
 	BuildCancelled  BuildStatus = "Cancelled"
 )
 
-func (this BuildStatus) Stopped() bool {
-	return this == BuildSucceeded || this == BuildFailed || this == BuildCancelled
+func (s BuildStatus) Stopped() bool {
+	return s == BuildSucceeded || s == BuildFailed || s == BuildCancelled
 }
 
 type CreateBuildUploadRequest struct {
--- a/pkg/build/versions.go
+++ b/pkg/build/versions.go
@@ -25,12 +25,12 @@ type Version struct {
 	Digest string `json:"digest"`
 }
 
-func (this Version) String() string {
-	if this.Image == "" || this.Digest == "" {
+func (v Version) String() string {
+	if v.Image == "" || v.Digest == "" {
 		return ""
 	}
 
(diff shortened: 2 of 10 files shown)


joshma:
> Let me know if anyone feels strongly about keeping the "this" receiver.

😂  @colinking

colinking:
Yeah, I prefer to use a standard receiver name. I generally disagree with https://github.com/golang/go/wiki/CodeReviewComments#receiver-names because:

1) **Refactoring**: If the receiver's name reflects the type's name, then it falls out of date as you change type's name.
2) **Interfaces**: Interfaces have multiple implementations, each with receivers. It's easy enough to copy a declaration and forget to change the the receiver name. When diffing two implementations with similar code, you end up with unnecessary deltas b/c of different receiver names.
3) **Functional Output**: A core part of this recommendation is this statement: *"In Go, the receiver of a method is just another parameter and therefore, should be named accordingly."* How often do you write functions that emit side effects through their parameters? The main exception to that is code similar to `json.Marshal` who returns its data through a pointer. On the other hand, side effects to a receiver are commonplace (f.e. `this.myProperty = "123"`). The fact that receivers are parameters under-the-hood is irrelevant considering the conventions are different.
4) **Readability**: While reading code, you have to remember the receiver name to know if a variable is the receiver or not. If the receiver is always `this`, then that's one less bit of information to remember while reading someone else's code.

Hehe, so yeah. Alas, it's the Go recommendation so what can I do :)

joshma:
Meh, I think this has me leaning towards agreeing with the golang code guidelines.

> 1. **Refactoring**: If the receiver's name reflects the type's name, then it falls out of date as you change type's name.

Yeah, this is true. But it's the same as what happens to variables (`client := APIClient{}`).

> 2. **Interfaces**: Interfaces have multiple implementations, each with receivers. It's easy enough to copy a declaration and forget to change the the receiver name. When diffing two implementations with similar code, you end up with unnecessary deltas b/c of different receiver names.

I don't think the "we have the same code everywhere" argument is that great, don't think we should optimize for that.

> 3. **Functional Output**: A core part of this recommendation is this statement: _"In Go, the receiver of a method is just another parameter and therefore, should be named accordingly."_ How often do you write functions that emit side effects through their parameters? The main exception to that is code similar to `json.Marshal` who returns its data through a pointer. On the other hand, side effects to a receiver are commonplace (f.e. `this.myProperty = "123"`). The fact that receivers are parameters under-the-hood is irrelevant considering the conventions are different.

I think this is handled by pointer vs non-pointers. You actually can't have a side-effect for non-pointers! And if it's a pointer, either way (param or receiver), you can have side-effects.

> 4. **Readability**: While reading code, you have to remember the receiver name to know if a variable is the receiver or not. If the receiver is always `this`, then that's one less bit of information to remember while reading someone else's code.

I think this is actually part of the intention behind the guideline - you don't have to know the difference! It's just another variable.

yields:
Kinda agree with Go's guidelines too since it's all over in there stdlib and the wider community.

francescaleung:
Just adding my two cents, but I'm leaning towards Go guidelines as well. I've used `this` in the past because it was already in the codebase ("code like no one knows you've been there") but it always felt weird to me.

joshma:
Let's wait for @colinking to get back, bikeshed over this, and resolve :)

colinking:
Heh thanks, but no need to have waited on me for this. Let's stick to Go conventions! Linting rules ftw

colinking:
@yields can we add this as a lint step in GHA?

yields:
> However, we have some `this` receivers in `airport` so now I'm curious why they didn't get flagged there... 🤔

Perhaps golangci-lint disables that check by default, I know that staticcheck allows users to disable any checks. but by default running the recent `staticcheck(1)` on this repo this is what I get:

```text
  λ cli (main) staticcheck ./...
pkg/taskdir/directory.go:69:7: receiver name should be a reflection of its identity; don't use generic names such as "this" or "self" (ST1006)
```

yields:
@colinking mind approving? :)

colinking (approved):

PR #143, opened 2021-06-01
init validation

yields:
Improves `init` command validation and error messages:

#### Missing `--slug`

Before:
```bash
$ ap init task.ts
Error: api: 404 - Could not find task
```

After:
```bash
$ ap init task.ts
Error: required flag(s) "slug" not set
```

#### Task not found

Before:
```bash
$ ap init --slug foo task.ts
Error: api: 404 - Could not find task
```

After:
```bash
$ ap init --slug foo task.ts

task with slug "foo" does not exist

Follow the URL below to create the task:
https://app.airplane.dev/tasks/new
```

### DIFF
--- a/pkg/api/errors.go
+++ b/pkg/api/errors.go
@@ -1,0 +1,22 @@
+package api
+
+import "fmt"
+
+// TaskMissingError implements an exaplainable error.
+type TaskMissingError struct {
+	app  string
+	slug string
+}
+
+// Error implementation.
+func (err TaskMissingError) Error() string {
+	return fmt.Sprintf("task with slug %q does not exist", err.slug)
+}
+
+// ExplainError implementation.
+func (err TaskMissingError) ExplainError() string {
+	return fmt.Sprintf(
+		"Follow the URL below to create the task:\n%s",
+		err.app+"/tasks/new",
+	)
+}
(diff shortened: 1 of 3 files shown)


pkg/api/errors.go
```diff
@@ -0,0 +1,22 @@
+package api
+
+import "fmt"
+
+// TaskMissingError implements an exaplainable error.
+type TaskMissingError struct {
+	app  string
```
joshma, line 7:
```suggestion
	appURL  string
```

pkg/api/errors.go
```diff
@@ -0,0 +1,22 @@
...
+	slug string
+}
+
+// Error implementation.
+func (err TaskMissingError) Error() string {
+	return fmt.Sprintf("task with slug %q does not exist", err.slug)
```
joshma, line 13: Should we capitalize error messages? Sort of tricky if we don't know if it'll get prefixed by something else I suppose...

```suggestion
	return fmt.Sprintf("Task with slug %q does not exist", err.slug)
```

joshma (approved):

pkg/api/errors.go
yields: Yeah, exactly why they're not supposed to be capitalized in go, this is why you see errors like:

```
net: <msg> net/http: <msg>
```

PR #145, opened 2021-06-01
Support .airplaneignore and improve default ignores

joshma:
- Add support for .airplaneignore file so that users can ignore without
  affecting docker
- Add `**/` prefixes to various paths that were intended to include
  subdirectories (e.g. `**/node_modules` to capture node_modules in
  subdirectories)
- Add `Debug` calls to help troubleshoot which files were being included


joshma:
@yields I'll CP this onto `main` after it gets onto `0.0`

PR #146, opened 2021-06-02
Calculate task root

yields:
Currently, we calculate the root on deploy. Upcoming PR will also
ask/suggest to create a package.json in the task dir and make it explicit
how deploy works.

To test this out you can create a directory structure like so:

```bash
monorepo/
  package.json
  my/task
  	index.js
```

The deploy command will automatically detect `package.json` and use monorepo as the root
of the task, it'll also update the entrypoint to execute `my/task/index.js`.


pkg/runtime/runtime.go
```diff
@@ -52,3 +62,27 @@ func Lookup(path string) (Interface, bool) {
...
 }
+
+const (
+	// Separator converted to string to abort pathof at root.
+	sep = string(filepath.Separator)
+)
+
+// Pathof attempts to find the path of the given filename.
+//
+// The method recursively descends the path until the given
```
joshma, line 73: "descends" might be confusing

```suggestion
// The method recursively visits parent dirs until the given
```

pkg/runtime/runtime.go
```diff
@@ -52,3 +62,27 @@ func Lookup(path string) (Interface, bool) {
...
+// filename is found, ok reports if the filename is found
+// and the string is the path.
+func Pathof(parent, filename string) (string, bool) {
+	var dst = filepath.Join(parent, filename)
```
joshma, line 77:
```suggestion
	dst := filepath.Join(parent, filename)
```

:)

joshma (approved):

PR #148, opened 2021-06-02
Adjust link comment

yields:
Previously github would ignore the task code because it detects
that it was "autogenerated", this is an attempt to change the task
special comment so that github does not hide the diff.


pkg/runtime/javascript/javascript.go
```diff
@@ -53,7 +53,7 @@ func (r Runtime) Slug(code []byte) (string, bool) {
 	for s.Scan() {
 		var line = strings.TrimSpace(s.Text())
 
-		if strings.HasPrefix(line, "// Generated by Airplane") {
+		if strings.HasPrefix(line, "// Linked to Airplane task:") {
```
joshma, line 56: Can you pull the prefix out to a constant, so it's easy to see that it's referenced somewhere else?

pkg/runtime/javascript/javascript.go
```diff
@@ -72,5 +72,5 @@ func (r Runtime) Slug(code []byte) (string, bool) {
 
 // Comment implementation.
 func (r Runtime) Comment(t api.Task) string {
-	return fmt.Sprintf("// Generated by Airplane, do not edit:\n// %s", t.URL)
+	return fmt.Sprintf("// Linked to Airplane task:\n// %s", t.URL)
```
joshma, line 75: I liked the "do not edit" - maybe "Linked to Airplane task, do not edit:"?

joshma (approved):

pkg/runtime/javascript/javascript.go
colinking: @yields @joshma wdyt of adding clarity on what shouldn't be edited?

```js
// https://app.airplane.dev/t/yolo [do not edit line]
```

Optionally, we could keep the first line as a clarifying comment, though our slug search logic can ignore it and just look for the first comment in a file that starts with `// https://app.airplane.dev/t/`.

```js
// The following line links this file to Airplane:
// https://app.airplane.dev/t/yolo [do not edit line]
```

joshma: I'm fine with either, what about one line?

    // Linked to https://app.airplane.dev/t/yolo [do not edit this line]

colinking: SGTM https://linear.app/airplane/issue/AIR-1230/adjust-link-comment

PR #149, opened 2021-06-02
Fix creating tasks from YAML

francescaleung:
Creating tasks from YAML was broken with the introduction of TaskMissingError; this needs to check for that instead of a generic `api.Error`.


yields (approved):
Nice catch!

PR #150, opened 2021-06-03
Create requirements.txt if missing

yields:
When building node tasks we automatically create package.json
if none was found, this pulls the same logic into the python
dockerfile creator.

### DIFF
--- a/pkg/build/python.go
+++ b/pkg/build/python.go
@@ -12,13 +12,16 @@ func python(root string, args Args) (string, error) {
 	var main = filepath.Join(root, entrypoint)
 	var reqs = filepath.Join(root, "requirements.txt")
 
-	if err := exist(reqs, main); err != nil {
+	if err := exist(main); err != nil {
 		return "", err
 	}
 
 	t, err := template.New("python").Parse(`
     FROM {{ .Base }}
     WORKDIR /airplane
+		{{if not .Requirements}}
+		RUN echo > requirements.txt
+		{{end}}
     COPY . .
     RUN pip install -r requirements.txt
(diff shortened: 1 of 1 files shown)


yields:
Once this is merged we can upgrade our remote builder CLI dependency and it should be fixed there as well ;)

pkg/build/python.go
```diff
@@ -12,13 +12,16 @@ func python(root string, args Args) (string, error) {
...
+	if err := exist(main); err != nil {
 		return "", err
 	}
 
 	t, err := template.New("python").Parse(`
     FROM {{ .Base }}
     WORKDIR /airplane
+		{{if not .Requirements}}
+		RUN echo > requirements.txt
+		{{end}}
```
joshma, line 24: Is the indentation here weird? Shouldn't it be outdented?

pkg/build/python.go
```diff
@@ -34,11 +37,13 @@ func python(root string, args Args) (string, error) {
 
 	var buf strings.Builder
 	if err := t.Execute(&buf, struct {
-		Base       string
-		Entrypoint string
+		Base         string
+		Entrypoint   string
+		Requirements bool
```
joshma, line 42: make it clearer it's a bool?

```suggestion
		RequirementsExist bool
```

joshma (approved):

pkg/build/python.go
yields: agree it's weird, but just follows the rest of the generators, we can clean them up later.

pkg/build/python.go
yields: Renamed to `HasRequirements`.

PR #151, opened 2021-06-03
Add runtime.Kind()

yields:
Previously we hardcoded api.TaskKindNode checks, this PR moves
that check to the runtime, that way adding a new runtime becomes
a bit easier.

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -41,6 +41,12 @@ type Interface interface {
 	// Typically runtimes will look for a specific file such as
 	// `package.json` or `requirements.txt`, they'll use `runtime.Pathof()`.
 	Root(path string) (dir string, ok bool)
+
+	// Kind returns a task kind that matches the runtime.
+	//
+	// Generate and other methods should not be called
+	// for a task that doesn't match the returned kind.
+	Kind() api.TaskKind
 }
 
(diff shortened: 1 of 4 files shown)


pkg/runtime/runtime.go
```diff
@@ -41,6 +41,12 @@ type Interface interface {
 	// Typically runtimes will look for a specific file such as
 	// `package.json` or `requirements.txt`, they'll use `runtime.Pathof()`.
 	Root(path string) (dir string, ok bool)
+
+	// Kind returns a task kind that matches the runtime.
+	//
+	// Generate and other methods should not be called
+	// for a task that doesn't match the returned kind.
```
joshma, line 48: minor but IDK if you need these sorts of comments - they tend to get stale as other code changes

joshma (approved):

pkg/runtime/runtime.go
yields: I think they're useful and show how the runtime should be used, e.g `Kind()` check should be done before `Generate()`.

yields: Typically I add unified tests too for all runtimes to verify they adhere to the interface, but since we don't have time comments should do it for now ;)

joshma: Ah OK - so you see this as part of the contract. Not opposed.

PR #152, opened 2021-06-03
Adjust REST JSONBody to be a string before sending to API server

francescaleung:
Sends REST jsonBody as `string` instead of `interface{}`. This does not "unpack" jsonBody as `interface{}` when fetching from the API.


pkg/taskdir/definitions/definitions.go
```diff
@@ -124,6 +125,14 @@ func (this Definition) GetKindAndOptions() (api.TaskKind, api.KindOptions, error
 		if err := mapstructure.Decode(this.REST, &options); err != nil {
 			return "", api.KindOptions{}, errors.Wrap(err, "decoding REST definition")
 		}
+		// API server expects JSONBody as a string.
+		if _, ok := options["jsonBody"].(string); options["jsonBody"] != nil && !ok {
```
joshma, line 129: Not sure I follow - why isn't jsonBody a string in the first place?

joshma (changes requested):

pkg/taskdir/definitions/definitions.go
francescaleung: As long as it's valid JSON (no leaf nodes are entirely templated booleans or numbers), you can write your jsonBody as structured yaml instead of as a string.
```
jsonBody:
  foo:
    - bar
    - baz
    - {{buz}}
```
is functionally equivalent to
```
body: '{"foo": ["bar", "baz", "{{buz}}"]}'
headers:
  - Content-Type: application/json
```

joshma: Hm I see, this is a CLI/YAML convenience. Can you clarify the comment then? Something like:

```
// API expects jsonBody to be a string, since it's handlebars-templated JSON and not always valid JSON.
// For convenience, we allow the YAML definition to be a structured object when the jsonBody is actually valid JSON.
// In that case, if it's not a string, we JSON-serialize it into a string.
```

joshma, line 129: minor suggestion, check `!ok` first:

```suggestion
		if _, ok := options["jsonBody"].(string); !ok && options["jsonBody"] != nil {
```

joshma (approved):

PR #153, opened 2021-06-04
Always apply the same default ignore rules

joshma:
Unify all airplaneignore rules regardless of kind - we want to ignore
e.g. pyc files even if we're uploading a JS task.

.airplaneignore and .dockerignore files ADD to the default rules rather
than REPLACE them. Otherwise, if you add a single rule in
.airplaneignore, you end up having to replicate all of the default
exclude rules (or at least the relevant ones).

### DIFF
--- a/pkg/build/remote.go
+++ b/pkg/build/remote.go
@@ -222,6 +222,7 @@ func getIgnorePatterns(path string) ([]string, error) {
 			return excludes, nil
 		}
 	}
+	return excludes, nil
 }
 
 func uploadArchive(ctx context.Context, client *api.Client, archivePath string) (string, error) {
--- a/pkg/build/remote.go
+++ b/pkg/build/remote.go
@@ -107,7 +107,7 @@ func archiveTaskDir(def definitions.Definition, root string, archivePath string)
 // This is modeled off of docker/cli.
 // See: https://github.com/docker/cli/blob/a32cd16160f1b41c1c4ae7bee4dac929d1484e59/vendor/github.com/docker/docker/pkg/archive/archive.go#L738
 func getIgnoreFunc(taskRootPath string, kind api.TaskKind) (func(filePath string, info os.FileInfo) (bool, error), error) {
-	excludes, err := getIgnorePatterns(taskRootPath, kind)
+	excludes, err := getIgnorePatterns(taskRootPath)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


joshma:
This can wait for @colinking to get back, but @yields your eyes are also welcome!

pkg/build/remote.go
```diff
@@ -179,59 +179,49 @@ func readIgnorefile(contextDir, filename string) ([]string, error) {
...
-func getIgnorePatterns(path string, kind api.TaskKind) ([]string, error) {
+func getIgnorePatterns(path string) ([]string, error) {
+	// Start with default set of excludes.
+	// We exclude the same files regardless of kind because you might have both JS and PY tasks and
+	// want pyc files excluded just the same.
+	// For inspiration, see:
+	// https://github.com/github/gitignore
+	// https://github.com/github/gitignore/blob/master/Go.gitignore
+	// https://github.com/github/gitignore/blob/master/Node.gitignore
+	excludes := []string{
```
yields, line 190: if you `[...]string{...}` (turn it into an array) it'll be allocated on the stack and be GCed right after the func is called, not a huge deal here obviously tho.

yields (approved):
lgtm

pkg/build/remote.go
joshma: hm not sure i understand - what should I do here instead?

pkg/build/remote.go
```diff
@@ -179,59 +179,50 @@ func readIgnorefile(contextDir, filename string) ([]string, error) {
...
+		".git",
+		".gitmodules",
+		".hg",
+		".svn",
+	}
+
+	// Allow user-specified ignore file. Note that users can re-INCLUDE files using !, so if our
+	// default excludes skip something necessary they can always add it back.
+
 	// Prefer .airplaneignore over .dockerignore
```
colinking, line 213: Thoughts on applying both?

joshma: Hm I think I'm a no here - you might have a dockerignore for something unrelated, and to prevent it from affecting airplane you could create an empty airplaneignore

colinking: But the `.dockerignore` will get applied remotely before the build. Why send those files across the wire?

joshma: Ah no I mean - let's say you were building some docker image for `./some-alpine-build` and so you decided to add `.dockerignore`:

```
# Only include ./some-alpine-build
*
!./some-alpine-build
!entrypoint.sh
```

And then you decide to use Airplane - you have a `.dockerignore` but it doesn't represent your desire for what gets included for Airplane tasks. So in that case you'd have to delete the .dockerignore or modify it to make Airplane work.

The broader issue is actually that, given we're not really emphasizing docker itself anymore, the fact that cli looks at dockerignore at all could be confusing?

colinking: Ah, hmm. Okay I'm onboard with not processing `.dockerignore` files. Less to hide Docker, more about how a `.dockerignore` usually correlates with a specific `Dockerfile` (see f.e. how recent versions of Docker now support `<dockerfilename>.dockerignore` files) and we end up generating the Dockerfile ourselves.

The one exception would be for BYODockerfile Airplane tasks. For that, I think we would want to apply both the `.dockerignore` (or `<dockerfilename>.dockerignore` instead, if present) + `.airplaneignore`. wdyt?

joshma: Ahh yep agreed! Filed https://linear.app/airplane/issue/AIR-1232/revive-dockerfile-builder :)

PR #154, opened 2021-06-04
Rename "args" to "params" in init-ed script

joshma:
This is more consistent with our UI ("parameters")

### DIFF
--- a/pkg/runtime/javascript/javascript.go
+++ b/pkg/runtime/javascript/javascript.go
@@ -26,8 +26,8 @@ const (
 // Code template.
 var code = template.Must(template.New("js").Parse(`{{.Comment}}
 
-export default async function(args){
-  console.log('arguments: ', args);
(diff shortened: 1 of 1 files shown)


yields (approved):

pkg/runtime/javascript/javascript.go
```diff
@@ -26,8 +26,8 @@ const (
 // Code template.
 var code = template.Must(template.New("js").Parse(`{{.Comment}}
 
-export default async function(args){
-  console.log('arguments: ', args);
+export default async function(params){
+  console.log('parameters: ', params);
```
colinking, line 30: FYI we'll want to rename the TS runtime too: https://github.com/airplanedev/cli/blob/650668ce401bfaf093e766bf8f1d9a8405bee8da/pkg/runtime/typescript/typescript.go#L27-L28

PR #156, opened 2021-06-04
Error when task is not deployed

yields:
Previously, if you attempt to execute a task that was not deployed
yet the CLI will ask you for parameters and attempt to execute the task
even though it was not deployed.

After this PR, the CLI will error asking you to deploy it.

```bash
  λ tasks apdev exec ./test_js/index.js               

task ./test_js/index.js was not deployed

to deploy the task:
	airplane deploy ./test_js/index.js
```

### DIFF
--- a/pkg/cmd/tasks/execute/execute.go
+++ b/pkg/cmd/tasks/execute/execute.go
@@ -73,6 +73,12 @@ func run(ctx context.Context, cfg config) error {
 		return errors.Wrap(err, "get task")
 	}
 
+	if task.Image == nil {
+		return &notDeployedError{
+			task: cfg.task,
+		}
+	}
+
 	req := api.RunTaskRequest{
 		TaskID:      task.ID,
 		ParamValues: make(api.Values),
@@ -237,4 +243,18 @@ func slugFromScript(file string) (string, error) {
 	}
 
 	return slug, nil
+}
+
+type notDeployedError struct {
+	task string
+}
+
+// Error implementation.
+func (err notDeployedError) Error() string {
+	return fmt.Sprintf("task %s was not deployed", err.task)
+}
+
+// ExplainError implementation.
+func (err notDeployedError) ExplainError() string {
+	return fmt.Sprintf("to deploy the task:\n\tairplane deploy %s", err.task)
 }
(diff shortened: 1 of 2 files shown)


pkg/cmd/tasks/execute/execute.go
```diff
@@ -238,3 +244,17 @@ func slugFromScript(file string) (string, error) {
...
 	return slug, nil
 }
+
+type notDeployedError struct {
+	task string
+}
+
+// Error implementation.
+func (err notDeployedError) Error() string {
+	return fmt.Sprintf("task %s was not deployed", err.task)
```
joshma, line 254: minor wording tweak + I think error messages user see should be capitalized

```suggestion
	return fmt.Sprintf("Task %s has not been deployed", err.task)
```

pkg/cmd/tasks/execute/execute.go
```diff
@@ -238,3 +244,17 @@ func slugFromScript(file string) (string, error) {
...
+}
+
+// ExplainError implementation.
+func (err notDeployedError) ExplainError() string {
+	return fmt.Sprintf("to deploy the task:\n\tairplane deploy %s", err.task)
```
joshma, line 259: Are we using `\t` everywhere? Can we just do two spaces for now, to match https://github.com/airplanedev/cli/blob/main/pkg/cmd/tasks/initcmd/init.go#L139

```suggestion
	return fmt.Sprintf("To deploy the task:\n  airplane deploy %s", err.task)
```

joshma (approved):

pkg/cmd/tasks/execute/execute.go
yields: Let's capitalize when we render instead.

pkg/cmd/tasks/execute/execute.go
yields: Same as the comment above.

yields: just to avoid lint errors btw, also cleanliness

joshma: huh interesting (https://github.com/golang/go/wiki/CodeReviewComments#error-strings for reference) - cc @colinking @francescaleung on this, so we're consistent.

(minor note - was the `\t` intentionally left behind? can we use spaces instead?)

yields: ah yikes, yeah I'll fix it right away.

PR #157, opened 2021-06-04
Support node.js workspaces

yields:
The PR adds support to `node.js` workspaces, it does this by allowing users
to define an "airplane" settings key within package.json:

```json
{
  "airplane": {
     "root": ".."
  }
}
```

There's an [example](https://github.com/airplanedev/examples/pull/10) of this working and some guide on how to it works, but
i'll mention it again here. Users define `airplane.root` key within the nearest
package.json, on deploy we detect that key and ensure we treat it as the root.

From there our node.js dockerfile does most of the work by calling `yarn` or `npm`
to install all dependencies and the task is executed successfully.

### DIFF
--- a/pkg/build/node.go
+++ b/pkg/build/node.go
@@ -27,6 +27,7 @@ func node(root string, args Args) (string, error) {
 	}
 
 	cfg := struct {
+		Workdir        string
 		Base           string
 		HasPackageJSON bool
 		HasPackageLock bool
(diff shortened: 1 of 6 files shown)


joshma (commented):
Doesn't this change the working directory of the node shim, etc.? I think `tsc` ends up getting called from the parent directory, right? I think this needs to distinguish between "what is included in the bundle" and "where the working directory is".

yields:
> Doesn't this change the working directory of the node shim, etc.? I think `tsc` ends up getting called from the parent directory, right? I think this needs to distinguish between "what is included in the bundle" and "where the working directory is"

I adjusted the PR to add a runtime `.Workdir()`, now the docker file looks like this (simplified example):

```docker file
WORKDIR /app/server
COPY /app /
```

joshma (approved):

PR #161, opened 2021-06-08
Clarify `do not edit` nature of linking comment

colinking:
This PR updates the comment we attach to JS files that links that file with an Airplane task to use the copy from: https://github.com/airplanedev/cli/pull/148#discussion_r647680710

While here, I also extracted the slug detection logic from the JS runtime and tweaked it to handle linking comments that are not on the first line of the script file. This should partially resolve: https://linear.app/airplane/issue/AIR-1228/detect-missing-slug-comment

### DIFF
--- a/pkg/runtime/comment.go
+++ b/pkg/runtime/comment.go
@@ -1,0 +1,41 @@
+package runtime
+
+import (
+	"net/url"
+	"path"
+	"regexp"
+
+	"github.com/airplanedev/cli/pkg/api"
+)
+
+var (
+	// commentRegex matches against the string produced by Comment() below.
+	//
+	// It is used to extract a slug from a comment in a script file.
+	commentRegex = regexp.MustCompile(`Linked to (https://.*air.*/t/.*) \[do not edit this line\]`)
+)
+
+// Comment generates a linking comment that is used
+// to associate a script file with an Airplane task.
+//
+// This comment can be parsed out of a script file using
+// ExtractSlugFromComment.
+func Comment(task api.Task) string {
+	return "Linked to " + task.URL + " [do not edit this line]"
+}
+
+func ExtractSlugFromComment(code []byte) (string, bool) {
+	result := commentRegex.FindSubmatch(code)
+	if len(result) == 0 {
+		return "", false
+	}
+
+	u, err := url.Parse(string(result[1]))
+	if err != nil {
+		return "", false
+	}
+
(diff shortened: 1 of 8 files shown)


pkg/runtime/comment.go
```diff
@@ -0,0 +1,41 @@
...
+// Comment generates a linking comment that is used
+// to associate a script file with an Airplane task.
+//
+// This comment can be parsed out of a script file using
+// ExtractSlugFromComment.
+func Comment(task api.Task) string {
+	return "Linked to " + task.URL + " [do not edit this line]"
+}
+
+func ExtractSlugFromComment(code []byte) (string, bool) {
```
yields, line 27: Let's add a comment and rename this into `runtime.Slug()` to match `runtime.Comment()`? it reads better imho, all we know is that this func receives source code and returns a slug.

This kind of thing matches the standard library too btw.

pkg/runtime/comment_test.go
```diff
@@ -0,0 +1,69 @@
+package runtime
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestExtractSlugFromComment(tt *testing.T) {
```
yields, line 9: 😍

pkg/runtime/runtime.go
```diff
@@ -35,17 +35,6 @@ type Interface interface {
 	// An error is returned if the code cannot be generated.
 	Generate(task api.Task) ([]byte, error)
 
-	// Slug returns the slug from the given code.
-	//
-	// If the comment was not found in code, the method
-	// returns empty string and false.
-	Slug(code []byte) (string, bool)
```
yields, line 42: sidenote, I think the runtimes just need a `LineComment() => "#" / "//"` and we'll be fine with:

```go
runtime.SlugOf(r, code)
runtime.CommentFor(r, task)
```

yields (approved):
LGTM! :) just a few comments on naming.

pkg/runtime/runtime.go
colinking: I considered that, but we don't need to match on the comment to be accurate (it's English after all, it wouldn't be valid code). And I want to keep the `runtime` interface as slim as possible.

yields: Right, but it'll break patching an existing task otherwise? currently it seems to be broken since there's no `# or //` when updating a task file.

PR #163, opened 2021-06-08
Bring back `--file/-f` flag to `deploy/execute` as a deprecated flag

colinking:
This PR brings back the `--file/-f` flag to the `deploy` and the `execute` commands. This was previously going to be a breaking change between `0.0.x` and `0.1.x`, but we want to reduce the incompatibility between these releases.

When used, it will log a warning message:

image

While here, I also fixed an issue with `execute`-ing slugs that match a local file. I also added the short/long command descriptions to our custom command help text.

### DIFF
--- a/pkg/cmd/tasks/deploy/deploy.go
+++ b/pkg/cmd/tasks/deploy/deploy.go
@@ -49,7 +49,7 @@ func New(c *cli.Config) *cobra.Command {
 		}),
 	}
 
-	cmd.Flags().BoolVarP(&cfg.local, "local", "L", false, "use a local (instead of Airplane-hosted) Docker daemon")
+	cmd.Flags().BoolVarP(&cfg.local, "local", "L", false, "use a local Docker daemon (instead of an Airplane-hosted builder)")
 	cmd.Flags().StringVarP(&cfg.file, "file", "f", "", "File to deploy (.yaml, .yml, .js, .ts)")
 	cli.Must(cmd.Flags().MarkHidden("file")) // --file is deprecated
 
--- a/pkg/cmd/tasks/deploy/deploy.go
+++ b/pkg/cmd/tasks/deploy/deploy.go
@@ -8,7 +8,9 @@ import (
 	"github.com/airplanedev/cli/pkg/api"
 	"github.com/airplanedev/cli/pkg/cli"
 	"github.com/airplanedev/cli/pkg/cmd/auth/login"
+	"github.com/airplanedev/cli/pkg/logger"
 	"github.com/airplanedev/cli/pkg/utils"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -24,24 +26,32 @@ func New(c *cli.Config) *cobra.Command {
...
+			}
 			return run(cmd.Root().Context(), cfg)
 		},
 		PersistentPreRunE: utils.WithParentPersistentPreRunE(func(cmd *cobra.Command, args []string) error {
 			return login.EnsureLoggedIn(cmd.Root().Context(), c)
 		}),
 	}
 
-	cmd.Flags().BoolVarP(&cfg.local, "local", "L", false, "Add to build the docker image locally.")
+	cmd.Flags().BoolVarP(&cfg.local, "local", "L", false, "use a local (instead of Airplane-hosted) Docker daemon")
```
joshma, line 52: eh?

```suggestion
	cmd.Flags().BoolVarP(&cfg.local, "local", "L", false, "use a local (instead of Airplane-hosted) Docker builder")
```

pkg/cmd/tasks/execute/execute.go
```diff
@@ -190,19 +207,20 @@ func flagset(task api.Task, args api.Values) *flag.FlagSet {
...
-			return slugFromYaml(file)
-		case ".js", ".ts":
-			return slugFromScript(file)
-		case "":
-			return "", fmt.Errorf("the file %s must have an extension", file)
-		default:
-			return "", fmt.Errorf("the file %s has unrecognized extension", file)
-		}
+	if !fs.Exists(file) {
+		return "", errors.Errorf("Unable to execute %s. No matching file or task slug.", file)
```
joshma, line 211:
```suggestion
		return "", errors.Errorf("Unable to execute %s - file not found.", file)
```

pkg/cmd/tasks/execute/execute.go
```diff
@@ -38,38 +38,55 @@ func New(c *cli.Config) *cobra.Command {
...
 
 // Run runs the execute command.
 func run(ctx context.Context, cfg config) error {
 	var client = cfg.root.Client
 
-	slug, err := slugFrom(cfg.task)
-	if err != nil {
-		return err
-	}
+	// cfg.task is either a slug or a local path. Try it as a slug first.
```
joshma, line 76: OOC does this actually happen ever? was this something we ran into?

joshma (approved):

pkg/cmd/tasks/execute/execute.go
colinking: Yeah, I ran into it while testing. I had a folder called `outputs` that included the code for a task with a slug `outputs` 😅

pkg/cmd/tasks/execute/execute.go
colinking: Isn't this error confusing if you're trying to execute a task by slug and don't provide the correct slug?

pkg/cmd/tasks/deploy/deploy.go
colinking: I'm using `daemon` to match Docker terminology here. Any reason to use builder instead?

joshma: daemon just felt overly technical idk - would rather folks think of this as a "local builder" vs "remote builder" and nothing docker-specific. maybe even "use a local Docker daemon (instead of an Airplane-hosted builder)"

pkg/cmd/tasks/execute/execute.go
joshma: it's just weird because this function only checks for the file, and assumes the slug has been checked already

PR #164, opened 2021-06-09
Rewrite ignore logic using gitignore library

joshma:
Drops support for .dockerignore for now, although we may re-add support
for that with dockerfile support

Replaces with (1) default ignores like before, but in gitignore-syntax
and (2) support for .airplaneignore like before, but in
gitignore-syntax.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,6 +23,7 @@ require (
 	github.com/moby/sys/mount v0.2.0 // indirect
 	github.com/olekukonko/tablewriter v0.0.5
 	github.com/pkg/errors v0.9.1
+	github.com/sabhiram/go-gitignore v0.0.0-20201211210132-54b8a0bf510f
 	github.com/segmentio/events/v2 v2.4.0
 	github.com/spf13/cobra v1.1.3
 	github.com/stretchr/testify v1.6.1
--- a/pkg/build/ignore/ignore.go
+++ b/pkg/build/ignore/ignore.go
@@ -1,0 +1,111 @@
+package ignore
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/airplanedev/cli/pkg/api"
+	"github.com/airplanedev/cli/pkg/logger"
+	"github.com/pkg/errors"
+	gitignore "github.com/sabhiram/go-gitignore"
+)
+
+// Returns an IgnoreFunc that can be used with airplanedev/archiver to filter
+// out files that match a default list or user-provided .airplaneignore.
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -23,6 +23,7 @@ require (
 	github.com/moby/sys/mount v0.2.0 // indirect
 	github.com/olekukonko/tablewriter v0.0.5
 	github.com/pkg/errors v0.9.1
+	github.com/sabhiram/go-gitignore v0.0.0-20201211210132-54b8a0bf510f
```
joshma, line 26: This was the best library I could find...open to suggestions.

pkg/build/ignore/ignore.go
```diff
@@ -0,0 +1,111 @@
+package ignore
```
joshma, line 1: Pulled out an `ignore` sub-package

pkg/build/ignore/ignore.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"github.com/airplanedev/cli/pkg/api"
+	"github.com/airplanedev/cli/pkg/logger"
+	"github.com/pkg/errors"
+	gitignore "github.com/sabhiram/go-gitignore"
+)
+
+// Returns an IgnoreFunc that can be used with airplanedev/archiver to filter
+// out files that match a default list or user-provided .airplaneignore.
+func GetIgnoreFunc(taskRootPath string, kind api.TaskKind) (func(filePath string, info os.FileInfo) (bool, error), error) {
```
joshma, line 17: Oops TODO: remove `kind` arg, unused

pkg/build/ignore/ignore.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+
+		skip := ig.MatchesPath(relFilePath)
+
+		// If we want to skip this file, and it's a directory, then we can skip it only if there
+		// are no inclusions.
+		// Ideally, we are smarter about this and skip the dir if there's no way for the inclusion
+		// pattern to match the dir. However, this is tricky with things like "!foo/" matching
+		// any directory named foo/ inside this directory.
+		if info.IsDir() && skip && !hasInclusion {
```
joshma, line 52: This is less powerful now but still correct - we just don't know what inclusions can actually do so we have to be more conservative

go.mod
yields: https://github.com/denormal/go-gitignore

joshma: @yields are you saying that one is better? hasn't been updated since 2018, although I suppose it's not necessarily something that _has_ to be updated periodically

yields (approved):

go.mod
yields: Yeah sorry should've added more details, it just seems to have a nicer API. either one works :)

PR #166, opened 2021-06-09
Set shim=true on scripts deploy

yields:
That way we do not override yaml config and use a shim there
also adds shim=true to all runtimes.


colinking (approved):
Much cleaner, nice!

PR #167, opened 2021-06-09
Tweak shim toggle behavior

colinking:
This is a follow-up to: https://github.com/airplanedev/cli/pull/166

Previously, we would pass a boolean to the remote build API, but this approach avoids API changes.

### DIFF
--- a/pkg/cmd/tasks/deploy/script.go
+++ b/pkg/cmd/tasks/deploy/script.go
@@ -81,6 +81,8 @@ func deployFromScript(ctx context.Context, cfg config) error {
 		return err
 	}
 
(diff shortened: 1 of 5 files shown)


joshma (approved):

pkg/cmd/tasks/deploy/script.go
```diff
@@ -81,6 +81,10 @@ func deployFromScript(ctx context.Context, cfg config) error {
 		return err
 	}
 
+	// Instruct the remote builder API to perform a shim-based build since
+	// we're deploying from a script.
+	kindOptions["shim"] = "true"
```
yields, line 86: 💥

PR #170, opened 2021-06-10
Add python runtime

yields:
Initial implementation of Python runtime, it's inspired by the simplicity
of the node.js shim.

Note that the shim is python3.5+ because of module loading, i don't think
it's a big issue because it matches our image version. We can iterate on
this later and add python 2 support if needed.

### DIFF
--- a/pkg/build/python.go
+++ b/pkg/build/python.go
@@ -1,19 +1,88 @@
 package build
 
 import (
+	_ "embed"
 	"path/filepath"
 	"strings"
 	"text/template"
 
 	"github.com/airplanedev/cli/pkg/api"
 	"github.com/airplanedev/cli/pkg/fsx"
+	"github.com/pkg/errors"
 )
 
+// Python shim code.
+//go:embed shim.py
+var pythonShim string
+
 // Python creates a dockerfile for Python.
-func python(root string, options api.KindOptions) (string, error) {
-	entrypoint, _ := options["entrypoint"].(string)
+func python(root string, args api.KindOptions) (string, error) {
+	entrypoint, _ := args["entrypoint"].(string)
 	main := filepath.Join(root, entrypoint)
+	init := filepath.Join(root, "__init__.py")
 	reqs := filepath.Join(root, "requirements.txt")
+
+	if args["shim"] != "true" {
+		return pythonLegacy(root, args)
+	}
+
+	if err := fsx.AssertExistsAll(main); err != nil {
+		return "", err
+	}
+
+	v, err := GetVersion(NamePython, "3")
+	if err != nil {
+		return "", err
+	}
+
+	shim, err := applyTemplate(pythonShim, struct {
+		Entrypoint string
+	}{
+		Entrypoint: filepath.Join("/airplane", entrypoint),
+	})
+	if err != nil {
+		return "", errors.Wrapf(err, "rendering shim")
+	}
+
+	const dockerfile = `
+    FROM {{ .Base }}
+    WORKDIR /airplane
+    RUN echo '{{.Shim}}' > /shim.py
+    {{if not .HasInit}}
+    RUN touch __init__.py
+    {{end}}
+    COPY . .
+		{{if .HasRequirements}}
+    RUN pip install -r requirements.txt
+		{{end}}
+    ENTRYPOINT ["python", "/shim.py", "/airplane/{{ .Entrypoint }}"]
+	`
+
+	df, err := applyTemplate(dockerfile, struct {
+		Base            string
+		Shim            string
+		Entrypoint      string
+		HasRequirements bool
+		HasInit         bool
+	}{
+		Base:            v.String(),
+		Shim:            strings.Join(strings.Split(shim, "\n"), "\\n\\\n"),
+		Entrypoint:      entrypoint,
+		HasRequirements: fsx.Exists(reqs),
+		HasInit:         fsx.Exists(init),
+	})
+	if err != nil {
+		return "", errors.Wrapf(err, "rendering dockerfile")
+	}
+
+	return df, nil
+}
+
+// PythonLegacy generates a dockerfile for legacy python support.
+func pythonLegacy(root string, args api.KindOptions) (string, error) {
+	var entrypoint, _ = args["entrypoint"].(string)
+	var main = filepath.Join(root, entrypoint)
+	var reqs = filepath.Join(root, "requirements.txt")
 
 	if err := fsx.AssertExistsAll(main); err != nil {
 		return "", err
--- a/pkg/build/shim.py
+++ b/pkg/build/shim.py
@@ -1,0 +1,22 @@
+# This file includes a shim that will execute your task code.
+
+import importlib.util as util
+import json
+import sys
+import os
+
+def run(args):
+	if len(args) != 3:
+		raise Exception("shim: expected 3 arguments, got {}".format(args))
+
+	spec = util.spec_from_file_location("mod.main", "{{ .Entrypoint }}")
+	mod = util.module_from_spec(spec)
+	spec.loader.exec_module(mod)
+
+	try:
+		mod.main(json.loads(args[2]))
+	except Exception as e:
+		raise Exception("shim: executing {{.Entrypoint}}") from e
+
+if __name__ == "__main__":
+	run(sys.argv)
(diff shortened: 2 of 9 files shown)


pkg/build/shim.py
```diff
@@ -0,0 +1,21 @@
...
+		raise Exception("shim: expected 3 arguments, got {}".format(args))
+
+	spec = util.spec_from_file_location("mod.main", "{{ .Entrypoint }}")
+	mod = util.module_from_spec(spec)
+	spec.loader.exec_module(mod)
+
+	try:
+		mod.main(json.loads(args[2]))
+	except e:
+		raise Exception("shim: executing {{.Entrypoint}} - {}".format(e))
```
joshma, line 18: You use `raise from` https://stackoverflow.com/questions/9157210/how-do-i-raise-the-same-exception-with-a-custom-message-in-python/29442282#29442282

```suggestion
		raise Exception("shim: executing {{.Entrypoint}}") from e
```

pkg/build/shim.py
```diff
@@ -0,0 +1,21 @@
+
+import importlib.util as util
+import json
+import sys
+import os
+
+def run(args):
+	if len(args) != 3:
```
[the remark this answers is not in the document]
joshma: Oops nevermind, ignore me 😅

pkg/build/python.go
```diff
@@ -1,17 +1,83 @@
...
+	`
+
+	df, err := templatize(dockerfile, struct {
+		Base            string
+		Shim            string
+		Entrypoint      string
+		HasRequirements bool
+	}{
+		Base:            v.String(),
+		Shim:            strings.Join(strings.Split(shim, "\n"), "\\n\\\n"),
```
colinking, line 64: wdyt of pulling this `\n\` formatting out into a function shared with JS?

pkg/build/shim.py
```diff
@@ -0,0 +1,21 @@
+
```
colinking, line 1: Could we drop up this leading line?

colinking, line 1: Could we add a comment here? Similar to JS: https://github.com/airplanedev/cli/blob/6336f8ab85b43b42bde1a6d9518477b474f4b6fd/pkg/build/node.go#L73

```
# This file includes a shim that will execute your task code.
```

pkg/build/shim.py
```diff
@@ -0,0 +1,21 @@
...
+import sys
+import os
+
+def run(args):
+	if len(args) != 3:
+		raise Exception("shim: expected 3 arguments, got {}".format(args))
+
+	spec = util.spec_from_file_location("mod.main", "{{ .Entrypoint }}")
+	mod = util.module_from_spec(spec)
+	spec.loader.exec_module(mod)
```
colinking, line 13: fancy

pkg/cmd/tasks/deploy/script.go
```diff
@@ -66,14 +66,17 @@ func deployFromScript(ctx context.Context, cfg config) error {
...
+		setEntrypoint(&def, filepath.Base(abs))
 	}
 
-	if wd, err := r.Workdir(abs); err == nil {
-		def.Node.Workdir = strings.TrimPrefix(wd, taskroot)
+	// TODO(amir): move to `d.SetWorkdir()`.
+	if def.Node != nil {
+		if wd, err := r.Workdir(abs); err == nil {
+			def.Node.Workdir = strings.TrimPrefix(wd, taskroot)
+		}
```
colinking, line 79: wdyt of passing `Workdir` through in the `Request` struct instead? Think it's fine that only Node would look at it for now.

pkg/cmd/tasks/deploy/script.go
```diff
@@ -146,3 +149,15 @@ func (u unlinked) ExplainError() string {
...
+// SetEntrypoint sets the entrypoint on d.
+//
+// TODO(amir): move this to `def.SetEntrypoint()` or whatever.
+func setEntrypoint(d *definitions.Definition, ep string) {
+	switch kind, _, _ := d.GetKindAndOptions(); kind {
+	case api.TaskKindNode:
+		d.Node.Entrypoint = ep
+	case api.TaskKindPython:
+		d.Python.Entrypoint = ep
+	}
```
colinking, line 162: How would you feel about `panic`-ing here in a `default` case?

pkg/runtime/python/python.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/airplanedev/cli/pkg/runtime"
+)
+
+// Init register the runtime.
+func init() {
+	runtime.Register(".py", Runtime{})
+}
+
+// Code template.
+var code = template.Must(template.New("py").Parse(`# {{.Comment}}
```
colinking, line 18: ^ noting that you'll need to add a `FormatComment` implementation from: https://github.com/airplanedev/cli/pull/172/files#diff-03794acab66295250e50c8ecc7a7d6371142b27407bce017384597dbe8341276R91

And remove the `#` here

pkg/cmd/tasks/execute/execute.go
```diff
@@ -214,11 +202,12 @@ func slugFrom(file string) (string, error) {
 	switch ext := filepath.Ext(file); ext {
 	case ".yml", ".yaml":
 		return slugFromYaml(file)
-	case ".js", ".ts":
-		return slugFromScript(file)
 	case "":
 		return "", fmt.Errorf("the file %s must have an extension", file)
 	default:
+		if _, ok := runtime.Lookup(file); ok {
```
colinking, line 208: 🔥

pkg/cmd/tasks/execute/execute.go
```diff
@@ -73,25 +72,14 @@ func New(c *cli.Config) *cobra.Command {
...
-		}
-
-		slug, err := slugFrom(cfg.task)
-		if err != nil {
-			return err
-		}
+	slug, err := slugFrom(cfg.task)
+	if err != nil {
+		return err
+	}
```
colinking, line 78: ^ looks like a bad merge

colinking (changes requested):
Looking good! Some merge-related comments

pkg/cmd/tasks/execute/execute.go
yields: hmm that was intentional, I removed the previous code snippet because it would fetch the task twice if it's a file...

pkg/cmd/tasks/deploy/script.go
yields: hmm, what's the benefit of doing so?

pkg/cmd/tasks/execute/execute.go
colinking: Yeah, but that handles the case where you have a file with the same name as the slug. Which is common enough in the case where you place your task in a directory with the same name as the slug and then try and execute that task from the parent directory.

For example:

```
outputs/
  main.js
```

```
airplane execute ./outputs/main.js
// vs
airplane execute outputs
```

Maybe a better fix would be for us to conditionally treat it as a slug/file if it references a _file_ (not a directory)?

yields: AFAIK, `airplane execute outputs` would treat it as a slug and error? if not that should be a bug hmm

colinking: The current behavior would see the outputs directory and try to deploy from script, even though we want to instead deploy the task with `slug=outputs`. If we can notice it is a directory and then skip straight to trying it as a slug then that should help.

pkg/build/python.go
```diff
@@ -1,20 +1,89 @@
...
 )
 
+// Python shim code.
+//go:embed shim.py
+var pythonShim string
+
 // Python creates a dockerfile for Python.
-func python(root string, options api.KindOptions) (string, error) {
-	entrypoint, _ := options["entrypoint"].(string)
+func python(root string, args api.KindOptions) (string, error) {
```
colinking, line 19: FYI moved these functions to use `api.KindOptions`, so I had renamed `args -> options`

pkg/runtime/runtime.go
```diff
@@ -62,7 +68,16 @@ type Interface interface {
...
+	// PrepareRun should prepare a local run of a task.
+	//
+	// It must create a temporary directory, install any dependencies
+	// and prepare the script to be run.
+	//
+	// On success the method returns a slice that represents an `cmd.Exec`
+	// options which contains the command to be run and its arguments.
+	//
+	// If running the script locally is not supported the method returns
+	// an `ErrNotImplemented`.
```
colinking, line 80: lmao.... Thanks for fixing this comment!

colinking (approved):

PR #173, opened 2021-06-10
Add `airplane dev` command for local task execution

colinking:
This PR adds a concept of local task execution via the new `airplane dev` command. Currently, this PR only supports Node, but we'll add support for Python in a future PR after https://github.com/airplanedev/cli/pull/170 lands.

I landed on a separate command (vs. `execute --remote`) for a couple reasons:
1. I'd like for `execute/dev` to (mostly) require zero flags. This leaves the door open to dropping `--` (f.e. Yarn dropped this in 1.0: https://github.com/yarnpkg/website/issues/666). Can't do much about `-o json` / `--debug`, but we could enforce those come before positional arguments for `dev/execute`. TBD for now.
2. The existing `airplane execute` supports execution [by slug, script, and YML file](https://github.com/airplanedev/cli/pull/138#discussion_r641829470). If we introduce a `--remote` flag, then we'd also want to support all three with local development. However, neither slugs nor YML file make sense with local execution. The former because we don't associate slugs with local files, and the latter because we don't support shim-based tasks with YML.

However, I'd love to find better command terminology than `dev` and `execute`. To me, `dev` implies "watch"-style behavior. For example, `next dev` vs. `next build`.

```sh
# now
airplane dev ./file.ts
airplane exec outputs

# proposal 1: different verbs
airplane test ./file.ts
airplane trigger outputs

# proposal 2: remove the top-level alias for remote execution
airplane [dev|test] ./file.ts
airplane runs [create|execute|trigger] outputs

# proposal 3: just merge them
airplane exec ./file.ts
# --remote must be passed before the first positional arg. Everything after the first
# positional arg is considered a task param.
airplane exec --remote outputs [...task parameters]

# proposal 4: just merge them, but slugs == remote, files == local (tbh, seems confusing to me)
airplane exec ./file.ts
airplane exec outputs [...task parameters]
```

---

In a follow-up PR, we'll add support for reading environment variables from a `.env`.


colinking:
Y'all can defer to @yields for the PR review, but cc @francescaleung + @joshma for extra terminology feedback :)

yields:
> However, I'd love to find better command terminology than `dev` and `execute`. To me, `dev` implies "watch"-style behavior. For example, `next dev` vs. `next build`.

Perhaps, after the initial version of this land we can watch the files too?

pkg/cmd/tasks/dev/dev.go
```diff
@@ -0,0 +1,167 @@
...
+// slugFromScript attempts to extract a slug from a script.
+func slugFromScript(file string) (string, error) {
+	code, err := ioutil.ReadFile(file)
+	if err != nil {
+		return "", fmt.Errorf("cannot read file %s - %w", file, err)
+	}
+
+	slug, ok := runtime.Slug(code)
+	if !ok {
+		return "", fmt.Errorf("cannot find a slug in %s", file)
```
yields, line 163: Perhaps we can add a `runtime.MissingSlugError` or whatever that implements the `ExplainError()` method, right now when you attempt to deploy a task with no slug you'll get an explanation on how to link it.

pkg/cmd/tasks/execute/execute.go
```diff
@@ -105,34 +105,24 @@ func run(ctx context.Context, cfg config) error {
...
-	} else {
-		// Otherwise, try to prompt for parameters
-		if err := promptForParamValues(cfg.root.Client, task, req.ParamValues); err != nil {
-			return err
-		}
-	}
+	logger.Log("Executing %s task: %s", logger.Bold(task.Name), logger.Gray(client.TaskURL(task.Slug)))
 
-	logger.Log(logger.Gray("Running: %s", task.Name))
+	req.ParamValues, err = params.CLI(cfg.args, client, task)
```
yields, line 110: 😍

pkg/outputs/outputs.go
```diff
@@ -0,0 +1,40 @@
...
+		trimmed := strings.TrimPrefix(s, nonDefaultPrefix)
+		outputName := strings.SplitN(trimmed, " ", 2)[0]
+		if outputName != "" {
+			return outputName
+		}
+	}
+	return defaultOutputName
+}
+
+func ParseOutputValue(s string) interface{} {
```
yields, line 28: Not a huge deal, but there's a nice `json.RawMessage` that we can return, all it does is verify that the JSON is valid, in case we can't parse it all we have to do is interpret it like a string and construct a `json.RawMessage` with it.

Nothing scientific haha, but it should be way faster to parse the outputs.

pkg/runtime/javascript/javascript.go
```diff
@@ -95,3 +103,117 @@ func (r Runtime) FormatComment(s string) string {
 	}
 	return strings.Join(lines, "\n")
 }
+
+func (r Runtime) PrepareRun(ctx context.Context, path string, paramValues api.Values, opts api.KindOptions) ([]string, error) {
```
yields, line 107: wdyt of moving those arguments into a struct defined in the runtime package:

```go
type PrepareRunInput struct {
  Values api.Values
  Options api.KindOptions
  Path string
}
```

pkg/runtime/javascript/javascript.go
```diff
@@ -95,3 +103,117 @@ func (r Runtime) FormatComment(s string) string {
...
+
+	if err := os.WriteFile(filepath.Join(root, ".airplane/shim.ts"), []byte(shim), 0644); err != nil {
+		return nil, errors.Wrap(err, "writing shim file")
+	}
+
+	if err := os.RemoveAll(filepath.Join(root, ".airplane/dist")); err != nil {
+		return nil, errors.Wrap(err, "cleaning dist folder")
+	}
+
+	if utils.FilesExist(filepath.Join(root, "package.json")) != nil {
```
yields, line 136: `fs.Exist()`?

pkg/runtime/runtime.go
```diff
@@ -60,6 +61,8 @@ type Interface interface {
 	// FormatComment formats a string into a comment using
 	// the relevant comment characters for this runtime.
 	FormatComment(s string) string
+
+	PrepareRun(ctx context.Context, path string, paramValues api.Values, opts api.KindOptions) ([]string, error)
```
yields, line 65: Comment?

pkg/utils/file.go
```diff
@@ -0,0 +1,17 @@
+package utils
+
+import (
+	"fmt"
+	"os"
+	"path"
+)
+
+// FilesExist ensures that all paths exists or returns an error.
+func FilesExist(paths ...string) error {
```
yields, line 10: There's `fs.Exist()`, I haven't seen us using `utils.Exists(paths...)` but if we need it I suggest adding `fs.ExistsAny` or something.

pkg/utils/templates.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"strings"
+	"text/template"
+
+	"github.com/pkg/errors"
+)
+
+// ApplyTemplate executes template t with the provided data and
+// returns the output.
+func ApplyTemplate(t string, data interface{}) (string, error) {
```
yields, line 12: I haven't seen it used outside of the build package, let's move it there and privatize it?

pkg/cmd/tasks/dev/dev.go
```diff
@@ -0,0 +1,167 @@
...
+	if errors.Is(err, flag.ErrHelp) {
+		return nil
+	} else if err != nil {
+		return err
+	}
+
+	logger.Log("Locally running %s task %s", logger.Bold(task.Name), logger.Gray("("+cfg.root.Client.TaskURL(task.Slug)+")"))
+	logger.Log("")
+
+	cmds, err := r.PrepareRun(ctx, cfg.file, paramValues, task.KindOptions)
```
yields, line 95: Should `PrepareRun` return the temporary directory created so that it is cleaned up after the run?

yields (approved):
LGTM overall! so excited for this, just a few comments but feel free to merge and we can address them later :)

pkg/build/node.go
```diff
@@ -207,12 +219,12 @@ func nodeLegacyBuilder(root string, args Args) (string, error) {
...
 	entrypoint = path.Join(buildWorkdir, entrypoint)
 
-	baseImage, err := getBaseNodeImage(args["nodeVersion"])
+	baseImage, err := getBaseNodeImage(nodeVersion)
 	if err != nil {
 		return "", err
 	}
 
-	return templatize(`
+	return utils.ApplyTemplate(`
```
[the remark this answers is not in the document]
colinking: I had originally pulled this out since I was templating the shim outside of the `build` package, but ended up moving that templating back into `build` and didn't undo this change. Thanks!

pkg/outputs/outputs.go
colinking: `json.RawMessage` is just a `[]byte` under the hood, so we would end up persisting a `[]byte` that we would need to know to further unmarshal later on. Seems cleaner to fully unmarshal now

pkg/cmd/tasks/dev/dev.go
colinking: I imagine we'll use `.airplane` to cache data between runs down the road, so I want to leave it around!

PR #175, opened 2021-06-11
Specify full image URIs

yields:
Kaniko presumes the images are hosted on GCR and so it attempts to fetch
them from there and then will fallback to dockerhub.

By specifying the image URIs we prevent kaniko from assuming GCR.

### DIFF
--- a/pkg/build/versions.json
+++ b/pkg/build/versions.json
@@ -1,43 +1,43 @@
 {
   "node": {
     "16": {
-      "image": "node",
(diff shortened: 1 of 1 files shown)


joshma (approved):
Huh interesting - does this save much time? Or is it just a nice to have?

yields:
> Huh interesting - does this save much time? Or is it just a nice to have?

not at all unfortunately, but nice to have since it produces a warn log. perhaps one day we'll consume the logs and alert on them, dunno.

PR #177, opened 2021-06-12
Improve missing `package.json` error handling

colinking:
If you do not have a `package.json`, you get a relatively cryptic error:

```
Error: Runtime: resource is missing
```

We now return a nicer error:

```
Error: A package.json could not be found
```

The `resource is missing` error was surfaced from `runtime.Pathof`. I've tweaked this function to return a boolean instead, so it's easier for callers to check if the file was found or not. I've also moved this into the `fsx` package with the other file-related helpers. It's now called `fsx.Find`.


yields (approved):
lgtm!

PR #180, opened 2021-06-14
Handle >64KB log lines from `airplane dev`

colinking:
We'll apply this fix to our agents, too, so they can handle programs that generate >64KB log lines. Starting with `airplane dev`.

### DIFF
--- a/pkg/utils/bufiox/scanner.go
+++ b/pkg/utils/bufiox/scanner.go
@@ -1,0 +1,56 @@
+package bufiox
+
+import (
+	"bufio"
+	"io"
+)
+
+const bufferSizeBytes = 64 * 1024
+
+// Scanner is a drop-in replacement for bufio.Scanner that will
+// handle log lines that are longer than the default limit in
+// bufio (64KB).
+type Scanner struct {
+	r     *bufio.Reader
+	token []byte
+
+	err error
+}
+
+func NewScanner(r io.Reader) *Scanner {
+	return &Scanner{
+		r: bufio.NewReaderSize(r, bufferSizeBytes),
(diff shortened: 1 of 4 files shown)


pkg/utils/bufiox/scanner.go
```diff
@@ -0,0 +1,56 @@
+package bufiox
```
colinking, line 1: I'm considering placing util-like packages under the `utils/` folder so that the top-level `pkg` directory isn't littered with one-off packages like this. wdyt?

yields: yeah sgtm, could call it `ext/` too indicate that it's not one-off utilities but extensions to stdlib packages. For one-off utilities I'm for trying to keep them as private functions in packages that need them.

pkg/utils/bufiox/scanner.go
```diff
@@ -0,0 +1,56 @@
...
+		s.token = append(s.token, line...)
+
+		done = !isPrefix
+	}
+
+	return true
+}
+
+func (s *Scanner) Bytes() []byte {
+	return s.token
```
yields, line 47: Should we copy here? or is `.Bytes()` usage generally safe for packages that call it?

yields (approved):
lgtm! just one question/comment about copying the buffer.

pkg/utils/bufiox/scanner.go
colinking: This takes the same approach as `bufio.Scanner`. The caller's responsible for copying the byte slice if they want to access it after their next call to `.Scan()`

PR #181, opened 2021-06-15
Add Django support to init

yields:
The PR adds Django support to the init command, I would like to merge
this as soon as possible, even if deploy is not implemented yet it doesn't block
us from releasing the CLI because `--use` is hidden and not documented.

See the individual commit descriptions for more info, but the gist is to
add support for:

```bash
$ airplane init --use django
```

I don't like the `--use` flag and i would like to find a better name
so suggestions are welcome! :)

Unlike `init --slug <slug> <path>` the command is expected to be run from
a root of the project. It prompts the user to select which commands they
would like to include in the airplane.json file.

Once this is merged, all deploy will have to do is the following:

- [ ] Read the `airplane.json` file and list the commands
- [ ] For each command, figure out its `api.Parameters`
- [ ] Create/Update tasks accordingly (this will likely need bulk API endpoint)
- [ ] Build a single dockerfile.
- [ ] For every task adjust the entrypoint.

### DIFF
--- a/pkg/cmd/tasks/initcmd/project.go
+++ b/pkg/cmd/tasks/initcmd/project.go
@@ -1,0 +1,88 @@
+package initcmd
+
+import (
+	"context"
+	"path/filepath"
+
+	"github.com/AlecAivazis/survey/v2"
+	"github.com/airplanedev/cli/pkg/ap"
+	_ "github.com/airplanedev/cli/pkg/ap/django"
+	"github.com/airplanedev/cli/pkg/logger"
+	"github.com/pkg/errors"
+)
+
+// Project initializes a project.
+func project(ctx context.Context, name string) error {
+	root, err := filepath.Abs(".")
+	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/cmd/tasks/initcmd/project.go
```diff
@@ -0,0 +1,88 @@
...
+		return createProject(name, root)
+	}
+
+	if err != nil {
+		return err
+	}
+
+	logger.Log("Project file already exists with %d tasks.", len(project.Tasks))
+	logger.Log("You can deploy all tasks using:")
+	logger.Log("  airplane deploy airplane.json")
```
yields, line 33: this is awful 😅, perhaps we can go with `airplane deploy` and have it look for `airplane.json`?

yields:
decided to go with a different strategy for a few reasons:

1. Django commands might be internal, not ideal to ship them to be managed by Airplane.
2. This strategy requires us to parse/infer the arguments in each Django `Command`, might be too fragile.

PR #184, opened 2021-06-16
Install node shim dependencies only if missing

colinking:
This PR improves the behavior for installing shim dependencies (currently just `@types/node`) in `airplane dev` and in our Node builder. Both will now check if `@types/node` is installed in the task's `package.json` and skip over the extra `npm install` if it isn't needed. For `airplane dev`, that means we only have to run `npm install` at most once per task rather than once per `airplane dev` invocation. This is a significant speed-up for projects with large dep trees. For our Node builder, this saves a bit of time since we only run one `npm install` rather than two (`npm install` vs. `npm install && npm install --save-dev @types/node`).

### DIFF
--- a/pkg/runtime/javascript/javascript.go
+++ b/pkg/runtime/javascript/javascript.go
@@ -149,7 +149,7 @@ func (r Runtime) PrepareRun(ctx context.Context, opts runtime.PrepareRunOptions)
 		return nil, errors.New("a package.json is missing")
 	}
 
-	if !build.HasNodeDeps(root, "@types/node") {
+	if !build.HasNodeShimDeps(root) {
 		isYarn := fsx.AssertExistsAll(filepath.Join(root, "yarn.lock")) == nil
 		var cmd *exec.Cmd
 		if isYarn {
--- a/pkg/runtime/javascript/javascript.go
+++ b/pkg/runtime/javascript/javascript.go
@@ -18,6 +18,7 @@ import (
 	"github.com/airplanedev/cli/pkg/fsx"
 	"github.com/airplanedev/cli/pkg/logger"
 	"github.com/airplanedev/cli/pkg/runtime"
+	"github.com/airplanedev/cli/pkg/utils"
 	"github.com/blang/semver/v4"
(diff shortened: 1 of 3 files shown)


pkg/runtime/javascript/javascript.go
```diff
@@ -142,27 +142,30 @@ func (r Runtime) PrepareRun(ctx context.Context, opts runtime.PrepareRunOptions)
 		return nil, errors.Wrap(err, "cleaning dist folder")
 	}
 
-	if fsx.AssertExistsAll(filepath.Join(root, "package.json")) != nil {
-		if err := os.WriteFile(filepath.Join(root, "package.json"), []byte("{}"), 0777); err != nil {
-			return nil, errors.Wrap(err, "creating default package.json")
+	// Confirm we have a `package.json`, otherwise we might install shim dependencies
```
joshma, line 145: (PrepareRun is only called for dev right? just checking my understanding)

pkg/runtime/javascript/javascript.go
```diff
@@ -142,27 +142,30 @@ func (r Runtime) PrepareRun(ctx context.Context, opts runtime.PrepareRunOptions)
...
+	// in the wrong folder.
+	hasPkgJSON := fsx.AssertExistsAll(filepath.Join(root, "package.json")) == nil
+	if !hasPkgJSON {
+		return nil, errors.New("a package.json is missing")
+	}
+
+	if !build.HasNodeShimDeps(root) {
```
joshma, line 152: It would be nice if this prompted the user to install this, since it would change the state of the workspace - "Airplane needs @types/node to build your task - OK to run `...`?"

joshma (approved):

pkg/runtime/javascript/javascript.go
colinking: Yeah

pkg/runtime/javascript/javascript.go
colinking: Good call! Added:

image

PR #185, opened 2021-06-16
Add support for local Python execution

colinking:
This PR adds support for local execution of Python tasks.

### DIFF
--- a/pkg/build/node.go
+++ b/pkg/build/node.go
@@ -120,9 +120,9 @@ func NodeShim(entrypoint string) (string, error) {
 	entrypoint = strings.TrimSuffix(entrypoint, ".ts")
 
 	shim, err := applyTemplate(nodeShim, struct {
-		ImportPath string
+		Entrypoint string
 	}{
-		ImportPath: entrypoint,
+		Entrypoint: entrypoint,
 	})
 	if err != nil {
 		return "", errors.Wrap(err, "templating shim")
--- a/pkg/build/node.go
+++ b/pkg/build/node.go
@@ -25,8 +25,7 @@ func node(root string, options api.KindOptions) (string, error) {
 
 	// Assert that the entrypoint file exists:
 	entrypoint, _ := options["entrypoint"].(string)
-	entrypoint = filepath.Join(root, entrypoint)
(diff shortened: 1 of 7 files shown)


pkg/build/python.go
```diff
@@ -35,41 +29,35 @@ func python(root string, args api.KindOptions) (string, error) {
...
+    RUN mkdir -p .airplane && echo '{{.Shim}}' > .airplane/shim.py
     {{if not .HasInit}}
     RUN touch __init__.py
     {{end}}
     COPY . .
 		{{if .HasRequirements}}
     RUN pip install -r requirements.txt
 		{{end}}
-    ENTRYPOINT ["python", "/shim.py", "/airplane/{{ .Entrypoint }}"]
+    ENTRYPOINT ["python", ".airplane/shim.py"]
```
colinking, line 48: The shim doesn't need the path to the entrypoint (since it is templated in). So I adjusted the shim to expect 2 args instead of 3.

pkg/build/python.go
```diff
@@ -35,41 +29,35 @@ func python(root string, args api.KindOptions) (string, error) {
...
 	if err != nil {
-		return "", errors.Wrapf(err, "rendering shim")
+		return "", err
 	}
 
 	const dockerfile = `
     FROM {{ .Base }}
     WORKDIR /airplane
-    RUN echo '{{.Shim}}' > /shim.py
+    RUN mkdir -p .airplane && echo '{{.Shim}}' > .airplane/shim.py
```
colinking, line 40: Moved the shim to `.airplane` to mirror how we do it in local dev.

pkg/build/node.go
```diff
@@ -58,7 +57,7 @@ func node(root string, options api.KindOptions) (string, error) {
 		return "", err
 	}
 
-	shim, err := NodeShim(root, entrypoint)
+	shim, err := NodeShim(entrypoint)
```
colinking, line 60: Had trouble identifying if `entrypoint` was relative or absolute. Went through and cleaned up the entrypoint-related code in Node/Python so that `entrypoint` = relative and `absEntrypoint` = absolute.

yields (approved):
💥

PR #187, opened 2021-06-17
Use local `tsc` if installed

colinking:
This PR updates `airplane dev` to use the local version of `tsc` if installed, by accessing the TS CLI through `npx`. This also means we'll auto-install `tsc` globally (with a confirmation) if it isn't available.

This PR also adds extra debug logs that print the version of `node`, `npx`, and `tsc`.

cc @joshma


yields (approved):
lgtm

PR #188, opened 2021-06-17
Fix workdir missing omitempty directive

joshma:
This was breaking existing YAML files because they didn't contain
`workdir`:

```
Error: Error reading myTask/myTask.yml

Node: workdir is required
```


joshma:
This is very subtle.... @yields @colinking we basically have to remember to add `omitempty` on new Definition fields

colinking (approved):

PR #191, opened 2021-06-20
Try Coveralls

colinking:
Experimenting with Coveralls. I haven't yet enabled a paid account on coveralls, so using this repo since it's public.


colinking:
Coveralls doesn't leave comments until it has a first build on the branch that is being merged into. I'm going to merge this PR now, but can incorporate any PR feedback after-the-fact.

PR #194, opened 2021-06-21
Add retries to API Client

yields:
Adds simple retries to the API client, defaults to 5 attempts.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,6 +20,7 @@ require (
 	github.com/gorilla/mux v1.8.0 // indirect
 	github.com/gosimple/slug v1.9.0
 	github.com/joho/godotenv v1.3.0
+	github.com/jpillora/backoff v1.0.0
 	github.com/kr/pty v1.1.8 // indirect
 	github.com/kr/text v0.2.0
 	github.com/mattn/go-isatty v0.0.12
--- a/go.mod
+++ b/go.mod
@@ -19,6 +19,7 @@ require (
 	github.com/golang/snappy v0.0.3 // indirect
 	github.com/gorilla/mux v1.8.0 // indirect
 	github.com/gosimple/slug v1.9.0
+	github.com/hashicorp/go-retryablehttp v0.7.0 // indirect
 	github.com/joho/godotenv v1.3.0
 	github.com/jpillora/backoff v1.0.0
 	github.com/kr/pty v1.1.8 // indirect
--- a/go.mod
+++ b/go.mod
@@ -19,9 +19,8 @@ require (
 	github.com/golang/snappy v0.0.3 // indirect
 	github.com/gorilla/mux v1.8.0 // indirect
 	github.com/gosimple/slug v1.9.0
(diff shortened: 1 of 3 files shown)


pkg/api/client.go
```diff
@@ -271,23 +303,55 @@ func (c Client) GetBuildLogs(ctx context.Context, buildID string, since time.Tim
...
-	err = c.do(ctx, "GET", "/resources/list", nil, &res)
+	err = c.Do(ctx, "GET", "/resources/list", nil, &res)
 	return
 }
 
+// Do sends a request with `method`, `path`, `payload` and `reply`.
+//
+// The method will retry up to the configured max attempts with backoff.
+func (c Client) Do(ctx context.Context, method, path string, payload, reply interface{}) error {
+	var max = c.maxAttempts()
```
joshma, line 319:
```suggestion
	max := c.maxAttempts()
```

pkg/api/client.go
```diff
@@ -271,23 +303,55 @@ func (c Client) GetBuildLogs(ctx context.Context, buildID string, since time.Tim
...
+	var err error
+	var j int
+
+	var b = &backoff.Backoff{
```
joshma, line 323:
```suggestion
	b := &backoff.Backoff{
```

pkg/api/client.go
```diff
@@ -30,6 +30,33 @@ func (err Error) Error() string {
 	return fmt.Sprintf("api: %d - %s", err.Code, err.Message)
 }
 
+// Temporary returns true if the error is temporary.
+func (err Error) Temporary() bool {
+	return err.Code == 503
```
joshma, line 35: I think 5xx except 501 can be considered Temporary

joshma (changes requested):
We use https://github.com/hashicorp/go-retryablehttp/blob/master/client.go in airport: pkg/net/net.go - WDYT of just using that same library instead? It seems like it handles some more cases in baseRetryPolicy

go.mod
```diff
@@ -19,7 +19,9 @@ require (
 	github.com/golang/snappy v0.0.3 // indirect
 	github.com/gorilla/mux v1.8.0 // indirect
 	github.com/gosimple/slug v1.9.0
+	github.com/hashicorp/go-retryablehttp v0.7.0 // indirect
 	github.com/joho/godotenv v1.3.0
+	github.com/jpillora/backoff v1.0.0
```
joshma, line 24: (Just a reminder to run go mod tidy again and get rid of this dep)

pkg/api/client.go
```diff
@@ -358,3 +368,15 @@ func (c Client) token() (string, error) {
 	}
 	return c.Token, nil
 }
+
+// Sleep sleeps a maximum of ctx or d.
+func sleep(ctx context.Context, d time.Duration) error {
```
joshma, line 373: Can remove?

joshma (approved):

PR #195, opened 2021-06-21
Improve layer caching

yields:
When a docker layer changes docker will invalidate all subsequent layers. This means that if we do the following:

	ADD . /src
	RUN npm install /src/package.json

It means that every time a user changes their source code we will re-install the package.json, instead we should do the following:

	ADD package.json /src
	RUN npm install /src/package.json
	ADD . /src

### DIFF
--- a/pkg/build/node.go
+++ b/pkg/build/node.go
@@ -95,10 +95,10 @@ func node(root string, options api.KindOptions) (string, error) {
 
 		RUN npm install -g typescript@4.2
 
-		COPY . /airplane
-
 		{{if not .HasPackageJSON}}
 		RUN echo '{}' > /airplane/package.json
+		{{else}}
+		ADD package.json ./airplane
 		{{end}}
 
 		{{if .IsYarn}}
@@ -114,6 +114,8 @@ func node(root string, options api.KindOptions) (string, error) {
 		RUN npm install @types/node
 		{{end}}
 		{{end}}
+
+		ADD . /airplane
 
 		RUN mkdir -p /airplane/.airplane/dist && \
 			echo '{{.Shim}}' > /airplane/.airplane/shim.ts && \
--- a/pkg/build/node.go
+++ b/pkg/build/node.go
@@ -94,11 +94,10 @@ func node(root string, options api.KindOptions) (string, error) {
 		RUN [ -z "${BUILD_NPM_TOKEN}" ] || echo "//registry.npmjs.org/:_authToken=${BUILD_NPM_TOKEN}" > .npmrc
 
 		RUN npm install -g typescript@4.2
+		COPY . /airplane
 
 		{{if not .HasPackageJSON}}
 		RUN echo '{}' > /airplane/package.json
-		{{else}}
-		ADD package.json ./airplane
 		{{end}}
 
 		{{if .IsYarn}}
@@ -114,8 +113,6 @@ func node(root string, options api.KindOptions) (string, error) {
 		RUN npm install @types/node
 		{{end}}
 		{{end}}
-
-		ADD . /airplane
 
 		RUN mkdir -p /airplane/.airplane/dist && \
 			echo '{{.Shim}}' > /airplane/.airplane/shim.ts && \
--- a/pkg/build/python.go
+++ b/pkg/build/python.go
@@ -38,13 +38,13 @@ func python(root string, args api.KindOptions) (string, error) {
     WORKDIR /airplane
     RUN mkdir -p .airplane && echo '{{.Shim}}' > .airplane/shim.py
 		{{if .HasRequirements}}
-		ADD requirements.txt .
+		COPY requirements.txt .
     RUN pip install -r requirements.txt
 		{{end}}
     {{if not .HasInit}}
     RUN touch __init__.py
     {{end}}
-    ADD . .
+    COPY . .
     ENTRYPOINT ["python", ".airplane/shim.py"]
 	`
 
--- a/pkg/build/python.go
+++ b/pkg/build/python.go
@@ -37,13 +37,14 @@ func python(root string, args api.KindOptions) (string, error) {
     FROM {{ .Base }}
     WORKDIR /airplane
     RUN mkdir -p .airplane && echo '{{.Shim}}' > .airplane/shim.py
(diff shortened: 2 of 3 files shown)


yields:
Removed the node changes to keep this working with workspaces. For node we may detect workspaces and then use that layer caching improvement. I'll put it in another PR.

pkg/build/node.go
```diff
@@ -95,10 +95,10 @@ func node(root string, options api.KindOptions) (string, error) {
 
 		RUN npm install -g typescript@4.2
 
-		COPY . /airplane
-
 		{{if not .HasPackageJSON}}
 		RUN echo '{}' > /airplane/package.json
+		{{else}}
+		ADD package.json ./airplane
```
joshma, line 101: BTW - can we use COPY instead of ADD? ADD does too much; IMO we should use COPY by default.

pkg/build/node.go
```diff
@@ -115,6 +115,8 @@ func node(root string, options api.KindOptions) (string, error) {
 		{{end}}
 		{{end}}
 
+		ADD . /airplane
```
joshma, line 118: This should be moved back to earlier (L98), since `.` is needed for `npm install`!

pkg/build/python.go
```diff
@@ -37,13 +37,14 @@ func python(root string, args api.KindOptions) (string, error) {
     FROM {{ .Base }}
     WORKDIR /airplane
     RUN mkdir -p .airplane && echo '{{.Shim}}' > .airplane/shim.py
-    {{if not .HasInit}}
-    RUN touch __init__.py
-    {{end}}
-    COPY . .
 		{{if .HasRequirements}}
+		ADD requirements.txt .
```
joshma, line 41: might as well fix the indentation here / L40? this is getting weird 😂

pkg/build/python.go
```diff
@@ -37,13 +37,14 @@ func python(root string, args api.KindOptions) (string, error) {
...
     RUN pip install -r requirements.txt
 		{{end}}
+    {{if not .HasInit}}
+    RUN touch __init__.py
+    {{end}}
+    ADD . .
```
joshma, line 47: (same comment, ADD -> COPY)

pkg/build/versions.json
```diff
@@ -38,8 +38,8 @@
   "python": {
     "3": {
       "image": "registry.hub.docker.com/library/python",
-      "tag": "3.9.4-buster",
-      "digest": "sha256:b004a71e38f8ace26e7554d5c2fa802a8bb39a5818cbe10ab49fd0b408a40c20"
+      "tag": "3.9.4-alpine3.13",
```
joshma, line 41: will `alpine` work for node/python? my understanding is that there tends to be os dependencies in both languages, so it might not always be portable to e.g. alpine (which is why we went with debian at first)

pkg/build/python.go
```diff
@@ -37,13 +37,14 @@ func python(root string, args api.KindOptions) (string, error) {
...
-    {{if not .HasInit}}
-    RUN touch __init__.py
-    {{end}}
-    COPY . .
 		{{if .HasRequirements}}
+		ADD requirements.txt .
     RUN pip install -r requirements.txt
 		{{end}}
+    {{if not .HasInit}}
+    RUN touch __init__.py
```
joshma, line 45: didn't see this the first time around, but is this actually needed here? it defines the directory it's in as a package, but...the working directory is not usually a package

joshma (changes requested):

pkg/build/node.go
yields: what's with the `!` 😂? is this because there's maybe a `lock` file for yarn or npm?

pkg/build/python.go
yields: there's an error while importing from the shim if we don't add it :/

pkg/build/node.go
joshma: Oh, the exclamation just because this is a tricky thing that I wanted to emphasize! :)

This is what I was mentioning earlier, regarding NPM workspaces. You cannot always assume that `package.json+package-lock.json+yarn.lock` are the only files necessary for `npm install / yarn install`. If you're installing in `/airplane/server` you might have code in `/airplane` and `/airplane/shared` that is needed for install, and so you have to make sure you add that earlier.

pkg/build/python.go
joshma: hm, can you repro / share the error? I'm like 70% sure that either this won't error or, if it does error, there's a more proper solution 😇

yields: you're right, it seems that it didn't work in the previous method we used to import modules in the shim. Removed `__init__.py`.

pkg/build/versions.json
yields: It depends what do you mean by work? of-course it will, Ibinaries/sharedlibs that are included in debian won't work here tho. but it's totally worth it imho, works for the simple case, other cases could use the `.yml` files or ship their own images.

joshma: Well alpine uses musl libc instead of glibc (https://hub.docker.com/_/node has a warning about this), and so I'm concerned that basic packages might not even install properly. The default node image uses debian. So by "work" I mean "work in ~90% of use cases". 

I gave it a quick try and it seems like numpy won't work on alpine, but e.g. openssl / node-gyp works on alpine.So it's not super clear what breaks.

> other cases could use the .yml files or ship their own images

By "use .yml files" do you mean use the dockerfile builder instead? I suppose we could eventually allow configuring the base OS if necessary, for `node`?

joshma (approved):

pkg/build/versions.json
yields: yeah that too, we can come up with a lot of ways to do it. but I think it's worth it and will work. I've been using alpine for Go consistently and musl libc works well.

joshma: Yeah ok we can see where this goes. :)

I get it that Go works, my main warning is that it's a lot hairier outside of Go. Node and Python packages have a history of depending on C libraries (e.g. node-gyp https://github.com/mhart/alpine-node/issues/27 https://github.com/nodejs/node-gyp/issues/1855).

PR #196, opened 2021-06-24
Add (opt-in) sentry and segment reporting

joshma:
This uses SEGMENT_WRITE_KEY and SENTRY_DSN to configure segment and sentry reporting.

Segment will send tracking on explicitly instrumented calls (`analytics.Track`), and Sentry will
report on all CLI errors at the root.

Note this is opt-in - we track `enableTelemetry` in the config file and the user is prompted
the first time they run this without that field set.


cmd/airplane/main.go
```diff
@@ -42,6 +41,12 @@ func main() {
 		}
 		logger.Log("")
 
+		// TODO: is this too noisy?
+		sentryID := sentry.CaptureException(err)
```
yields, line 45: Doesn't sentry allow us to filter/ignore errors from their UI? if so, we can remove the TODO?

yields (commented):
lgtm overall

cmd/airplane/main.go
joshma: Yeah that's true! Let's try that first. In my experience the "ignore" parts of Sentry have been more as a last resort, since you still have to ignore them as they come up and they still count towards your quota. But we can give it a shot first.

joshma:
@colinking we should probably add analytics on `dev` usage, esp. since we don't see it on our servers. didn't get to that in this PR.

pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -56,12 +62,38 @@ func New(c *cli.Config) *cobra.Command {
...
+	analytics.Track(cfg.root, "CLI deploy", map[string]interface{}{
+		"from":             props.from,
+		"kind":             props.kind,
+		"task_id":          props.taskID,
+		"task_slug":        props.taskSlug,
+		"task_name":        props.taskName,
+		"build_id":         props.buildID,
+		"errored":          err != nil,
+		"duration_seconds": time.Since(start).Seconds(),
+	})
```
colinking, line 97: Nit, but I'd opt to duplicate the track call rather than introducing track properties as part of the function contract here.

pkg/cmd/auth/login/login.go
```diff
@@ -33,6 +34,7 @@ func run(ctx context.Context, c *cli.Config) error {
 	}
 
 	logger.Log("You're all set!\n\nTo see what tasks you can run, try:\n    airplane tasks list")
+	analytics.Track(c, "CLI login", nil)
```
colinking, line 37: nit: `User Logged In`

pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -56,12 +62,38 @@ func New(c *cli.Config) *cobra.Command {
...
+	var err error
 	if ext == ".yml" || ext == ".yaml" {
-		return deployFromYaml(ctx, cfg)
+		props, err = deployFromYaml(ctx, cfg)
+	} else {
+		props, err = deployFromScript(ctx, cfg)
 	}
 
-	return deployFromScript(ctx, cfg)
+	analytics.Track(cfg.root, "CLI deploy", map[string]interface{}{
```
colinking, line 88: nit: `Task Deployed`

pkg/analytics/analytics.go
```diff
@@ -0,0 +1,107 @@
...
+		// Now try again.
+		return Init()
+	}
+	if c.EnableTelemetry != "yes" {
+		return nil
+	}
+	segmentClient = analytics.New(segmentWriteKey)
+	return sentry.Init(sentry.ClientOptions{
+		Dsn:   sentryDSN,
+		Debug: true,
```
colinking, line 40: What does `Debug: true` do here? Should we make this conditional on `--debug`?

pkg/analytics/analytics.go
```diff
@@ -0,0 +1,107 @@
...
+	})
+}
+
+func telemetryOptIn(c conf.Config) error {
+	var allow bool
+	logger.Log("Welcome to the Airplane CLI!")
+	logger.Log("")
```
colinking, line 47: nit but eh, I'd remove this :)

pkg/analytics/analytics.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+func Init() error {
+	c, err := conf.ReadDefault()
+	if err != nil {
+		return err
+	}
+	if c.EnableTelemetry == "" {
+		// User has not specified one way or the other, ask them to opt-in.
+		if err := telemetryOptIn(c); err != nil {
```
colinking, line 28: Could you also add an `airplane analytics` command so that users who initially opt-out can opt-in after-the-fact to help us debug their issues?

Or, what if we offer a `--send-telemetry` flag that will forward analytics for a singular CLI invocation? If set, this could check and if the user hasn't explicitly opted-in, it could ask if they want to temporarily authorize telemtry for that single invocation. We could then simply ask users to re-run a command with the `--send-telemetry` flag.

pkg/analytics/analytics.go
```diff
@@ -0,0 +1,107 @@
...
+func telemetryOptIn(c conf.Config) error {
+	var allow bool
+	logger.Log("Welcome to the Airplane CLI!")
+	logger.Log("")
+	logger.Log("Is it OK for Airplane to collect usage analytics and error reports? This data will solely be used to improve the service.")
+	logger.Log("")
+	prompt := &survey.Confirm{
+		Message: "Opt in",
+		Default: true,
+	}
```
colinking, line 53: btw you can use `utils.Confirm` for this: https://github.com/airplanedev/cli/blob/073c9c7fe9089e30098899d482503faebc3f8704/pkg/cmd/auth/login/login.go#L48-L56

pkg/analytics/analytics.go
```diff
@@ -0,0 +1,107 @@
...
+	return nil
+}
+
+func Close() {
+	if segmentClient != nil {
+		if err := segmentClient.Close(); err != nil {
+			logger.Debug("error closing segment client: %v", err)
+		}
+	}
+	sentry.Flush(1 * time.Second)
```
colinking, line 78: Does Sentry do buffering by default? Would be nice if we can disable that, so events are sent immediately.

.goreleaser.yml
```diff
@@ -10,6 +10,8 @@ builds:
       - CGO_ENABLED=0
     ldflags:
       - -s -w -X github.com/airplanedev/cli/pkg/version.version={{.Version}} -X github.com/airplanedev/cli/pkg/version.date={{.Date}}
+      - -X github.com/airplanedev/cli/pkg/analytics/analytics.segmentWriteKey={{.Env.SEGMENT_WRITE_KEY}}
+      - -X github.com/airplanedev/cli/pkg/analytics/analytics.sentryDSN={{.Env.SENTRY_DSN}}
```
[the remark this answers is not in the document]
colinking: And you can test this with the `goreleaser` CLI: https://github.com/airplanedev/cli/blob/main/.github/CONTRIBUTING.md

This would produce a binary that you can then run locally. You could use that to confirm the events end up in Segment.

pkg/cli/config.go
```diff
@@ -23,6 +25,33 @@ type Config struct {
...
+	claims, ok := t.Claims.(jwt.MapClaims)
+	if !ok {
+		return res
+	}
+	res.UserID = claims["userID"].(string)
+	res.TeamID = claims["teamID"].(string)
+	return res
+}
+
+type TokenParse struct {
```
colinking, line 50: nit: but `TokenParse` sounds like the function and `TokenInfo` sounds like the struct to me 😅

pkg/cli/config.go
```diff
@@ -23,6 +25,33 @@ type Config struct {
 	Version bool
 }
 
+// TokenInfo parses UNVERIFIED JWT information - this information can be spoofed.
+// Should only be used for low-sensitivity use cases like analytics.
+func (c Config) TokenInfo() TokenParse {
```
colinking, line 30: I'd do something like React here (and similar to what the JWT library does below) and make this obvious in the func name, f.e. `ParseTokenForAnalytics`.

Goal being to make it very obvious to writers + PR reviewers to notice if we accidentally use this where we shouldn't :)

colinking (approved):
LGTM! +1 on adding more analytic events over time

pkg/analytics/analytics.go
```diff
@@ -0,0 +1,107 @@
...
+type TrackOpts struct {
+	UserID string
+	TeamID string
+	// Specify SkipSlack to avoid sending this event to Slack
+	SkipSlack bool
+}
+
+// Track sends a track event to Segment.
+// event should match "[event] by [user]" - e.g. "[Invite Sent] by [Alice]"
+func Track(c *cli.Config, event string, properties map[string]interface{}) {
```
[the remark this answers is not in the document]
joshma: I don't this would do what you want it to do? We could add in `traits` manually like we did for `api` but we don't have that user info here.

We should have already `Identify()`'d the user when they signed into the API.

So I don't know if we need to add anything here, does that sound right?

pkg/analytics/analytics.go
joshma: It does do buffering by default, I wasn't sure if we wanted to keep it buffered in case e.g. there were errors being sent in a for loop. I can disable buffering for now?

joshma: The segment client also buffers yeah?

pkg/cli/config.go
joshma: Nice, I like that.

pkg/cmd/tasks/execute/execute.go
```diff
@@ -140,11 +141,16 @@ func run(ctx context.Context, cfg config) error {
 
 	print.Outputs(state.Outputs)
 
+	analytics.Track(cfg.root, "CLI execute", map[string]interface{}{
```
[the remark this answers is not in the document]
joshma: Yeah this is fair, I knew about this but just went w/ whatever @swag had in https://docs.google.com/spreadsheets/d/1pbyYX-f0SiLSv7mzfz_xbwpucSSafZwjxj-wk6xcpJA/edit 😂 

I'll do a pass through on names. What do you do for actions like "Logged In"? It's OK to deviate from Object Action then?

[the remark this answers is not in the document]
joshma: "CLI Logged In"?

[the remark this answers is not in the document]
swag: my bad! yeah i agree with that spec advice generally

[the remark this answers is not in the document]
colinking: `User Logged In` works for that

pkg/analytics/analytics.go
colinking: By default, yeah

pkg/cmd/tasks/deploy/deploy.go
joshma: Yeahh that was my initial approach. Do you have any ideas on how to structure this to handle partial failure? The way I have it now, it returns a half-prepared `props` struct with e.g. task_id if it made it that far, so even if `err != nil` there's an event fired.

I guess I could `defer` a func to track maybe? And use a named return value for `err`?

.goreleaser.yml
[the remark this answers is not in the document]
joshma: > Can we also stamp in an environment field so we can distinguish local vs. prod errors?

How about adding the `version.Get()` as release instead? local errors will be `<unknown>`? We should probably add in the API host we're using as well.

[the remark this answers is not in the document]
joshma: Eugh adding host is a bit too complicated, might punt for now.

pkg/cmd/tasks/deploy/deploy.go
colinking: Yeah that sounds reasonable

pkg/analytics/analytics.go
joshma: Eh actually - @colinking why not let it buffer? segment delays it by 5s, sentry i think sends immediately but then batches the next set that has built up while sending - it seems like these libraries really don't like it if we buffer

joshma:
CleanShot 2021-06-29 at 23 33 46@2x

^ Added a `--with-telemetry` flag to override any opt-outs

joshma:
Hmm @colinking can you take another review of this? I'm not happy with this one: https://github.com/airplanedev/cli/pull/196/files/d92e6535a45c9adf2659e93a01fbb6d81d4a522d..712ea40a8b8be0a70bb965e8c11d9f901b232b50

pkg/analytics/analytics.go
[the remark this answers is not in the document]
colinking: True, we would always send an identify call for every user from the site

pkg/analytics/analytics.go
```diff
@@ -1,68 +1,70 @@
...
-		return Init()
+		return Init(cfg)
 	}
-	if c.EnableTelemetry != "yes" {
+	if !*c.EnableTelemetry && !cfg.WithTelemetry {
 		return nil
 	}
-	segmentClient = analytics.New(segmentWriteKey)
+	if cfg.WithTelemetry && !*c.EnableTelemetry {
+		logger.Warning("Enabling usage analytics and error reports because --with-telemetry was set.")
```
colinking, line 39: Maybe emphasize this is temporary, not permanent:

```suggestion
		logger.Warning("Temporarily enabling usage analytics and error reports because --with-telemetry was set.")
```

pkg/analytics/analytics.go
```diff
@@ -1,68 +1,70 @@
...
+	}
+	segmentClient, err = analytics.NewWithConfig(segmentWriteKey, analytics.Config{
+		DefaultContext: &analytics.Context{
+			App: analytics.AppInfo{
+				Name:    "cli",
+				Version: version.Get(),
```
colinking, line 45: :awyis:

pkg/cmd/tasks/deploy/script.go
```diff
@@ -16,65 +18,83 @@ import (
...
+		analytics.Track(cfg.root, "Task Deployed", map[string]interface{}{
+			"from":             props.from,
+			"kind":             props.kind,
+			"task_id":          props.taskID,
+			"task_slug":        props.taskSlug,
+			"task_name":        props.taskName,
+			"build_id":         props.buildID,
+			"errored":          err != nil,
+			"duration_seconds": time.Since(start).Seconds(),
+		})
```
colinking, line 38: nice

pkg/cmd/tasks/deploy/script.go
```diff
@@ -16,65 +18,83 @@ import (
...
 	if task.Kind != r.Kind() {
-		return props, fmt.Errorf("'%s' is a %s task. Expected a %s task.", task.Name, task.Kind, r.Kind())
+		err = errors.Errorf("'%s' is a %s task. Expected a %s task.", task.Name, task.Kind, r.Kind())
+		return
 	}
 
 	def, err := definitions.NewDefinitionFromTask(task)
 	if err != nil {
-		return props, err
+		return
```
colinking, line 80: I'd still recommend doing `return err` because it's easy to shadow the returned `err`, f.e.:

```go
if err := thing(); err != nil {
  // the returned err will be nil
  return
}
```

pkg/cmd/tasks/deploy/yaml.go
```diff
@@ -12,42 +14,58 @@ import (
...
-		return props, err
+	err = ensureConfigsExist(ctx, client, def)
+	if err != nil {
+		return
 	}
 
 	kind, kindOptions, err := def.GetKindAndOptions()
 	if err != nil {
-		return props, err
+		return
```
colinking, line 60: ^ actually here's an example -- I don't believe this `err` will be returned

colinking: Maybe we should use the `rX` convention from: https://airplanedev.slack.com/archives/C01DYGWC3P0/p1617906736003300

```go
func deployFromYaml(ctx context.Context, cfg config) (rErr error) {
  // ...
}
```

colinking (approved):

PR #198, opened 2021-06-25
Use `airplane dev` instead of `airplane exec`

colinking:
Replaces references to locally executing with `airplane dev` commands.

### DIFF
--- a/pkg/cmd/tasks/deploy/script.go
+++ b/pkg/cmd/tasks/deploy/script.go
@@ -116,7 +116,7 @@ func deployFromScript(ctx context.Context, cfg config) error {
 		return err
 	}
 
-	cmd := fmt.Sprintf("airplane execute %s", cfg.file)
+	cmd := fmt.Sprintf("airplane dev %s", cfg.file)
 	if len(def.Parameters) > 0 {
 		cmd += " -- [parameters]"
 	}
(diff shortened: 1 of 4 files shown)


pkg/cmd/tasks/deploy/script.go
```diff
@@ -116,7 +116,7 @@ func deployFromScript(ctx context.Context, cfg config) error {
 		return err
 	}
 
-	cmd := fmt.Sprintf("airplane execute %s", cfg.file)
+	cmd := fmt.Sprintf("airplane dev %s", cfg.file)
```
joshma, line 119: Shouldn't this still be `exec`? This is to run the task "for real" right? (Or did you want to start discouraging that?)

pkg/cmd/tasks/deploy/yaml.go
```diff
@@ -132,7 +132,7 @@ func deployFromYaml(ctx context.Context, cfg config) error {
 		return errors.Wrapf(err, "updating task %s", def.Slug)
 	}
 
-	cmd := fmt.Sprintf("airplane execute %s", def.Slug)
+	cmd := fmt.Sprintf("airplane dev %s", def.Slug)
```
joshma, line 135: Same comment as above

joshma (approved):

pkg/cmd/tasks/deploy/script.go
colinking: Hmm, I guess this is the output from the `deploy` command, I suppose we should leave as-is.

PR #201, opened 2021-06-28
Fix issue with single quotes in Node shim

colinking:
This PR fixes a bug introduced in https://github.com/airplanedev/cli/pull/199, which has not yet been released.

That PR added comments to the shim file that contained single quotes, which broke the Dockerfile we generated for Node tests. That PR had just been tested with `airplane dev`, which does not perform a Docker build.


joshma (approved):
This is why it's nicer iMO to `npm install` a package! Echoing shims is tricky business.

PR #204, opened 2021-06-29
Ignore `.airplane` directory during local build

colinking:
While debugging https://github.com/airplanedev/cli/pull/203, I noticed that the `.airplane` folder created by `airplane dev` was getting copied into the local Docker build context (it's ignored during remote builds).

This PR updates our local Docker builder to apply the same file ignores as remote builds.

### DIFF
--- a/pkg/build/builder.go
+++ b/pkg/build/builder.go
@@ -126,13 +126,9 @@ func (b *Builder) Build(ctx context.Context, taskID, version string) (*Response,
 	var name = "task-" + sanitizeTaskID(taskID)
 	var uri = repo + "/" + name + ":" + version
 
-	patterns, err := ignore.Patterns(b.root)
+	patterns, err := ignore.DockerignorePatterns(b.root)
 	if err != nil {
 		return nil, err
-	}
-	for i := range patterns {
-		// Convert .gitignore-style format -> .dockerignore-style format.
-		patterns[i] = "**/" + patterns[i]
 	}
 	tree, err := NewTree(TreeOptions{
 		ExcludePatterns: patterns,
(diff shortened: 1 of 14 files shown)


pkg/build/builder.go
```diff
@@ -125,7 +126,17 @@ func (b *Builder) Build(ctx context.Context, taskID, version string) (*Response,
 	var name = "task-" + sanitizeTaskID(taskID)
 	var uri = repo + "/" + name + ":" + version
 
-	tree, err := NewTree()
+	patterns, err := ignore.Patterns(b.root)
+	if err != nil {
+		return nil, err
+	}
+	for i := range patterns {
+		// Convert .gitignore-style format -> .dockerignore-style format.
```
joshma, line 134: hmmmmm lol this isn't foolproof, there's other ways that they differ - https://github.com/LinusU/gitignore-to-dockerignore/blob/master/index.js seems to take a stab at this, maybe adopt what they have there?

pkg/build/ignore/ignore.go
```diff
@@ -13,8 +13,8 @@ import (
 
 // Returns an IgnoreFunc that can be used with airplanedev/archiver to filter
 // out files that match a default list or user-provided .airplaneignore.
-func GetIgnoreFunc(taskRootPath string) (func(filePath string, info os.FileInfo) (bool, error), error) {
-	excludes, err := getIgnorePatterns(taskRootPath)
+func Func(taskRootPath string) (func(filePath string, info os.FileInfo) (bool, error), error) {
```
joshma, line 16: 😍

joshma (approved):

PR #205, opened 2021-06-29
Fix Python entrypoint and import path

joshma:
Fixes entrypoint from e.g. /task/my_command.py to task/my_command.py -
it shouldn't be an absolute path, this was breaking the shim import.

Fixes shim to append the taskroot to `sys.path` - otherwise,
task/my_command.py would not be able to import mod:

```
. is task root
task/
  my_command.py
mod/
  __init__.py
```

Also, switches python-shim.py to four-space indents instead of tabs.

### DIFF
--- a/pkg/build/python-shim.py
+++ b/pkg/build/python-shim.py
@@ -5,7 +5,7 @@ import json
 import sys
 
 def run(args):
-    sys.path.append("{{.Taskroot}}")
+    sys.path.append("{{.TaskRoot}}")
     
     if len(args) != 2:
         raise Exception("usage: python ./shim.py <args>")
--- a/pkg/build/python-shim.py
+++ b/pkg/build/python-shim.py
@@ -3,20 +3,21 @@
 import importlib.util as util
 import json
 import sys
-import os
 
 def run(args):
-	if len(args) != 2:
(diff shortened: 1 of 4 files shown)


pkg/build/python-shim.py
```diff
@@ -3,20 +3,21 @@
 import importlib.util as util
 import json
 import sys
-import os
 
 def run(args):
-	if len(args) != 2:
-		raise Exception("usage: python ./shim.py <args>")
+    sys.path.append("{{.Taskroot}}")
```
colinking, line 8:
```suggestion
    sys.path.append("{{.TaskRoot}}")
```

pkg/build/python.go
```diff
@@ -65,10 +65,12 @@ func python(root string, args api.KindOptions) (string, error) {
 var pythonShim string
 
 // PythonShim generates a shim file for running Python tasks.
-func PythonShim(entrypoint string) (string, error) {
+func PythonShim(taskroot, entrypoint string) (string, error) {
 	shim, err := applyTemplate(pythonShim, struct {
+		Taskroot   string
```
colinking, line 70:
```suggestion
		TaskRoot   string
```

colinking (approved):
:awyis:

PR #208, opened 2021-07-06
Fix indentation of dockerfile templates

joshma:
Just to reduce confusion and improve consistency, this dedents our
dockerfile templates so there's no question of how leading space is
interpreted.

Before this, we had a mix of spaces/tabs, and Docker docs discourage
leading whitespace: https://docs.docker.com/engine/reference/builder/

> For backward compatibility, leading whitespace before comments (#) and
> instructions (such as RUN) are ignored, but discouraged.

### DIFF
--- a/pkg/build/node.go
+++ b/pkg/build/node.go
@@ -82,42 +82,42 @@ func node(root string, options api.KindOptions) (string, error) {
 	// Down the road, we may want to give customers more control over this build process
 	// in which case we could introduce an extra step for performing build commands.
 	return applyTemplate(`
-		FROM {{.Base}}
+FROM {{.Base}}
 
-		WORKDIR /airplane{{.Workdir}}
+WORKDIR /airplane{{.Workdir}}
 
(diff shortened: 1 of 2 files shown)


pkg/build/node.go
```diff
@@ -82,42 +82,42 @@ func node(root string, options api.KindOptions) (string, error) {
...
-		RUN npm install @types/node
-		{{end}}
-		{{end}}
-
-		RUN mkdir -p /airplane/.airplane/dist && \
-			{{.InlineShim}} > /airplane/.airplane/shim.ts && \
-			tsc {{.TscArgs}}
-		ENTRYPOINT ["node", "/airplane/.airplane/dist/.airplane/shim.js"]
-	`, cfg)
+FROM {{.Base}}
```
colinking, line 85: Could you use `heredoc.Doc` instead? https://pkg.go.dev/github.com/makenowjust/heredoc

joshma: Oops didn't realize we already used it, will do.

PR #211, opened 2021-07-14
Simplify `airplane exec` example when a task has deployed

joshma:
It's easier for the user to just `airplane exec [slug]` instead of
trying to figure out the `--flags`.


joshma:
@colinking lmk what you think!

colinking (approved):

PR #212, opened 2021-07-21
Update install script to from /usr/local to $HOME/.airplane

joshma:
It's not safe to assume /usr/local is writeable - folks coming from
non-homebrew setups (vanilla setups) especially have trouble.

For now, we'll write to $HOME/.airplane and (still) have the
instructinos asking user to add it to $PATH.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -19,7 +19,7 @@ else
     download_uri="https://github.com/airplanedev/cli/releases/download/${1}/airplane_${target}.tar.gz"
 fi
 
-airplane_install="${AIRPLANE_INSTALL:-/usr/local}"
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -19,7 +19,7 @@ else
     download_uri="https://github.com/airplanedev/cli/releases/download/${1}/airplane_${target}.tar.gz"
 fi
 
-airplane_install="${AIRPLANE_INSTALL:-/usr/local}"
+airplane_install="${AIRPLANE_INSTALL:-$HOME/.airplane}"
```
colinking, line 22: Seems good for now, but we should eventually considering following: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

colinking (approved):

PR #213, opened 2021-07-23
Handle quoted output names correctly

francescaleung:
This matches the regexp dance we do in the API server.


joshma (approved):
Hate to encourage duplication, but can you add the tests over as well?

PR #215, opened 2021-07-27
Fix reverted permissions when deploying from CLI

colinking:
Previously, if you had explicit permissions set on a task, then deploying that task from the CLI would revert those permissions.

Resolves AIR-1556


pkg/api/types.go
```diff
@@ -29,24 +29,36 @@ type CreateTaskRequest struct {
...
+	Arguments                  []string          `json:"arguments" yaml:"arguments"`
+	Parameters                 Parameters        `json:"parameters" yaml:"parameters"`
+	Constraints                RunConstraints    `json:"constraints" yaml:"constraints"`
+	Env                        TaskEnv           `json:"env" yaml:"env"`
+	ResourceRequests           map[string]string `json:"resourceRequests" yaml:"resourceRequests"`
+	Resources                  map[string]string `json:"resources" yaml:"resources"`
+	Kind                       TaskKind          `json:"kind" yaml:"builder"`
+	KindOptions                KindOptions       `json:"kindOptions" yaml:"builderConfig"`
+	Repo                       string            `json:"repo" yaml:"repo"`
+	RequireExplicitPermissions bool              `json:"requireExplicitPermissions"`
```
joshma, line 46: Missing a `yaml` tag?

pkg/print/types.go
```diff
@@ -6,23 +6,25 @@ import (
...
+	Parameters                 api.Parameters       `json:"parameters" yaml:"parameters"`
+	Constraints                api.RunConstraints   `json:"constraints" yaml:"constraints"`
+	Env                        api.TaskEnv          `json:"env" yaml:"env"`
+	ResourceRequests           api.ResourceRequests `json:"resourceRequests" yaml:"resourceRequests"`
+	Resources                  api.Resources        `json:"resources" yaml:"resources"`
+	Kind                       api.TaskKind         `json:"builder" yaml:"builder"`
+	KindOptions                api.KindOptions      `json:"builderConfig" yaml:"builderConfig"`
+	Repo                       string               `json:"repo" yaml:"repo"`
+	RequireExplicitPermissions bool                 `json:"requireExplicitPermissions" yaml:"-"`
+	Permissions                api.Permissions      `json:"permissions" yaml:"-"`
```
joshma, line 26: If this is just for printing...can we omit it from this struct?

pkg/build/remote.go
```diff
@@ -97,19 +97,21 @@ func updateKindAndOptions(ctx context.Context, client *api.Client, def definitio
...
+		Command:                    task.Command,
+		Arguments:                  task.Arguments,
+		Parameters:                 task.Parameters,
+		Constraints:                task.Constraints,
+		Env:                        task.Env,
+		ResourceRequests:           task.ResourceRequests,
+		Resources:                  task.Resources,
+		Repo:                       task.Repo,
+		RequireExplicitPermissions: task.RequireExplicitPermissions,
+		Permissions:                task.Permissions,
```
joshma, line 113: Unfortunate that it has to be specified - can we not just remove this from the apiext endpoint? Is this actually used (by new or old)?

joshma (approved):

pkg/print/types.go
colinking: We do a type conversion, so it's needed to compile. But decided not to print this metadata for now

pkg/build/remote.go
colinking: Agreed it's a pain that you have to specify fields that haven't changed. Age old problem with CRUD APIs 😄 

And it seems reasonable to me that the external API can control permissions, so I'd rather not remove it from the API.

pkg/print/types.go
joshma: I think you need to do `json:"-"` too, for `-o json`?

pkg/build/remote.go
joshma: Okay, SG

PR #217, opened 2021-07-29
Add CodeQL scanning

joshma:
This file was autogenerated by GH - was thinking we could add to repo just to see what happens.


colinking (approved):
Did this just count the lines of code? 🤔 That's all it looks like it did from the metrics in the GHA run.

joshma:
It only runs on main lol, so we'll see what it does!

joshma:
Wait nevermind, it doesn't? Taking another look.

joshma:
screenshot_20210729_a8tZU8Do

^ looks like the metrics is just some outputs it can produce, but there's a lot of queries that don't have metrics

PR #218, opened 2021-07-29
Refactor analytics.ReportError and ignore terminal.InterruptErr

joshma:
terminal.InterruptErr is returned from survey when ctrl-c'd during a
prompt. Ignore those errors.


colinking (approved):
🙏🏻

PR #220, opened 2021-08-04
Add prototype of shell support

joshma:
This doesn't have everything yet (primarily missing params), but this allows you to
deploy a `script.sh` file and, optionally, looks for the nearest `Dockerfile` to build
the surrounding environment. (If not found, it uses a hardcoded `ubuntu` image.)


pkg/build/builder.go
```diff
@@ -148,7 +148,8 @@ func (b *Builder) Build(ctx context.Context, taskID, version string) (*Response,
 	}
 	logger.Debug(strings.TrimSpace(dockerfile))
 
-	if err := tree.Write("Dockerfile", strings.NewReader(dockerfile)); err != nil {
+	dockerfilePath := "Dockerfile.airplane"
```
joshma, line 151: Without this, we were attempting to override an existing `Dockerfile` (i.e. the one being used for the script!) and it wasn't actually successfully writing.

pkg/build/shell-shim.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+# TODO: set up env vars - for each argument, split by first =, set AP_FIRST_PART=SECOND_PART
```
joshma, line 3: Doing this in a separate PR

pkg/build/shell.go
```diff
@@ -0,0 +1,110 @@
...
+
+	// Extend template with our own logic - set up a WORKDIR and shim.
+	dockerfileTemplate = dockerfileTemplate + heredoc.Doc(`
+		WORKDIR /airplane
+		RUN mkdir -p .airplane && {{.InlineShim}} > .airplane/shim.sh
+		
+		COPY . .
+		RUN chmod +x {{.Entrypoint}}
+		
+		ENTRYPOINT ["bash", ".airplane/shim.sh"]
```
joshma, line 83: All this does right now is `exec` into the entrypoint, but in the future it'll set up env vars from params and possibly other things.

pkg/build/shell.go
```diff
@@ -0,0 +1,102 @@
...
+
+	"github.com/MakeNowJust/heredoc"
+	"github.com/airplanedev/cli/pkg/api"
+	"github.com/airplanedev/cli/pkg/fsx"
+	"github.com/pkg/errors"
+)
+
+func shell(root string, options api.KindOptions) (string, error) {
+	// Assert that the entrypoint file exists:
+	entrypoint, _ := options["entrypoint"].(string)
```
francescaleung, line 16: Not a bad idea to check the second return value here

pkg/runtime/shell/shell.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/airplanedev/cli/pkg/api"
+	"github.com/airplanedev/cli/pkg/fsx"
+	"github.com/airplanedev/cli/pkg/logger"
+	"github.com/airplanedev/cli/pkg/runtime"
+	"github.com/pkg/errors"
+)
+
+// Init register the runtime.
+func init() {
+	runtime.Register(".sh", Runtime{})
```
francescaleung, line 20: This means you can't deploy a script without a file extension, right?

francescaleung (approved):

pkg/build/shell.go
joshma: Can you explain? This'll be empty string if it's !ok, so this handles both cases in one.

pkg/runtime/shell/shell.go
joshma: Yeah, correct - for now I'm thinking we start with this, and require .sh. I was thinking to generalize we'd bring back `.yml` support and you'd deploy the `yml` file with an explicit `entrypoint` set to the possibly-extensionless file

joshma:
cc @colinking - you should do a post-merge review on this

pkg/build/shell.go
joshma: I guess I can handle empty string more elegantly, right now it errosr below

pkg/build/builder.go
colinking: Could we do `.airplane/Dockerfile` instead?

joshma: Yeah good idea - https://github.com/airplanedev/cli/pull/228

PR #222, opened 2021-08-05
Add param env var support to shell shim

joshma:
Now shell-shim.sh accepts params from arguments and sets them as env
vars.

This:

    shell-shim.sh foo=v1 bar=v2

Becomes:

    PARAM_FOO=v1
    PARAM_BAR=v2

### DIFF
--- a/pkg/build/shell-shim.sh
+++ b/pkg/build/shell-shim.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-# Params are based in as e.g. param_slug_1=value1, param_slug_2=value2
-# Export as environment varaibles, e.g. AP_PARAM_SLUG_1=value1, AP_PARAM_SLUG_2=value2
+# Params are based in as param_slug_1=value1, param_slug_2=value2
+# Export as environment varaibles, PARAM_SLUG_1=value1, PARAM_SLUG_2=value2
 for param in "$@"; do
     param_slug="$(cut -d '=' -f 1 <<< "${param}")"
     param_value="$(cut -d '=' -f 2 <<< "${param}")"
--- a/pkg/build/shell-shim.sh
+++ b/pkg/build/shell-shim.sh
@@ -1,4 +1,14 @@
 #!/bin/bash
 
-# TODO: set up env vars - for each argument, split by first =, set AP_FIRST_PART=SECOND_PART
+# Params are based in as e.g. param_slug_1=value1, param_slug_2=value2
+# Export as environment varaibles, e.g. AP_PARAM_SLUG_1=value1, AP_PARAM_SLUG_2=value2
+for param in "$@"; do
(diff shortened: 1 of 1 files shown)


pkg/build/shell-shim.sh
```diff
@@ -1,4 +1,14 @@
 #!/bin/bash
 
-# TODO: set up env vars - for each argument, split by first =, set AP_FIRST_PART=SECOND_PART
+# Params are based in as e.g. param_slug_1=value1, param_slug_2=value2
+# Export as environment varaibles, e.g. AP_PARAM_SLUG_1=value1, AP_PARAM_SLUG_2=value2
+for param in "$@"; do
+    param_slug="$(cut -d '=' -f 1 <<< "${param}")"
+    param_value="$(cut -d '=' -f 2 <<< "${param}")"
+    # Convert to uppercase
+    var_name="$(echo "PARAM_${param_slug}" | tr '[:lower:]' '[:upper:]')"
```
francescaleung, line 9: AP_PARAM?

joshma: Yeah I was considering AP_ at first, then went to PARAM_, not sure if we want both 😅  - feels a bit long at that point. I figured PARAM_X was unique enough, AP_X we're using for our own config and AIRPLANE_ is being used for customer-facing things (e.g. AIRPLANE_RUNNER_ID). We could do AIRPLANE_PARAM_X but that's a bit too long.

@colinking adding this to your queue as well if you have input on it.

joshma: @francescaleung so...I'd prefer just PARAM_ over AP_PARAM_, does that make sense?

francescaleung: sure, just change the comment so it matches the code :)

francescaleung (approved):

pkg/build/shell-shim.sh
joshma: oops, ty :)

pkg/build/shell-shim.sh
```diff
@@ -1,4 +1,14 @@
 #!/bin/bash
 
-# TODO: set up env vars - for each argument, split by first =, set AP_FIRST_PART=SECOND_PART
+# Params are based in as param_slug_1=value1, param_slug_2=value2
+# Export as environment varaibles, PARAM_SLUG_1=value1, PARAM_SLUG_2=value2
```
colinking, line 4: Is `X=Y` a common format for parameters in shell land? I would have expected `--X=Y`

joshma: Shrug I think that's just positional args vs flags? Like we could `--X=Y` but then we'd have to do shell parsing. I didn't think too hard about this one since it's completely internal.

PR #227, opened 2021-08-09
Add basic implementation of PrepareRun for shell

joshma:
PrepareRun replicates passing parameters to .airplane/shim.sh, which
executes entrypoint like normal, with env vars set.

### DIFF
--- a/pkg/build/shell.go
+++ b/pkg/build/shell.go
@@ -28,7 +28,7 @@ func shell(root string, options api.KindOptions) (string, error) {
 
 	// Build off of the dockerfile if provided:
 	var dockerfileTemplate string
-	if dockerfilePath := filepath.Join(root, "Dockerfile"); fsx.Exists(dockerfilePath) {
+	if dockerfilePath := FindDockerfile(root); dockerfilePath != "" {
 		contents, err := ioutil.ReadFile(dockerfilePath)
 		if err != nil {
 			return "", errors.Wrap(err, "opening dockerfile")
@@ -102,4 +102,20 @@ func ShellShim(entrypoint string) (string, error) {
 	}
 
 	return shim, nil
+}
+
+// FindDockerfile looks for variants of supported Dockerfile locations and returns non-empty path
+// to the file, if found.
+func FindDockerfile(root string) string {
+	for _, filePath := range []string{
+		"Dockerfile.airplane",
+		".airplane/Dockerfile",
+		"Dockerfile",
+	} {
+		dockerfilePath := filepath.Join(root, filePath)
+		if fsx.Exists(dockerfilePath) {
+			return dockerfilePath
+		}
+	}
+	return ""
 }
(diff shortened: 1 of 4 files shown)


pkg/build/shell.go
```diff
@@ -103,3 +103,19 @@ func ShellShim(entrypoint string) (string, error) {
 
 	return shim, nil
 }
+
+// FindDockerfile looks for variants of supported Dockerfile locations and returns non-empty path
+// to the file, if found.
+func FindDockerfile(root string) string {
```
joshma, line 109: Pulled out this function to print a warning message that we don't support dockerfile - note this also adds support for `Dockerfile.airplane` and `.airplane/Dockerfile`

pkg/runtime/shell/shell.go
```diff
@@ -40,7 +43,44 @@ type Runtime struct{}
...
+	if err != nil {
+		return nil, err
+	}
+
+	if err := os.WriteFile(filepath.Join(root, ".airplane/shim.sh"), []byte(shim), 0644); err != nil {
+		return nil, errors.Wrap(err, "writing shim file")
+	}
+
+	cmd := []string{"bash", filepath.Join(root, ".airplane/shim.sh")}
+	// TODO: this is a rough approximation of how interpolateParameters works in prod
```
joshma, line 75: Felt a bit hacky but didn't want to have to pull out all the functionality here yet...

pkg/runtime/shell/shell.go
```diff
@@ -64,11 +104,8 @@ func (r Runtime) Workdir(path string) (string, error) {
 func (r Runtime) Root(path string) (string, error) {
 	root, ok := fsx.Find(path, "Dockerfile")
 	if !ok {
-		logger.Warning("No Dockerfile file found - using a basic Ubuntu image.")
```
joshma, line 67: I had to get rid of this because calls to `Root()` were printing this out repeatedly - can't think of a good place to put this, the general `script.go` file doesn't leave much extensibility. :( LMK if you have ideas.

pkg/build/shell.go
joshma: Actually - removed `.airplane/Dockerfile`, that's going to be reserved for ourselves and users should not be writing stuff there

pkg/runtime/shell/shell.go
```diff
@@ -40,7 +43,45 @@ type Runtime struct{}
...
-	return nil, errors.New("not supported")
+	root, err := r.Root(opts.Path)
+	if err != nil {
+		return nil, err
+	}
+
+	if dockerfilePath := build.FindDockerfile(root); dockerfilePath != "" {
+		logger.Warning("Found Dockerfile at %s.", dockerfilePath)
+		logger.Warning("`airplane dev` does not currently support running inside a Docker image.")
+		logger.Warning("The script will run inside your local machine environmetnn.")
```
colinking, line 54:
```suggestion
		logger.Warning("The script will run inside your local machine environment.")
```

colinking (approved):
🔥

PR #229, opened 2021-08-10
Fix interpolating paths into python-shim with backslashes

joshma:
On windows, TaskRoot and Entrypoint have a backslashes, resulting in
`sys.path.append("foo\bar")`, which results in unexpected behavior. (One
user had `C:\Users\...` which resulted in an invalid unicode escsape
sequence.)

To fix this, this uses raw strings so that backslashes are treated as
literals.

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -60,3 +60,14 @@ func inlineString(s string) string {
 	s = strings.ReplaceAll(s, "'", `'"'"'`)
 	return "echo '" + s + "'"
 }
+
+// backslashEscape escapes s by replacing `\` with `\\` and all runes in chars with `\{rune}`.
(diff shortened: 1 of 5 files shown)


colinking (approved):
Ah, good catch. Does this handle paths with double-quotes though? I figure we'll need to escape those before templating the shim. We likely have the same problem in the Node and Shell shims.

We had to do similar escaping for inserting the shims into Dockerfiles: https://github.com/airplanedev/cli/blob/cf52382f5e5becad4866b5e1d6cac61cca57b223/pkg/build/build.go#L55

joshma:
@colinking take a look at the slightly modified approach

colinking (approved):

PR #230, opened 2021-08-10
Detect expired CLI tokens

colinking:
This PR detects if a user's token is expired and sends them through the login flow again. This will happen every 3 months. We should eventually support auto-refreshing tokens.

This PR also adds a logger to `retryablehttp` that prints out HTTP logs when `--debug` is enabled.

This PR also tweaks `airplane login` to re-login even if currently logged in so that you don't need to `airplane logout` first.

### DIFF
--- a/pkg/cmd/auth/login/login.go
+++ b/pkg/cmd/auth/login/login.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 
 	"github.com/airplanedev/cli/pkg/analytics"
+	"github.com/airplanedev/cli/pkg/api"
 	"github.com/airplanedev/cli/pkg/cli"
 	"github.com/airplanedev/cli/pkg/conf"
 	"github.com/airplanedev/cli/pkg/logger"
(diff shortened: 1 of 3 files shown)


pkg/logger/http.go
```diff
@@ -0,0 +1,58 @@
+package logger
+
+import (
+	"strings"
+
+	"github.com/hashicorp/go-retryablehttp"
+)
+
+// HTTPLogger is a wrapper around the default logger that
+// can be used for HTTP requests via `hashicorp/go-retryablehttp`.
```
joshma, line 10: maybe mention that it goes to Debug?

pkg/cmd/auth/login/login.go
```diff
@@ -42,8 +41,28 @@ var (
...
+// validateToken returns a boolean indicating whether or not the current
+// client token is valid.
+func validateToken(ctx context.Context, c *cli.Config) (bool, error) {
+	if c.Client.Token == "" {
+		return false, nil
+	}
+
+	_, err := c.Client.AuthInfo(ctx)
+	if e, ok := err.(api.Error); ok && e.Code == 401 {
+		logger.Debug("Found an expired token. Re-authenticating.")
```
joshma, line 53: Is this actually re-authenticating though?

joshma (approved):

pkg/cmd/auth/login/login.go
colinking: Yeah it does, it re-prompts you to login and then updates your local token and then it continues on.

PR #231, opened 2021-08-10
Add explicit mention of all kinds in NeedsBuilding

joshma:
Makes it easier to notice if we forgot to handle a case or explicitly
don't want it to build.

(Super minor change since this is only used by legacy yml builder.)

### DIFF
--- a/pkg/build/builder.go
+++ b/pkg/build/builder.go
@@ -288,9 +288,13 @@ const (
 
 func NeedsBuilding(kind api.TaskKind) bool {
 	switch Name(kind) {
-	case NameGo, NameDeno, NamePython, NameNode, NameDockerfile:
+	case NameGo, NameDeno, NamePython, NameNode, NameDockerfile, NameShell:
 		return true
+	case NameImage:
+		return false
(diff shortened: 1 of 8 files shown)


pkg/build/builder.go
```diff
@@ -288,9 +288,13 @@ const (
...
 	switch Name(kind) {
-	case NameGo, NameDeno, NamePython, NameNode, NameDockerfile:
+	case NameGo, NameDeno, NamePython, NameNode, NameDockerfile, NameShell:
 		return true
+	case NameImage:
+		return false
 	default:
+		// Unexpected, default to false:
+		logger.Debug("NeedsBuilding got unexpected kind %s, defaulting to false", kind)
 		return false
```
colinking, line 298: wdyt of returning an error? Seems better to crash in this case

colinking (approved):

pkg/build/builder.go
joshma: yeaah i didnt want to add to signature but that probably seems best

PR #233, opened 2021-08-11
Auto-paginate `list` commands

colinking:
This PR adds auto-pagination for `runs list`. A similar strategy could be used for listing sessions and runbooks, but those aren't supported yet in the CLI.

You can see it working by adjusting the limit (from 1000 -> 1):

```sh
❯ lairlocal runs list --task ping --debug
# snip...
[debug] performing request GET https://api.airplane.so:5000/v0/runs/list?limit=1&page=0&taskID=tsk20210727z53qmsk
[debug] performing request GET https://api.airplane.so:5000/v0/runs/list?limit=1&page=1&taskID=tsk20210727z53qmsk
[debug] performing request GET https://api.airplane.so:5000/v0/runs/list?limit=1&page=2&taskID=tsk20210727z53qmsk
[debug] performing request GET https://api.airplane.so:5000/v0/runs/list?limit=1&page=3&taskID=tsk20210727z53qmsk
          ID         |  STATUS   |        CREATED AT         |         ENDED AT
---------------------+-----------+---------------------------+----------------------------
  run20210811z5hfzuv | Succeeded | 2021-08-11T10:12:29-04:00 | 2021-08-11T09:35:27-04:00
  run20210811zuyzgch | Succeeded | 2021-08-11T10:12:26-04:00 | 2021-08-11T09:35:27-04:00
  run20210811zo02z4m | Succeeded | 2021-08-11T09:35:23-04:00 | 2021-08-11T09:35:27-04:00
```

### DIFF
--- a/pkg/api/client.go
+++ b/pkg/api/client.go
@@ -10,6 +10,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"strconv"
 	"strings"
 	"time"
 
@@ -22,6 +23,11 @@ import (
 var (
 	// Client tolerates minor outages and retries.
 	client *http.Client
+)
+
+const (
+	// pageLimit is the default number of results to fetch per paginated request.
+	pageLimit = 1000
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


francescaleung:
I feel like asking for all the runs, all the time, is a bad idea... can we add other params so people can still see more than just the last 100, but not show all the runs?

colinking:
@francescaleung there's more context in the Linear ticket: https://linear.app/airplane/issue/AIR-1509/[emi]-add-global-runs-list-with-auto-pagination

LMK what you think

pkg/api/client.go
```diff
@@ -24,6 +25,11 @@ var (
 	client *http.Client
 )
 
+const (
+	// pageLimit is the default number of results to fetch per paginated request.
+	pageLimit = 1000
```
joshma, line 30: This feels high? Fetching 100 objects seems saner.

joshma (approved):
OK with this change, but I'm having second thoughts on how `runs list` should work.

1. Should've thought of this earlier, but we'll need start/end times for it to be useful to customers.
2. Is this reverse cron or cron? If this is reverse cron order, then capping at an absolute max could make sense here - maybe refuse after 10k runs or something.
3. Especially because this isn't streaming, the user has to sit and see all runs downloaded before they can do something with it. I think it's just more reason to not download _all_ runs?

pkg/api/client.go
colinking: Looking at some example runs, the JSON representation is ~1KB so 1000 results is up to 1MB. Doing one 1MB request sounds like the preferred approach vs. ten 100KB requests.

joshma: > Doing one 1MB request sounds like the preferred approach vs. ten 100KB requests.

Is that for any specific reason? 1MB is a lot, the overhead on each request is going to be small compared to the transfer time and so ten 100KB would probably get approx the same perf at the network layer. Didn't realize that each item is 1KB, that actually makes me more inclined to push for a smaller limit.

On the DB layer, doing 100 items at a time helps reduce how long a transaction takes, which helps in general when it comes to locking etc. (not that this specific call is high traffic, but esp. if there's a bunch of list operations going on). And the more items returned, the more expensive bulk joins can be if the planner loses confidence in cardinality (thinks it'll pull in a large % of e.g. users).

colinking:
https://github.com/airplanedev/cli/pull/235

PR #235, opened 2021-08-13
Add `--since/--until` filtering to `runs list`

colinking:
This PR updates `runs list` s.t. you can optionally fetch all runs. It does this by introducing a `--since`, `--until`, and `--limit` flag. `--limit` defaults to `100`, but can be overridden to `0` (or any negative number) to fetch all runs. This can be optionally combined with `--since` and `--until` to filter by a time range.

For example, you could list all runs with:

```
airplane runs list --until $(date +%Y-%m-%dT%H:%M:%S%z) --limit=0
```

To support time-based flags, which cobra does not support out-of-the-box (!), I've added a `utils.TimeValue` which is a `pflag.Value`. This means you can use it with `cmd.Flags().Var(&timeValue, ...)`. It tries to be generous about parsing times and provide useful error messages.

The auto-pagination is a bit complicated. Would be a good use-case for generics!

### DIFF
--- a/pkg/cmd/runs/list/list.go
+++ b/pkg/cmd/runs/list/list.go
@@ -39,8 +39,8 @@ func New(c *cli.Config) *cobra.Command {
 
 	cmd.Flags().StringVarP(&cfg.slug, "task", "t", "", "Filter runs by task slug")
 	cmd.Flags().IntVar(&cfg.limit, "limit", 100, "If >0, returns at most --limit items.")
-	cmd.Flags().Var(&cfg.since, "since", "Filter runs by the time they were created at")
-	cmd.Flags().Var(&cfg.until, "until", "Filter runs by the time they were created at")
+	cmd.Flags().Var(&cfg.since, "since", "Include only runs created after the given time")
+	cmd.Flags().Var(&cfg.until, "until", "Include only runs created before the given time")
 
 	return cmd
 }
--- a/pkg/cmd/runs/list/list.go
+++ b/pkg/cmd/runs/list/list.go
@@ -2,18 +2,26 @@ package list
 
 import (
 	"context"
+	"time"
 
 	"github.com/MakeNowJust/heredoc"
 	"github.com/airplanedev/cli/pkg/api"
 	"github.com/airplanedev/cli/pkg/cli"
 	"github.com/airplanedev/cli/pkg/print"
+	"github.com/airplanedev/cli/pkg/utils"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
+type config struct {
+	slug  string
+	since utils.TimeValue
+	until utils.TimeValue
+}
+
 // New returns a new list command.
 func New(c *cli.Config) *cobra.Command {
-	var slug string
+	var cfg config
 
 	cmd := &cobra.Command{
 		Use:   "list",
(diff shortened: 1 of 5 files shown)


pkg/utils/cobra.go
```diff
@@ -22,3 +28,53 @@ func WithParentPersistentPreRunE(f func(cmd *cobra.Command, args []string) error
...
+//
+// Which could be set as: `--since="2020-01-02 01:02:03"`
+//
+// TimeValue's are alias types of time.Time. You can convert safely via `time.Time(tv)`.
+type TimeValue time.Time
+
+var _ pflag.Value = &TimeValue{}
+
+func (tv *TimeValue) Set(s string) error {
+	for _, format := range []string{
```
joshma, line 47: Can we add a few more formats?

- Date without time (2006-01-02)
- I may be mis-reading this, how does `--since="2020-01-02 01:02:03"` work above? Don't we need to add a format w/o the T? (Can we?)
- Date with time but not seconds (2006-01-02 15:04)

pkg/utils/cobra.go
```diff
@@ -22,3 +28,53 @@ func WithParentPersistentPreRunE(f func(cmd *cobra.Command, args []string) error
...
+	} {
+		v, err := time.ParseInLocation(format, s, time.Now().Location())
+		if err == nil {
+			*tv = TimeValue(v)
+			return nil
+		}
+	}
+
+	// If we did not find a match, return a helpful error message:
+	return errors.New("expected timestamp formatted as '2006-01-02T15:04:05' (local time) or '2006-01-02T15:04:05+07:00'")
```
joshma, line 63: This error message can use friendlier examples if we support e.g. `2006-01-02 15:04`

pkg/utils/cobra.go
```diff
@@ -22,3 +28,53 @@ func WithParentPersistentPreRunE(f func(cmd *cobra.Command, args []string) error
...
+func (tv *TimeValue) Set(s string) error {
+	for _, format := range []string{
+		// If a user does not specify a time zone, we interpret the time zone
+		// as local time:
+		"2006-01-02T15:04:05", // RFC3339 without the "Z07:00"
+		// Otherwise, we look for a time zone.
+		time.RFC3339,
+		"2006-01-02T15:04:05Z0700", // RFC3339 without the timezone ":"
+	} {
+		v, err := time.ParseInLocation(format, s, time.Now().Location())
```
joshma, line 55: Hrm, is it awkward to pipe out an error here? If you enter an invalid time this ends up causing the time to get ignored yeah?

pkg/cmd/runs/list/list.go
```diff
@@ -24,24 +33,31 @@ func New(c *cli.Config) *cobra.Command {
...
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return run(cmd.Root().Context(), c, slug)
+			return run(cmd.Root().Context(), c, cfg)
 		},
 	}
 
-	cmd.Flags().StringVarP(&slug, "task", "t", "", "Filter runs by task slug")
+	cmd.Flags().StringVarP(&cfg.slug, "task", "t", "", "Filter runs by task slug")
+	cmd.Flags().IntVar(&cfg.limit, "limit", 100, "If >0, returns at most --limit items.")
+	cmd.Flags().Var(&cfg.since, "since", "Filter runs by the time they were created at")
```
joshma, line 42: Maybe something like this for since (and similar `before the given time` for until)?

```suggestion
	cmd.Flags().Var(&cfg.since, "since", "Filter runs created since the given time")
```

pkg/cmd/runs/list/list.go
```diff
@@ -24,24 +33,31 @@ func New(c *cli.Config) *cobra.Command {
...
 		`),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return run(cmd.Root().Context(), c, slug)
+			return run(cmd.Root().Context(), c, cfg)
 		},
 	}
 
-	cmd.Flags().StringVarP(&slug, "task", "t", "", "Filter runs by task slug")
+	cmd.Flags().StringVarP(&cfg.slug, "task", "t", "", "Filter runs by task slug")
+	cmd.Flags().IntVar(&cfg.limit, "limit", 100, "If >0, returns at most --limit items.")
```
joshma, line 41: This limit can prob be higher, maybe 1,000?

Although overall I'm wondering what a user would do if they do a `--since` and hit the limit. Doing another `--since` is awkward because there might be overlap. You could do `--since + --until` to avoid hitting the limit? And if you did it week-by-week that'd be OK?

joshma (approved):

pkg/utils/cobra.go
colinking: Cobra doesn't support spaces in string flags afaict: https://github.com/spf13/cobra/issues/1114

For example:

```
❯ lairlocal runs list --since="2020 01 01"

Error: Invalid argument "2020" for "--since" flag: expected timestamp formatted as '2006-01-02T15:04:05' (local time) or '2006-01-02T15:04:05+07:00'
```

Added a few other formats that don't include spaces and fixed the comment above.

colinking:
@joshma re-requesting

pkg/cmd/runs/list/list.go
joshma: Yeah but that's still sort of awkward if you _want_ to do it in batches.

But this is fine, I don't know if we need to put more into this for now.

pkg/utils/cobra.go
joshma: Wow that's nuts!

joshma (approved):

PR #240, opened 2021-08-19
Improve runtime detection to favor task kind

joshma:
Rather than solely base it off of extension, determine runtime by
figuring out the task kind first. In some cases, like TS/JS, the exact
runtime will depend on the extension.

This allows using e.g. some_task.rb with shell script, since some_task
can already be created with kind=shell.

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -9,11 +9,11 @@ package runtime
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"path/filepath"
 
 	"github.com/airplanedev/cli/pkg/api"
+	"github.com/pkg/errors"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


pkg/runtime/runtime.go
```diff
@@ -104,11 +104,24 @@ func Register(ext string, r Interface) {
...
+		if runtime.Kind() != kind {
+			continue
+		}
+		if pathExt == ext {
+			return runtime, true
+		}
+		possible = append(possible, runtime)
+	}
+	if len(possible) != 1 {
+		return nil, false
```
colinking, line 122: Should we return an error here instead?

colinking (approved):

pkg/runtime/runtime.go
joshma: Hrm yeah I'll just make the func return error instead of bool

PR #242, opened 2021-08-20
Resolve user-supplied `tsconfig.json` not getting applied

colinking:
If you run `tsc` with a path to a `.ts` file to build, f.e. `tsc ./hello.ts`, then `tsc` will not look for `tsconfig.json` configs. Honestly, that's surprising to me considering that `tsconfig.json` configs specify a _lot_ more than just what files to build.

To workaround this, we generate our own `tsconfig.json` that use the `"files"` option to specify the path of the file to compile and `"extends"` to allow a user to override this `tsconfig.json`.

---

This PR also caught a bug with how we install shim dependencies that fails any builds that used a custom root. 

What happens is `tsc` attempts to compile `root/.airplane/shim.ts` which requires you have `@types/node` or else it won't know what `process`, `Promise`, etc. are. We previously fixed this issue by requiring users to install `@types/node`, however it turns out that breaks for custom roots. 

Let's see an example. Say you have a directory structure like this:

```sh
❯ tree root -a
root
├── .airplane
│   └── shim.ts
└── tasks
    ├── my-task.ts
    ├── node_modules
    └── package.json

3 directories, 3 files
```

In this case, `@types/node` is installed in `root/tasks/node_modules`. However, when `tsc` tries to find `@types/node` while compiling `root/.airplane/shim.ts`, it'll look inside of `root/.airplane/node_modules` and `root/node_modules`, neither of which exist / contain `@types/node`. (if you didn't have a custom root, then the root and workdir would be the same, in which case we would run `npm install` in the root to create a `root/node_modules` which will have `@types/node`).

This PR introduces a workaround that actually allows us to avoid having users install `@types/node` -- we instead manage the shim's dependencies inside of `root/.airplane/node_modules` via a `root/.airplane/package.json` that we create at build time.

Sidenote: if you're debugging an issue with `tsc` not finding a module as you'd expect, try running it with `tsc --traceResolution`. There are also some great docs on module resolution here: https://www.typescriptlang.org/docs/handbook/module-resolution.html

---

I tested this in a few ways with both the `deploy` and `dev` commands (just the latter is listed below):

<details>
<summary>Without tsconfig</summary>

This simply confirms that these commands continue to work without a `tsconfig.json`. In other words, it tests the path where we do not set an `extends` and instead use our default values.

```sh
$ tree test1-no-tsconfig -I node_modules
test1-no-tsconfig
├── lib
│   └── greetings.ts
├── main.ts
├── package-lock.json
└── package.json

1 directory, 4 files

$ cat test1-no-tsconfig/main.ts
// Linked to https://app.airplane.so:5000/t/tsconfig_test [do not edit this line]

import { hello } from './lib/greetings'

export default async function() {
  console.log(`test1: ${hello}`);
}

$ lairlocal dev ./test1-no-tsconfig/main.ts
[log] test1: Hello, world!
```

</details>

<details>
<summary>With tsconfig</summary>

This tests that we properly extend the user-provided tsconfig. I confirmed this by setting `compilerOptions.listFiles` and confirming that the `tsc` command printed out a list of found files.

```sh
$ tree test2-tsconfig -I node_modules
test2-tsconfig
├── lib
│   └── greetings.ts
├── main.ts
├── package-lock.json
├── package.json
└── tsconfig.json

1 directory, 5 files

$ cat test2-tsconfig/tsconfig.json
{
  "compilerOptions": {
    "listFiles": true
  }
}

$ cat test2-tsconfig/main.ts
// Linked to https://app.airplane.so:5000/t/tsconfig_test [do not edit this line]

import { hello } from './lib/greetings'

export default async function() {
  console.log(`test2: ${hello}`);
}

$ lairlocal dev ./test2-tsconfig/main.ts
// (listed files snipped)
[log] test2: Hello, world!
```

</details>

<details>
<summary>With custom root</summary>

This confirms that we pick the tsconfig closest to the user's entrypoint and properly handle custom roots.

```sh
$ tree test3-tsconfig-alt-root -I node_modules
test3-tsconfig-alt-root
├── lib
│   └── greetings.ts
├── package.json
├── tasks
│   ├── main.ts
│   ├── package.json
│   └── tsconfig.json
└── tsconfig.json

2 directories, 6 files

$ cat test3-tsconfig-alt-root/tasks/tsconfig.json
{
  "compilerOptions": {
    "listFiles": true
  }
}

$ cat test3-tsconfig-alt-root/tasks/main.ts
// Linked to https://app.airplane.so:5000/t/tsconfig_test [do not edit this line]

import { hello } from '../lib/greetings'

export default async function() {
  console.log(`test3: ${hello}`);
}

$ lairlocal dev ./test3-tsconfig-alt-root/tasks/main.ts --debug
// (listed files snipped)
[log] test3: Hello, world!
```

</details>


colinking:
(Sending by @joshma since he has worked on the Node builder before)

pkg/build/node_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/airplanedev/cli/pkg/api"
+	"github.com/stretchr/testify/require"
+)
+
+// -(map[string]interface {}) {
```
joshma, line 11: (can delete?)

pkg/build/node.go
```diff
@@ -96,31 +104,137 @@ func node(root string, options api.KindOptions) (string, error) {
...
+		rp, err := filepath.Rel(filepath.Join(root, ".airplane"), p)
+		if err != nil {
+			return nil, errors.Wrap(err, "creating relative tsconfig path")
+		}
+		tsconfig.Extends = rp
+	}
+
+	// Apply defaults to a few of the tsconfig fields, but let the user override
+	// them from their tsconfig.json:
+	if utsc.CompilerOptions.AllowJS == nil {
```
joshma, line 204: maybe debug log each of these values (AllowJS, Module, etc.) as you set them? i'm guessing they'll come in handy later

joshma (approved):

PR #252, opened 2021-08-24
Set `BuildID` from `airplane deploy`

colinking:
This PR updates `airplane deploy` to set the `buildID` field after running a remote build, at the same point at which it sets the `Image`. This will be used to show build metadata for tasks and runs.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -46,7 +46,8 @@ type UpdateTaskRequest struct {
 	RequireExplicitPermissions bool              `json:"requireExplicitPermissions" yaml:"-"`
 	Permissions                Permissions       `json:"permissions" yaml:"-"`
 	// TODO(amir): friendly type here (120s, 5m ...)
-	Timeout int `json:"timeout" yaml:"timeout"`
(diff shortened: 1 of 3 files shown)


pkg/api/types.go
```diff
@@ -46,7 +46,8 @@ type UpdateTaskRequest struct {
 	RequireExplicitPermissions bool              `json:"requireExplicitPermissions" yaml:"-"`
 	Permissions                Permissions       `json:"permissions" yaml:"-"`
 	// TODO(amir): friendly type here (120s, 5m ...)
-	Timeout int `json:"timeout" yaml:"timeout"`
+	Timeout int     `json:"timeout" yaml:"timeout"`
+	BuildID *string `json:"buildID" yaml:"-"`
```
joshma, line 50: (wondering out loud, why is there even a yaml tag?)

joshma (approved):

pkg/api/types.go
colinking: Took a peek around, doesn't seem to be used compared to the other yaml tags in this file, which are either used for yaml printing or reading `airplane.yml` files. Went ahead and removed them.

PR #256, opened 2021-08-30
Normalize entrypoint

yields:
When on windows, and we are about to send a build API request we normalize
the entrypoint so that it uses forward-slash regardless of the OS.


pkg/build/remote.go
```diff
@@ -92,11 +93,18 @@ func updateKindAndOptions(ctx context.Context, client *api.Client, def definitio
...
 		return err
 	}
+
 	// Conditionally instruct the remote builder API to perform a shim-based build.
 	if shim {
 		kindOptions["shim"] = "true"
 	}
 
+	// Make sure we are using `/` regardless of the OS.
+	if ep, ok := kindOptions["entrypoint"].(string); ok && runtime.GOOS == "windows" {
```
joshma, line 103: Is this the right function to be doing this in? Might be the best for now but I'm wondering if (1) there's other paths in kindOptions we're missing and (2) if there's other places where entrypoint is set and we're not normalizing it. I think the answer to (2) is that this is currently right, and it'll change with AIR-1767 anyways. And for (1), double check that node workdir vs root are handled correctly?

joshma (approved):

PR #263, opened 2021-09-10
Fix execute with slug

joshma:
Fixes bug where CLI assumes slug is a file - it errors when you use a slug,
saying the file does not exist.  Instead, we should check for a file and - if
not - look up the task.


pkg/cmd/tasks/execute/execute.go
```diff
@@ -73,11 +75,18 @@ func New(c *cli.Config) *cobra.Command {
 func run(ctx context.Context, cfg config) error {
 	var client = cfg.root.Client
 
-	slug, err := slugFrom(cfg.task)
-	if err != nil {
-		return err
+	var slug string
+	var err error
+	if _, err := os.Stat(cfg.task); errors.Is(err, os.ErrNotExist) {
```
colinking, line 80: Can you you also treat it as a slug if it matches a directory? For example:

```sh
~ /
$ cat /my_task/main.ts
# Linked to my_task
...

~ /
$ airplane exec my_task
~ /
$ airplane exec my_task/main.ts
~ /my_task
$ airplane exec main.ts
```

```suggestion
	if f, err := os.Stat(cfg.task); errors.Is(err, os.ErrNotExist) || f.IsDir() {
```

colinking (approved):

PR #265, opened 2021-09-14
Deemphasize API Key ID in output message when creating key

joshma:
Users often use the key ID instead of the key value because the message
prints the key ID last.

This makes it easier to read the key instead:

```
  Creating API key named API Key (created 2021-09-14T16:11:27-07:00)...
  Done! Created API key key20210914zyyy.

  Save this somewhere safe, as you won't be able to retrieve it later:

  API Key: tkn_xxx

  (Team ID: xyz)
```


pkg/cmd/apikeys/create/create.go
```diff
@@ -50,15 +50,15 @@ func run(ctx context.Context, c *cli.Config, name string) error {
 	}
 
 	apiKey := resp.APIKey
-	logger.Log("  Done!")
+	logger.Log("  Done! Created API key %s.", apiKey.ID)
```
colinking, line 53: Does the API key ID provide any value to customers? If anything, should we debug log it instead?

pkg/cmd/apikeys/create/create.go
```diff
@@ -50,15 +50,15 @@ func run(ctx context.Context, c *cli.Config, name string) error {
...
+		logger.Log(`
+  Save this somewhere safe, as you won't be able to retrieve it later:
 
-  %s
+  API Key: %s
 
-  Key ID: %s
-  Team ID: %s
-`, logger.Blue(apiKey.Key), apiKey.ID, apiKey.TeamID)
+  (Team ID: %s)
```
colinking, line 60: The team ID is only useful when configuring an agent, right? What if we didn't print the team ID and instead updated the agent docs with a command for fetching the team ID? `airplane auth info`

We should also prioritize updating our api key format to include the team ID so folks don't even need that

colinking (approved):

pkg/cmd/apikeys/create/create.go
joshma: Hmmm yes, but we only really use api keys when setting up agents. But OK, I think `airplane auth info` will work. :)

pkg/cmd/apikeys/create/create.go
joshma: 👍

PR #267, opened 2021-09-16
Add support for ESM in node builder

colinking:
This PR updates our Node builder and Node dev workflow to use [`esbuild`](https://esbuild.github.io/) instead of `tsc`. Doing so adds support for tsconfig aliases and importing ECMAScript modules. This PR adds tests for both using the [new test suite](https://github.com/airplanedev/cli/pull/268).

We made this move because we wanted to add support for building Node tasks that import ESM packages. Historically, npm packages have been released as CommonJS, but the new [general recommendation is to release as ESM](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c). In order to import ESM packages from a TS project, [you need to configure your TS project as an ES module](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-make-my-typescript-project-output-esm). Instead, we switch to `esbuild` which [will convert ESM packages to CommonJS](https://esbuild.github.io/api/#format-commonjs).

The primary trade-off of using esbuild over tsc is that we no longer perform type-checking as part of the build process. I'm willing to accept that over [instructing folks to follow these instructions](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-make-my-typescript-project-output-esm) in order to deploy TS code to Airplane. I'd bet it'll become easier to interop between CommonJS + ESM with tsc over time, at which point we could revisit this.

Resolves: AIR-1734, AIR-1839


pkg/build/node.go
```diff
@@ -71,12 +71,7 @@ func node(root string, options api.KindOptions) (string, error) {
 	}
 	cfg.InlineShim = inlineString(shim)
 
-	tsconfig, err := GenTSConfig(root, filepath.Join(root, entrypoint), options)
-	if err != nil {
-		return "", err
-	}
-	cfg.InlineTSConfig = inlineString(string(tsconfig))
-
+	// TODO: update this comment
```
joshma: guessing you meant to update this comment?

pkg/build/node.go
```diff
@@ -46,20 +41,22 @@ func node(root string, options api.KindOptions) (string, error) {
...
 		InlineShim            string
-		InlineTSConfig        string
 		InlineShimPackageJSON string
+		NodeVersion           string
 	}{
 		Workdir:        workdir,
 		HasPackageJSON: fsx.AssertExistsAll(filepath.Join(root, "package.json")) == nil,
 		IsYarn:         fsx.AssertExistsAll(filepath.Join(root, "yarn.lock")) == nil,
+		// esbuild is relatively generous in the node versions it supports:
+		// https://esbuild.github.io/api/#target
```
joshma, line 51: I think you can delete this, you already have that comment in GetNodeVersion

pkg/runtime/javascript/javascript.go
```diff
@@ -198,18 +188,25 @@ func (r Runtime) PrepareRun(ctx context.Context, opts runtime.PrepareRunOptions)
...
+		Bundle: true,
+
+		EntryPoints: []string{filepath.Join(tmpdir, "shim.js")},
+		Outfile:     filepath.Join(tmpdir, "dist/shim.js"),
+		Write:       true,
+
+		Platform: esbuild.PlatformNode,
+		Engines: []esbuild.Engine{
+			// esbuild is relatively generous in the node versions it supports:
+			// https://esbuild.github.io/api/#target
```
joshma, line 201: (same, can delete this comment?)

pkg/runtime/javascript/javascript.go
```diff
@@ -198,18 +188,25 @@ func (r Runtime) PrepareRun(ctx context.Context, opts runtime.PrepareRunOptions)
...
+	for _, e := range res.Errors {
+		logger.Warning("esbuild(error): %v", e)
 	}
-	cmd.Dir = root
-	logger.Debug("Running %s (in %s)", logger.Bold(strings.Join(cmd.Args, " ")), root)
-	out, err = cmd.CombinedOutput()
-	if err != nil {
-		logger.Log(strings.TrimSpace(string(out)))
-		return nil, nil, errors.Errorf("failed to compile %s", opts.Path)
+	for _, w := range res.Warnings {
```
joshma, line 208: print warnings first? so it ends on the errors

joshma (approved):

pkg/build/node.go
colinking: Nah, I had removed the comment from `GetNodeVersion`. Think you saw two different versions of the files, since I was pushing changes to that as you were reviewing :)

PR #268, opened 2021-09-16
Add test suite for testing build/dev workflows

colinking:
This PR adds a test suite that we can use to validate the behavior of our builders and dev workflows. This PR also hooks up the test suite for our Node builder and Node dev workflows. In future PRs, we should hook up our other builders to this suite.


pkg/runtime/typescript/typescript_test.go
```diff
@@ -0,0 +1,88 @@
...
+		},
+		{
+			Kind: api.TaskKindNode,
+			Opts: runtime.PrepareRunOptions{
+				Path: "typescript/esnext/main.ts",
+				// See nodeVersion comment above.
+				KindOptions: api.KindOptions{"nodeVersion": "14"},
+			},
+		},
+		// TODO: debug why yarn workspaces aren't working. Seems like we would need to compile
```
joshma, line 66: Hmm well you're running `npm install` below, should it be `yarn install`? 😅 

IIUC it should be a `yarn install` inside `pkg2/`

joshma (approved):

pkg/build/node.go
```diff
@@ -214,12 +217,27 @@ func GenTSConfig(root string, entrypoint string, opts api.KindOptions) ([]byte,
...
+		}
 	}
+	if strings.HasPrefix(nodeVersion, "14") {
+		// Node 14 supports ES2020: https://node.green/#ES2020
+		target = "es2020"
+	} else if strings.HasPrefix(nodeVersion, "15") || strings.HasPrefix(nodeVersion, "16") || nodeVersion == "" {
+		// Node 15 and 16 support ES2022: https://node.green/#ES2022
+		// However, tsc does not yet support specifying es2022, so we use esnext instead:
+		// https://github.com/microsoft/TypeScript/issues/44571
+		target = "esnext"
```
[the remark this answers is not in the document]
colinking: It just instructs `esbuild` not to compile any modern syntax down. So basically a no-op.

pkg/build/node.go
```diff
@@ -214,12 +217,27 @@ func GenTSConfig(root string, entrypoint string, opts api.KindOptions) ([]byte,
...
+	// https://node.green/#ES2019
+	target := "es2019"
+	var nodeVersion string
+	if opts != nil && opts["nodeVersion"] != nil {
+		nv, ok := opts["nodeVersion"].(string)
+		if ok {
+			nodeVersion = nv
+		}
 	}
+	if strings.HasPrefix(nodeVersion, "14") {
```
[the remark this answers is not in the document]
colinking: Heh, good call. Tweaked to parse with `semver`!

pkg/examples/examples.go
```diff
@@ -0,0 +1,19 @@
+package examples
```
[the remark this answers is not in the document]
colinking: It's not a common convention, afaik. But I think it most accurately describes the contents of the folder (over something generic like `tests/`). If anything, I believe it should live underneath the `build/` folder to show that these are build examples. But since they're used by both `runtime/` and `build/`, neither makes sense. Maybe we revisit once we merge those two packages?

pkg/runtime/typescript/typescript_test.go
colinking: Haha, fair enough though the problem is not with `yarn` but with `tsc`. `tsc` needs the type declarations so we need to compile the other package first. I believe the issue is with the `tsconfig` configuration where we use `references`.  

I've spent enough time on this for now. Will revisit in the future!

joshma: So does this mean yarn workspaces don't work right now? So e.g. if victor@luma tries picking it back up it we won't be able to support it?

joshma: Oh nevermind I'm mixing up PRs, this is just an issue with tsc in this test env?

PR #274, opened 2021-09-24
Mark pg-native as external as a temporary workaround

joshma:
Until we figure out a more general solution to this, this tells esbuild
to NOT compile in pg-native.

The pg package does `require("pg-native)`, but it marks pg-native as an
optional peer dependency. If it can't require it, it silently skips.

However, esbuild tries to bundle it in, and thus it hits errors. It
seems like folks have solved this in other bundlers like webpack by
ignoring the module, shimming it, etc.

The temporary solution is to mark it as external, so we don't attempt to
build it into the bundle.


pkg/build/node.go
```diff
@@ -123,6 +123,7 @@ func node(root string, options api.KindOptions) (string, error) {
 		RUN {{.InlineShim}} > /airplane/.airplane/shim.js && \
 			esbuild /airplane/.airplane/shim.js \
 				--bundle \
+				--external:pg-native \
```
colinking, line 126: You'll need to do this in `dev` too, where we call `esbuild` via Go directly.

Could you add a comment with context, too?

colinking (approved):
How does this fix the issue? Is `pg-native` correctly installed for users in their `node_modules`?

pkg/build/node.go
joshma: 👍

colinking: Oh I see about the optional peer dep. Interesting.

joshma:
> How does this fix the issue? Is `pg-native` correctly installed for users in their `node_modules`?

Users in this case _don't_ have pg-native installed, and this is an optional dep. So (1) esbuild doesn't attempt to find pg-native, and then (2) at run-time nothing happens (require fails, pg moves along).

If a user _did_ have pg-native installed, I believe we'd still `npm install` it and `require` should work.

PR #275, opened 2021-09-28
Mark all node dependencies as external for esbuild

joshma:
Read package.json and return all dependencies and devDependencies as
external for esbuild.

This fixes the issue where e.g. pg requires pg-native but gracefully
handles the module not being installed. esbuild will error on this when
trying to bundle.

We don't really need esbuild to bundle _anything_, so we just mark all
dependencies as external.


pkg/build/node.go
```diff
@@ -43,15 +46,30 @@ func node(root string, options api.KindOptions) (string, error) {
...
+	// Workaround to get esbuild to not bundle dependencies.
+	// See nodeexternals.ExternalPackages for details.
+	if cfg.HasPackageJSON {
+		deps, err := nodeexternals.ExternalPackages(pathPackageJSON)
+		if err != nil {
+			return "", err
+		}
+		var flags []string
+		for _, dep := range deps {
+			flags = append(flags, fmt.Sprintf("--external:%s", dep))
```
joshma, line 68: I'm... a bit anxious about whether or not this will work but it seems OK?

joshma:
cc @zhan-airplane  - this is a temporary fix, longer term i think we should `npm install` some build library that can do `esbuild` + other potential transforms

pkg/build/nodeexternals/nodeexternals.go
```diff
@@ -0,0 +1,45 @@
+package nodeexternals
```
colinking, line 1: We should avoid small single-use packages like this. Is there a reason you need to pull this function out? Can you move this into the `build` package? You could put it in `node_externals.go` if you still want to keep it separate from `node.go`, but I don't think you need to do that.

pkg/build/nodeexternals/nodeexternals.go
```diff
@@ -0,0 +1,45 @@
...
+		return nil, errors.Wrap(err, "opening package.json")
+	}
+	defer f.Close()
+	b, err := ioutil.ReadAll(f)
+	if err != nil {
+		return nil, errors.Wrap(err, "reading package.json")
+	}
+	var d struct {
+		Dependencies    map[string]string `json:"dependencies"`
+		DevDependencies map[string]string `json:"devDependencies"`
```
colinking, line 32: There are also `optionalDependencies`, maybe include that? https://github.com/sindresorhus/type-fest/blob/9cfbfc5acc4bdbc8abbdbbe8e1154c4bfc8fae7a/source/package-json.d.ts#L468-L486

colinking (approved):

pkg/build/node.go
```diff
@@ -120,12 +138,9 @@ func node(root string, options api.KindOptions) (string, error) {
 		RUN npm install
 		{{end}}
 
-		# --external:pg-native temporarily fixes issue where pg requires pg-native and esbuild
-		# fails when it's not installed (it's an optional dependency).
 		RUN {{.InlineShim}} > /airplane/.airplane/shim.js && \
 			esbuild /airplane/.airplane/shim.js \
-				--bundle \
-				--external:pg-native \
+				--bundle {{.ExternalFlags}} \
```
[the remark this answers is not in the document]
joshma: Yeaah I just didn't like the empty line - can I move this `--platform=node`? :)

PR #277, opened 2021-09-30
Fix filepaths on windows

zhan-airplane:
Turns out that `filepath.Clean()` undoes the work of `filepath.ToSlash()`. Also adds in something to avoid infinite-recursion on some Windows systems where the root is `C:\` or something along those lines rather than the empty string

### DIFF
--- a/pkg/build/remote.go
+++ b/pkg/build/remote.go
@@ -100,7 +100,7 @@ func updateKindAndOptions(ctx context.Context, client *api.Client, def definitio
 
 	// Normalize entrypoint to `/` regardless of OS.
 	if ep, ok := kindOptions["entrypoint"].(string); ok {
-		kindOptions["entrypoint"] = filepath.Clean(filepath.ToSlash(ep))
+		kindOptions["entrypoint"] = filepath.ToSlash(filepath.Clean(ep))
 	}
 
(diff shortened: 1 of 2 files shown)


zhan-airplane:
cc @joshma

pkg/build/remote.go
```diff
@@ -100,7 +100,7 @@ func updateKindAndOptions(ctx context.Context, client *api.Client, def definitio
 
 	// Normalize entrypoint to `/` regardless of OS.
 	if ep, ok := kindOptions["entrypoint"].(string); ok {
-		kindOptions["entrypoint"] = filepath.Clean(filepath.ToSlash(ep))
+		kindOptions["entrypoint"] = filepath.ToSlash(filepath.Clean(ep))
```
colinking, line 103: Is this something we could add tests for? Or is it too dependent on OS-inspection?

colinking (approved):

pkg/build/remote.go
zhan-airplane: Depends on if we can run tests in a Windows environment easily, I'll take a quick look and see haha

joshma:
Eugh ty ty @zhan-airplane @colinking for this quick fix!

We can probably set up something on GHA using a windows runner?

PR #279, opened 2021-10-04
Normalize entrypoint to match case on filesystem

joshma:
First does a refactor to clean up some Definition code. Removes an extraneous
`filepath.Clean` call, which isn't needed because all of the calls around it
clean the path already.

Adds a fix to find the proper casing of entrypoint by listing the directory and
comparing (case-insensitive) to get the proper filesystem casing.

This fixes a bug where a user can `airplane deploy somescript.js`, the filename
is called `someScript.js`, and we later error trying to open `somescript.js` on
a case-sensitive filesystem (e.g. ext4 on linux).


joshma:
@zhan-airplane I have this broken out as 2 commits for easier review

pkg/utils/pathcase/pathcase.go
```diff
@@ -0,0 +1,37 @@
...
+	files, readErr := os.ReadDir(dir)
+	if readErr != nil {
+		logger.Debug("error reading directory %q - going to iterate through what files were found anyways: %s", dir, readErr)
+	}
+
+	for _, file := range files {
+		if strings.EqualFold(file.Name(), filepath.Base(filename)) {
+			return filepath.Join(dir, file.Name()), nil
+		}
+		fmt.Println(file.Name())
```
zhan-airplane, line 32: nit: remove print?

zhan-airplane (approved):

pkg/utils/pathcase/pathcase.go
joshma: 🤦

PR #281, opened 2021-10-04
Use ExecuteTask endpoint for running tasks

francescaleung:
Accommodates internal changes to Airplane's API.

Resolves AIR-1727.


joshma (approved):
LGTM but - what if we made `triggerID` optional on `/tasks/execute`? it could default to the only form trigger (if any) and 400 otherwise (if 0 or > 1)?

francescaleung:
> LGTM but - what if we made `triggerID` optional on `/tasks/execute`? it could default to the only form trigger (if any) and 400 otherwise (if 0 or > 1)?

I like it

PR #286, opened 2021-10-14
Send PR notifications to full eng team

colinking:
This PR updates `CODEOWNERS` to cc the eng team on all PRs, so that we are all notified of PRs. I have disabled code reviewer auto-assignment for the eng team, so it will no longer do that.

To support auto-assignments for Renovate, I enabled [`reviewersFromCodeOwners `](https://docs.renovatebot.com/configuration-options/#reviewersfromcodeowners) which should auto-assign a random engineer.

Resolves AIR-2026


joshma (approved):
Does this set "assignees" and not "reviewers"? We can try it out and see if it works.

PR #300, opened 2021-10-29
Create shell scripts with +x flag

joshma:
Otherwise, you have to remember to do it yourself to run it locally.

Also creates folders with 0755 instead of 0700, allowing others to read
and execute (list), which seems to be the normal behavior of e.g. mkdir.


pkg/runtime/runtime.go
```diff
@@ -40,7 +41,7 @@ type Interface interface {
 	// Generate accepts a task and generates code to match the task.
 	//
 	// An error is returned if the code cannot be generated.
-	Generate(task api.Task) ([]byte, error)
+	Generate(task api.Task) ([]byte, os.FileMode, error)
```
colinking, line 44: Could you add a comment here about what `os.FileMode` is for / common values

colinking (approved):

PR #304, opened 2021-11-05
Implement spinner when building

leeweisberger:
Peek 2021-11-05 15-31

Write spinner to stderr when in a loading state. Stop the spinner before the next log.

---
This is a little jank - you have the make sure the loader is stopped and then restarted when writing a new log, but unsure if there is another way to do it. I updated buildLog to handle the loader correctly for you.


pkg/build/remote.go
```diff
@@ -161,6 +166,10 @@ func archiveTaskDir(def definitions.Definition, root string, archivePath string)
 }
 
 func uploadArchive(ctx context.Context, root string, client *api.Client, archivePath string) (string, error) {
+	loader := logger.NewLoader()
+	defer loader.Stop()
+	loader.Start()
```
colinking, line 171: Maybe pass `loader` down from `remote` into `uploadArchive`.  Same with `waitForBuild`. I'm surprised this doesn't render two loaders.

pkg/build/remote.go
```diff
@@ -25,15 +25,20 @@ func remote(ctx context.Context, req Request) (*Response, error) {
 	if err := confirmBuildRoot(req.Root); err != nil {
 		return nil, err
 	}
-	buildLog(api.LogLevelInfo, logger.Gray("Building with %s as root...", relpath(req.Root)))
+	loader := logger.NewLoader()
+	defer loader.Stop()
+	loader.Start()
```
colinking, line 30: Nice. Only other way that comes to mind is to make this global to `logger`, f.e. `logger.Loading(true)`, such that each `logger.Log|Debug|...` method can handle loading states directly. Has the benefit that any `logger.Log|...` call "just works".

colinking (approved):
:awyis:

pkg/build/remote.go
leeweisberger: Yeah considered doing this, but would require having once instance of logger and passing it around which seems like a real pain. If we didn't maintain one instance, n loggers would create n loaders

colinking:
@leeweisberger just noticed this doesn't work well when `--debug` is enabled. Should we just disable the spinner when debug mode is on?

leeweisberger:
> @leeweisberger just noticed this doesn't work well when `--debug` is enabled. Should we just disable the spinner when debug mode is on?

Yup have seen the same thing. I'll do that as part of the work I'm doing. I'm also introducing some concurrency and there are some corner cases there as well. Might just want a different strategy

PR #312, opened 2021-11-15
Batch deploy a list of files or a directory

leeweisberger:
Support deploying multiple files/directories. The code recursively searches through all provided files/directories and deploys all airplane tasks in them concurrently.

The code waits for all of the tasks to finish and then prints a summary of all successes/errors. If at least one task errors, the entire script errors.

Logs for each task are interwoven

---

Screenshot of what it looks like with two tasks being deployed simultaneously
image


pkg/build/remote.go
```diff
@@ -87,6 +92,21 @@ func remote(ctx context.Context, req Request) (*Response, error) {
 	}, nil
 }
 
+func (d *Deployer) getRegistryToken(ctx context.Context, client *api.Client) (registryToken api.RegistryTokenResponse, err error) {
```
leeweisberger, line 95: This ensures that we only authenticate once.

pkg/build/remote.go
```diff
@@ -21,27 +22,31 @@ import (
...
+func (d *Deployer) remote(ctx context.Context, req Request) (*Response, error) {
 	if err := confirmBuildRoot(req.Root); err != nil {
 		return nil, err
 	}
 	loader := logger.NewLoader()
 	defer loader.Stop()
 	loader.Start()
 
-	buildLog(api.LogLevelInfo, loader, logger.Gray("Building with %s as root...", relpath(req.Root)))
+	buildLog(api.LogLevelInfo, loader, req.Def.Slug, logger.Gray("Building with %s as root...", relpath(req.Root)))
```
joshma, line 38: heh minor but I think this could be a valid place to place `slug` onto the `ctx`? 😇 up to you but would make this cleaner

pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -38,14 +38,14 @@ func New(c *cli.Config) *cobra.Command {
 			airplane tasks deploy ./task.ts
 			airplane tasks deploy --local ./task.js
 			airplane tasks deploy ./my-task.yml
+			airplane tasks deploy my-directory
+			airplane tasks deploy ./my-task1.yml ./my-task2.yml
 		`),
-		Args: cobra.MaximumNArgs(1),
+		Args: cobra.MinimumNArgs(1),
```
joshma, line 44: By doing this, this removes the `-f` flag support (in which case it would have N=0 args). Is that intentional?

pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -38,14 +38,14 @@ func New(c *cli.Config) *cobra.Command {
...
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if cfg.file != "" {
-				// A file was provided with the -f flag. This is deprecated.
-				logger.Warning(`The --file/-f flag is deprecated and will be removed in a future release. File paths should be passed as a positional argument instead: airplane deploy %s`, cfg.file)
-			} else if len(args) > 0 {
+			if len(args) > 0 {
 				cfg.file = args[0]
+				cfg.filesOrDirectories = args
```
joshma, line 48: what's `cfg.file` here for? in the case of `N=1`?

pkg/cmd/tasks/deploy/script.go
```diff
@@ -15,10 +18,115 @@ import (
...
+		cmd := fmt.Sprintf("airplane exec %s", task.file)
+		logger.Suggest(
+			"⚡ To execute the task from the CLI:",
+			cmd,
+		)
+
+		logger.Suggest(
+			"⚡ To execute the task from the UI:",
+			cfg.client.TaskURL(task.Slug),
+		)
```
joshma, line 75: while you're here, can we deemphasize the CLI cmd? I want to stop suggesting `exec` locally.

```
some_task_slug:
    Deploy status: succeeded
    Visit the task: https://...
```

(I was considering adding more fields, but maybe that should go at the start? Will do a sep. comment)

pkg/cmd/tasks/deploy/script.go
```diff
@@ -15,10 +18,115 @@ import (
...
+	}
+
+	return groupErr
+}
+
+func (d *scriptDeployer) deployFromScriptHelper(ctx context.Context, cfg config, g *errgroup.Group) (rErr error) {
```
joshma, line 81: for each of the scripts we deploy, can print some summary at the beginning? I've wished we had this, something to allow a user to sanity check:
- task slug (maybe also URL)
- task type / runtime (python 3.7, node 14.x, etc.)
- root directory (and working directory if different) - helps users realize if e.g. we picked up a requirements.txt from a parent directory

dunno which of these are easy to include, but would be nice to print these out before the deploy starts

pkg/cmd/tasks/deploy/script.go
```diff
@@ -15,10 +18,115 @@ import (
 	"github.com/airplanedev/cli/pkg/taskdir/definitions"
 	"github.com/airplanedev/cli/pkg/utils/pointers"
 	"github.com/pkg/errors"
+	"golang.org/x/sync/errgroup"
 )
 
-// DeployFromScript deploys from the given script.
-func deployFromScript(ctx context.Context, cfg config) (rErr error) {
+var ignoredDirectories = map[string]bool{
```
joshma, line 24: we have airplane ignore - https://docs.airplane.dev/platform/airplane-ignore - maybe we should just honor that?

pkg/cmd/tasks/deploy/script.go
```diff
@@ -15,10 +18,115 @@ import (
...
+			for _, nestedFile := range nestedFiles {
+				if ignoredDirectories[nestedFile.Name()] {
+					continue
+				}
+				nestedFilenames = append(nestedFilenames, path.Join(fileOrDirectory, nestedFile.Name()))
+			}
+			configCopy := cfg
+			configCopy.filesOrDirectories = nestedFilenames
+			return d.deployFromScriptHelper(ctx, configCopy, g)
+		} else {
```
joshma, line 105: minor but can this use an early return instead?

pkg/cmd/tasks/deploy/script.go
```diff
@@ -15,10 +18,115 @@ import (
...
+			var nestedFilenames []string
+			for _, nestedFile := range nestedFiles {
+				if ignoredDirectories[nestedFile.Name()] {
+					continue
+				}
+				nestedFilenames = append(nestedFilenames, path.Join(fileOrDirectory, nestedFile.Name()))
+			}
+			configCopy := cfg
+			configCopy.filesOrDirectories = nestedFilenames
+			return d.deployFromScriptHelper(ctx, configCopy, g)
```
joshma, line 104: this recursion is a bit funny...you're doing a search while kicking off builds. I think this can be separated more cleanly. `deployFromScriptHelper` could be renamed to `discoverScripts` and recursively search to return a slice of files, and then in `deployFromScript` you could take that slice and `g.Go()` each item.

joshma (changes requested):

pkg/cmd/tasks/deploy/deploy.go
leeweisberger: Yeah I've removed it. Seems like it was deprecated and I didn't want to change code to support it if it is deprecated. Let me know if I should keep it around

pkg/cmd/tasks/deploy/deploy.go
leeweisberger: Removed

pkg/cmd/tasks/deploy/script.go
leeweisberger: Updated - this is what it looks like now: 
image

pkg/cmd/tasks/deploy/script.go
leeweisberger: This is what it looks like now
image

Didn't do runtime b/c that's a bit harder to get

pkg/cmd/tasks/deploy/script.go
leeweisberger: Tried doing this but reverted:
1. It's kind of hard because we don't have the information needed for our ignore call at the time we want to use it (requires reading the file, fetching type, getting root).
2. I don't think it's exactly the same concept - ignore are files that we don't want to push even if they are in a root directory. This is places where we will not discover airplane tasks.

This code isn't even necessary - it just avoids searching through folders we know won't have airplane tasks. Could see us adding configuration later where we allow deploying all tasks in a directory except tasks of a slug or in a sub directory

pkg/cmd/tasks/deploy/script.go
leeweisberger: Yup good idea. Changed to this pattern

pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -80,11 +76,11 @@ func run(ctx context.Context, cfg config) error {
 		return err
 	}
 
-	ext := filepath.Ext(cfg.file)
+	ext := filepath.Ext(cfg.filesOrDirectories[0])
```
leeweisberger, line 79: We now only support yml for single files. Not a regression but I didn't want to implement the same discovery stuff for a deprecated path

pkg/cmd/tasks/deploy/script.go
```diff
@@ -109,55 +212,118 @@ More information: https://apn.sh/jst-upgrade`)
...
+	taskRoot         string
+	workingDirectory string
+	task             api.Task
+	def              definitions.Definition
+	kind             api.TaskKind
+	kindOptions      api.KindOptions
+}
+
+// getTaskConfig a task and associated information from a script.
+func getTaskConfigFromScript(ctx context.Context, client api.Client, script script) (getTaskConfigFromScriptResp, error) {
```
leeweisberger, line 267: I don't love this, but we need this info when printing the summary at the start and when running the build. And I don't want to run through this twice

pkg/cmd/tasks/deploy/script.go
```diff
@@ -37,66 +186,20 @@ func deployFromScript(ctx context.Context, cfg config) (rErr error) {
...
-	if err != nil {
-		return err
-	}
-	tp.kind = task.Kind
+	tp.kind = taskConfig.kind
 	tp.taskID = task.ID
 	tp.taskSlug = task.Slug
 	tp.taskName = task.Name
 
-	r, err := runtime.Lookup(cfg.file, task.Kind)
```
leeweisberger, line 59: moved all of this to a helper below

pkg/cmd/tasks/deploy/deploy.go
joshma: Hrm I think it's probably fine to remove cc @colinking

pkg/cmd/tasks/deploy/script.go
```diff
@@ -124,40 +227,99 @@ More information: https://apn.sh/jst-upgrade`)
...
+		workingDirectory: wd,
+		def:              def,
+		kind:             kind,
+		kindOptions:      kindOptions,
+		task:             task,
+	}, nil
+}
+
+// Relpath returns the relative using root and the cwd.
+func relpath(root string) string {
```
leeweisberger, line 314: moved this from another file

pkg/cmd/tasks/deploy/script.go
```diff
@@ -15,10 +17,165 @@ import (
...
+			configCopy.filesOrDirectories = nestedFilenames
+			nestedScripts, err := d.discoverScripts(ctx, configCopy)
+			if err != nil {
+				return nil, err
+			}
+			scripts = append(scripts, nestedScripts...)
+			continue
+		}
+		// We found a file.
+		code, err := ioutil.ReadFile(fileOrDirectory)
```
joshma, line 158: AD - I think we can do this in a separate PR, but I'm concerned about perf + reading too much data. I think we can do (1) have a list of extensions we'll actually open, and (2) only read the first 4096 bytes or something to determine the slug.

pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -8,18 +8,17 @@ import (
...
 )
 
 type config struct {
-	root   *cli.Config
-	client *api.Client
-	file   string
-	local  bool
+	root               *cli.Config
+	client             *api.Client
+	filesOrDirectories []string
```
joshma, line 20: Can we rename this to `paths`?

pkg/cmd/tasks/deploy/script.go
joshma: update: AD, add `.git` and double check what happens. it should at least error, I suppose.

pkg/cmd/tasks/deploy/script.go
```diff
@@ -15,10 +17,165 @@ import (
...
+		return err
+	}
+	loader.Stop()
+
+	if len(scriptsToDeploy) == 0 {
+		logger.Log("No tasks to deploy")
+		return nil
+	}
+
+	var pluralize string
```
joshma, line 58: minor but I sort of like doing `noun := "task"` and then setting `noun = "tasks"` if it's plural?

pkg/cmd/tasks/deploy/script.go
```diff
@@ -124,40 +235,99 @@ More information: https://apn.sh/jst-upgrade`)
...
 		Permissions:                task.Permissions,
-		Timeout:                    def.Timeout,
+		Timeout:                    taskConfig.def.Timeout,
 		BuildID:                    pointers.String(resp.BuildID),
 		InterpolationMode:          interpolationMode,
 	})
+	return err
+}
+
+type getTaskConfigFromScriptResp struct {
```
joshma, line 261: not super important to name this since its a private struct, but...maybe `type taskConfig`? or `type taskSummary`?

joshma (approved):

pkg/build/remote.go
```diff
@@ -165,7 +188,7 @@ func archiveTaskDir(def definitions.Definition, root string, archivePath string)
 	return nil
 }
 
-func uploadArchive(ctx context.Context, root string, client *api.Client, archivePath string, loader *logger.Loader) (string, error) {
+func uploadArchive(ctx context.Context, client *api.Client, taskSlug, archivePath string, loader *logger.Loader) (string, error) {
```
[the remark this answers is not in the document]
leeweisberger: yeah it was never used. Also removed taskSlug

PR #313, opened 2021-11-15
Cache and reuse archive uploads

leeweisberger:
If two tasks use the same root (e.g. same package.json), we only want to archive and upload once and use the same upload for both builds.

---

I imagine this will be very common, so just adding in the optimization now.


joshma (approved):

joshma:
cc @colinking on this - sanity check this is OK? this means we'd use the same upload across 2 builds, which seems fine?

pkg/build/remote.go
```diff
@@ -185,7 +208,27 @@ func archiveTaskDir(def definitions.Definition, root string, archivePath string)
 	return nil
 }
 
-func uploadArchive(ctx context.Context, client *api.Client, taskSlug, archivePath string, loader *logger.Loader) (string, error) {
+func (d *Deployer) uploadArchive(ctx context.Context, client *api.Client, taskSlug, archivePath, rootPath string, loader *logger.Loader) (string, error) {
```
colinking, line 211: Would recommend using a `singleflight` here: https://pkg.go.dev/golang.org/x/sync/singleflight

colinking (approved):
Yep, I don't see any issues with re-using. Good call @leeweisberger!

pkg/build/remote.go
leeweisberger: Yes! That was the coolest package I've ever used it fit this perfectly

leeweisberger: I thought I had such a good idea :sob:

PR #316, opened 2021-11-16
Update node dockerfile to be more cachable

leeweisberger:
Convert Dockerfile to use [nodejs best practices](https://nodejs.org/en/docs/guides/nodejs-docker-webapp/#creating-a-dockerfile)

In practice, we are currently reinstalling all dependencies whenever code changes. After this pr, we will use cached dependencies unless the dependencies themselves change.


pkg/build/node.go
```diff
@@ -120,13 +120,14 @@ func node(root string, options api.KindOptions) (string, error) {
...
 		RUN npm install -g typescript@4.2 && \
 			npm install -g esbuild@0.12 --unsafe-perm
-		COPY . /airplane
 
 		RUN mkdir -p /airplane/.airplane && \
 			cd /airplane/.airplane && \
 			{{.InlineShimPackageJSON}} > package.json && \
 			npm install
 
+		COPY package*.json /airplane/
```
colinking, line 129: IIRC, we did it that way as a temporary solution to support npm/yarn workspaces. We have some WIP tests for workspaces, which would be great to get running: https://github.com/airplanedev/cli/blob/a1ff65cfdd2767a44cabaa80d3fa67ff6a179e39/pkg/build/node_test.go#L114-L124

You might be able to get away with detecting if a task uses yarn/npm workspaces and falling back on the previous behavior.

joshma: Yepp @colinking  is right. If you are in /src/foo/main, `yarn install` with workspaces looks at /src/foo/main/package.json but also /src/foo/package.json (we allow this via distinct root vs working dir). This in turn might reference other silbling packages, so you might need /src/foo/somedep/* as a result.

Maybe we can just check if root and working dir are the same. If they are the same, then I think it's safe to cache this way.

Note that longer term, I think the _right_ behavior is to skip Docker caching and explicitly save/restore NPM packages from cache.

joshma (changes requested):

pkg/build/node.go
leeweisberger: Ok good callouts - didn't even know yarn workspaces was a thing ;)

I updated to handle yarn workspaces (we only do the optimization if we're not using workspaces). I also enabled the test - had to make a small change to get it to pass, but it's better than no test - the test failed on my previous change and passes now

pkg/examples/typescript/yarnworkspaces/pkg2/src/index.ts
```diff
@@ -1,7 +1,7 @@
 // Linked to https://app.airplane.dev/t/typescript_yarnworkspaces [do not edit this line]
 
 import airplane from 'airplane'
-import {name as pkg1name} from 'pkg1'
+import {name as pkg1name} from 'pkg1/src'
```
colinking, line 4: :awyis:

pkg/build/node.go
```diff
@@ -120,23 +141,32 @@ func node(root string, options api.KindOptions) (string, error) {
...
+		COPY . /airplane
+		{{end}}
+
 		{{if .IsYarn}}
 		RUN yarn --non-interactive
 		{{else}}
 		RUN npm install
 		{{end}}
 
+		{{if not .UsesYarnWorkspaces}}
```
colinking, line 166: There are also npm workspaces: https://docs.npmjs.com/cli/v7/using-npm/workspaces

Could you add support for detecting those too? And add a test? 🙏🏻

colinking (approved):

pkg/build/node.go
leeweisberger: yes! the code as is will work for both and I'll add a test to confirm

pkg/build/node_test.go
```diff
@@ -11,109 +11,106 @@ func TestNodeBuilder(t *testing.T) {
 	ctx := context.Background()
 
 	tests := []Test{
```
joshma, line 13: what happened to these tests?

joshma (approved):
approving but removed `merge` label since the tests were commented out?

PR #317, opened 2021-11-16
Use API key from env var

leeweisberger:
Allow for an alternative method of authn - if an api key and team id are passed in via env vars, use them for auth.

I did not document this anywhere - I think it's fine to keep internal only for now.


pkg/api/client.go
```diff
@@ -407,3 +422,19 @@ func (c Client) token() (string, error) {
 	}
 	return c.Token, nil
 }
+
+// apiKey returns the configured api key or an error.
+func (c Client) apiKey() (string, error) {
+	if c.APIKey == "" {
+		return "", errors.New("api: api key is missing")
```
joshma, line 429:
```suggestion
		return "", errors.New("api: API key is missing")
```

pkg/api/client.go
```diff
@@ -407,3 +422,19 @@ func (c Client) token() (string, error) {
...
+	}
+	return c.APIKey, nil
+}
+
+// teamID returns the configured team ID or an error.
+func (c Client) teamID() (string, error) {
+	if c.TeamID == "" {
+		return "", errors.New("api: team id is missing")
```
joshma, line 437:
```suggestion
		return "", errors.New("api: team ID is missing")
```

pkg/api/client.go
```diff
@@ -350,12 +354,23 @@ func (c Client) do(ctx context.Context, method, path string, payload, reply inte
...
 	}
 
-	token, err := c.token()
-	if err != nil {
-		return err
+	token, terr := c.token()
+	apiKey, aerr := c.apiKey()
+	if terr != nil && aerr != nil {
+		return terr
+	}
```
joshma, line 361: lol these helper functions seem to do more harm than good. can't you just read `token`, and if it's non-empty, set the header? otherwise, read api key and team id, and if non-empty set the header. else, error?

joshma (approved):

pkg/api/client.go
leeweisberger: Ha agreed. I was just trying to be consistent. Simplified it

PR #321, opened 2021-11-18
Only search for slug in first N lines

leeweisberger:
Improve performance of slug search by avoiding searching large files


pkg/runtime/comment.go
```diff
@@ -24,10 +30,38 @@ func Comment(r Interface, task api.Task) string {
...
 // Ok is true if the slug was found and isn't empty.
-func Slug(code []byte) (slug string, ok bool) {
+func Slug(filePath string) (slug string, ok bool) {
+	file, err := os.Open(filePath)
+	if err != nil {
+		return
+	}
+	defer file.Close()
+
+	return slugFromScanner(bufio.NewScanner(file))
```
joshma, line 43: What does this do for non-ascii files? Or just a single really long line? I actually think reading up to N (4096?) bytes would be better.

joshma (approved):

pkg/runtime/comment.go
leeweisberger: alright updated to do this

pkg/runtime/comment.go
```diff
@@ -24,10 +28,32 @@ func Comment(r Interface, task api.Task) string {
...
 // Ok is true if the slug was found and isn't empty.
-func Slug(code []byte) (slug string, ok bool) {
+func Slug(filePath string) (slug string, ok bool) {
+	file, err := os.Open(filePath)
+	if err != nil {
+		return
+	}
+	defer file.Close()
+
+	fileInfo, err := file.Stat()
```
joshma, line 41: maybe minor but I think you can skip stat-ing? and just suppress `EOF` and/or `ErrUnexpectedEOF`. If I'm reading docs right, it should still read what bytes it can, and it'll return the number of bytes read.

leeweisberger: Will try that ^

leeweisberger: dope that worked

joshma: 👏

PR #322, opened 2021-11-18
Use code from lib

leeweisberger:
Rely on newly moved build code from the Airplane lib

---
No real changes in here. The only thing I couldn't delete is the examples/ dir because the runtime tests depend on it and don't work using the examples from lib


pkg/build/build.go
```diff
@@ -29,45 +27,9 @@ type Response struct {
...
 // Run runs the build and returns an image URL.
-func Run(ctx context.Context, deployer *Deployer, req Request) (*Response, error) {
+func Run(ctx context.Context, deployer *Deployer, req Request) (*build.Response, error) {
 	if req.Local {
 		return deployer.local(ctx, req)
 	}
 	return deployer.remote(ctx, req)
 }
-
-// applyTemplate executes template t with the provided data and
```
leeweisberger, line 39: was not being used by this file. Has been moved to lib

pkg/build/build.go
```diff
@@ -2,12 +2,10 @@ package build
```
colinking, line 2: Wonder if we should move the remaining `pkg/build` files into `cmd/tasks/deploy`?

Would avoid the package name collision (`libbuild`; "what belongs in cli/pkg/build vs. lib/pkg/build").

colinking (approved):
What do you mean by the runtime test don't work w/ the examples in `lib`? We should unify these, somehow, before merging. But otherwise, skimmed through and looks good.

leeweisberger:
> What do you mean by the runtime test don't work w/ the examples in `lib`? We should unify these, somehow, before merging. But otherwise, skimmed through and looks good.

I can work on this, but I want to get this merged so the two codebases stop drifting.

PR #323, opened 2021-11-19
Fix constraints.options being required in YAML

joshma:
This was missing a `yaml:"options,omitempty"` directive and so users of
legacy YAML files were seeing an error if they didn't specify `options`
for a constraint.

Resolves AIR-2414


pkg/cmd/tasks/deploy/yaml.go
```diff
@@ -151,7 +151,9 @@ More information: https://apn.sh/jst-upgrade`)
 			TaskID: task.ID,
 		})
 		props.buildLocal = cfg.local
-		props.buildID = resp.BuildID
+		if resp != nil {
+			props.buildID = resp.BuildID
+		}
```
leeweisberger, line 156: Instead of this, can you move the error check below up above this?

leeweisberger (changes requested):

pkg/cmd/tasks/deploy/yaml.go
joshma: `props` gets used for telemetry, so I'd like to set `buildID` if we can

leeweisberger (approved):

PR #324, opened 2021-11-22
Push docker image with cli on release

leeweisberger:
Push to our gcr registry when creating a CLI release. Push two tags, `:latest` and `v0.1.0` to `airplane-cli`.

We will use this as a base image for our github action. From the action you can just call `airplane` to run the airplane cli

---

I tested this locally, but didn't test that the pushing to gcr works. Might also need help getting the secret setup (unsure if I have access?)


Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM alpine:3.10
+RUN mkdir .airplane
+COPY ./airplane .airplane
+ENV PATH=".airplane:${PATH}"
+ENTRYPOINT ["airplane"]
```
joshma, line 5:
```suggestion
FROM alpine:3.10

RUN mkdir -p /airplane/bin
COPY ./airplane /airplane/bin/airplane
ENV PATH="/airplane/bin:${PATH}"

ENTRYPOINT ["/airplane/bin/airplane"]
```

- i think you can just use a root directory here, don't rely on current user or current directory
- ENTRYPOINT to absolute path, again don't rely on current directory

joshma (approved):

joshma:
@colinking can you help @leeweisberger w/ the GCP auth stuff?

Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM alpine:3.10
```
colinking, line 1: Would be nice to base on `debian` to make this image more DX friendly. Alternatively, we could expose a `cli-dev` image based on `debian`.

colinking:
> @colinking can you help @leeweisberger w/ the GCP auth stuff?

@leeweisberger let's chat IRL tomorrow

Dockerfile
joshma: I think this also depends on usage, if it's just to run the CLI with a mount, I feel like `alpine` is fine. I'd rather start here and switch to debian later if needed.

.goreleaser.yml
```diff
@@ -64,3 +64,7 @@ release:
     # Upload our installation scripts as release artifacts:
     - glob: ./scripts/install.sh
     - glob: ./scripts/install.ps1
+dockers:
+  - image_templates:
+    - gcr.io/airplane-public/{{.ProjectName}}-cli:latest
+    - gcr.io/airplane-public/{{.ProjectName}}-cli:{{ .Tag }}
```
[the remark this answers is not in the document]
leeweisberger: done!

PR #326, opened 2021-11-23
Write gcloud credentials to temp directory

leeweisberger:
This should fix the issue we are having - currently gcloud credentials are being written to a location that "dirties" the files. Instead, write to a temp directory so it can be used by other steps in the job but not dirty the workspace


.goreleaser.yml
```diff
@@ -68,3 +68,4 @@ dockers:
   - image_templates:
     - us-docker.pkg.dev/airplane-prod/public/cli:latest
     - us-docker.pkg.dev/airplane-prod/public/cli:{{ trimprefix .Tag "v" }}
+    - us-docker.pkg.dev/airplane-prod/public/cli:{{ .Major }}.{{ .Minor }}
```
leeweisberger, line 71: also added a new tag :)

.github/workflows/release.yml
```diff
@@ -23,6 +23,7 @@ jobs:
           project_id: airplane-prod
           service_account_key: ${{ secrets.GCP_KEY }}
           export_default_credentials: true
+          credentials_file_path: ${{ env.RUNNER_TEMP }}
```
colinking, line 26: Nice

colinking (approved):

PR #327, opened 2021-11-24
Store credentials file in a gitignored directory

leeweisberger:
Force git to ignore the location where google credentials are stored during the release pipeline - following the instructions here: https://goreleaser.com/errors/dirty/?h=dirty

---

Another try at this - I can't figure out how to test this without merging and tagging - the check that is failing only runs if there is a tag (it skips if there is no tag). I have validated that the setup gcloud cli will save credentials where I specify.


.github/workflows/release.yml
```diff
@@ -23,7 +26,7 @@ jobs:
           project_id: airplane-prod
           service_account_key: ${{ secrets.GCP_KEY }}
           export_default_credentials: true
-          credentials_file_path: ${{ env.RUNNER_TEMP }}
+          credentials_file_path: ${{ env.GOOGLE_CREDENTIALS_FILE_PATH }}
       - name: Configure Docker to use Artifact Registry
         run: gcloud auth configure-docker us-docker.pkg.dev
       - name: Run GoReleaser
```
colinking, line 32: You could remove the `tags: -...` so it runs on this branch and then add `--skip-publish` to `args: release ...` below so it doesn't perform a publish.

colinking (approved):

.github/workflows/release.yml
leeweisberger: Yeah I did that ^, but I think the goreleaser tool has something built in that doesn't run the go diff check unless there is a publish :)

PR #328, opened 2021-11-24
Add git metadata to build

leeweisberger:
Plumb git metadata from a combination of flags and reading local git config. Send to the API.

I am still allowing two flags to be passed in:
1. user - we are guessing that the user of the deploy is the same as the author of the last commit, but the user who made the last commit might not be the same person who triggered the deploy. So we allow this to be overriden by a passed in flag. In our github action, we will set this flag with the person who triggered the action.
2. repository - a single local git config might have more than one remote repository. Moreover, it doesn't really make sense to set a remote repository when somebody is deploying locally because the deploy had nothing to do with a remote repository. Our action will also pass this in.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -457,10 +457,11 @@ type CreateBuildRequest struct {
 }
 
 type BuildGitConfig struct {
-	CommitHash string `json:"commitHash"`
-	Branch     string `json:"branch"`
-	User       string `json:"gitUser"`
-	Repository string `json:"repository"`
+	CommitHash    string `json:"commitHash"`
+	Branch        string `json:"branch"`
+	User          string `json:"gitUser"`
+	Repository    string `json:"repository"`
+	CommitMessage string `json:"commitMessage"`
 }
 
 type CreateBuildResponse struct {
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -450,9 +450,17 @@ type GetBuildResponse struct {
 }
 
 type CreateBuildRequest struct {
-	TaskID         string  `json:"taskID"`
-	SourceUploadID string  `json:"sourceUploadID"`
-	Env            TaskEnv `json:"env"`
+	TaskID         string         `json:"taskID"`
+	SourceUploadID string         `json:"sourceUploadID"`
+	Env            TaskEnv        `json:"env"`
+	GitConfig      BuildGitConfig `json:"gitConfig"`
+}
+
+type BuildGitConfig struct {
+	CommitHash string `json:"commitHash"`
+	Branch     string `json:"branch"`
+	User       string `json:"gitUser"`
+	Repository string `json:"repository"`
 }
 
 type CreateBuildResponse struct {
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -450,18 +450,19 @@ type GetBuildResponse struct {
 }
 
 type CreateBuildRequest struct {
-	TaskID         string         `json:"taskID"`
-	SourceUploadID string         `json:"sourceUploadID"`
-	Env            TaskEnv        `json:"env"`
-	GitConfig      BuildGitConfig `json:"gitConfig"`
+	TaskID         string       `json:"taskID"`
+	SourceUploadID string       `json:"sourceUploadID"`
+	Env            TaskEnv      `json:"env"`
+	GitMeta        BuildGitMeta `json:"gitMeta"`
 }
 
-type BuildGitConfig struct {
+type BuildGitMeta struct {
 	CommitHash    string `json:"commitHash"`
-	Branch        string `json:"branch"`
+	Ref           string `json:"gitRef"`
 	User          string `json:"gitUser"`
 	Repository    string `json:"repository"`
 	CommitMessage string `json:"commitMessage"`
+	IsDirty       bool   `json:"isDirty"`
 }
 
 type CreateBuildResponse struct {
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -462,6 +462,7 @@ type BuildGitMeta struct {
 	User          string `json:"gitUser"`
 	Repository    string `json:"repository"`
 	CommitMessage string `json:"commitMessage"`
+	FilePath      string `json:"filePath"`
 	IsDirty       bool   `json:"isDirty"`
 }
 
--- a/pkg/cmd/tasks/deploy/deploy.go
+++ b/pkg/cmd/tasks/deploy/deploy.go
@@ -15,6 +15,12 @@ import (
 	"github.com/spf13/cobra"
 )
 
+type gitConfig struct {
+	commitHash string
+	branch     string
+	user       string
+	repository string
+}
 type config struct {
 	root   *cli.Config
 	client *api.Client
@@ -22,6 +28,7 @@ type config struct {
 	local  bool
 
 	upgradeInterpolation bool
+	git                  gitConfig
 }
 
 func New(c *cli.Config) *cobra.Command {
@@ -57,6 +64,14 @@ func New(c *cli.Config) *cobra.Command {
 
 	cmd.Flags().BoolVarP(&cfg.local, "local", "L", false, "use a local Docker daemon (instead of an Airplane-hosted builder)")
 	cmd.Flags().BoolVar(&cfg.upgradeInterpolation, "jst", false, "Upgrade interpolation to JST")
+	cmd.Flags().StringVar(&cfg.git.commitHash, "commitHash", "", "The commit hash of the source code of the deployed task")
+	cmd.Flags().StringVar(&cfg.git.branch, "branch", "", "The branch containing the source code of the deployed task")
+	cmd.Flags().StringVar(&cfg.git.repository, "repository", "", "The repository containing the source code of the deployed task")
+	cmd.Flags().StringVar(&cfg.git.user, "gitUser", "", "The git user who deployed the task")
+	cli.Must(cmd.Flags().MarkHidden("commitHash")) // internal use only
+	cli.Must(cmd.Flags().MarkHidden("branch"))     // internal use only
+	cli.Must(cmd.Flags().MarkHidden("repository")) // internal use only
+	cli.Must(cmd.Flags().MarkHidden("gitUser"))    // internal use only
 
 	return cmd
 }
--- a/pkg/cmd/tasks/deploy/deploy.go
+++ b/pkg/cmd/tasks/deploy/deploy.go
@@ -2,25 +2,30 @@ package deploy
 
 import (
 	"context"
+	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/MakeNowJust/heredoc"
 	"github.com/airplanedev/cli/pkg/api"
 	"github.com/airplanedev/cli/pkg/cli"
 	"github.com/airplanedev/cli/pkg/cmd/auth/login"
+	"github.com/airplanedev/cli/pkg/logger"
 	"github.com/airplanedev/cli/pkg/utils"
 	"github.com/airplanedev/cli/pkg/version"
 	"github.com/airplanedev/lib/pkg/build"
+	"github.com/go-git/go-git/v5"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
-type gitConfig struct {
+type gitMeta struct {
 	commitHash    string
 	commitMessage string
-	branch        string
+	ref           string
 	user          string
 	repository    string
+	isDirty       bool
 }
 type config struct {
 	root   *cli.Config
@@ -29,7 +34,7 @@ type config struct {
 	local  bool
 
 	upgradeInterpolation bool
-	git                  gitConfig
+	gitMeta              api.BuildGitMeta
 }
 
 func New(c *cli.Config) *cobra.Command {
@@ -56,6 +61,10 @@ func New(c *cli.Config) *cobra.Command {
 			} else {
 				return errors.New("expected 1 argument: airplane deploy ./path/to/file")
 			}
+			err := addGitMeta(&cfg.gitMeta)
+			if err != nil {
+				logger.Debug("failed to gather git metadata: %v", err)
+			}
 			return run(cmd.Root().Context(), cfg)
 		},
 		PersistentPreRunE: utils.WithParentPersistentPreRunE(func(cmd *cobra.Command, args []string) error {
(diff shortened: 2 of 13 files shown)


pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -57,6 +64,14 @@ func New(c *cli.Config) *cobra.Command {
...
 	cmd.Flags().BoolVarP(&cfg.local, "local", "L", false, "use a local Docker daemon (instead of an Airplane-hosted builder)")
 	cmd.Flags().BoolVar(&cfg.upgradeInterpolation, "jst", false, "Upgrade interpolation to JST")
+	cmd.Flags().StringVar(&cfg.git.commitHash, "commitHash", "", "The commit hash of the source code of the deployed task")
+	cmd.Flags().StringVar(&cfg.git.branch, "branch", "", "The branch containing the source code of the deployed task")
+	cmd.Flags().StringVar(&cfg.git.repository, "repository", "", "The repository containing the source code of the deployed task")
+	cmd.Flags().StringVar(&cfg.git.user, "gitUser", "", "The git user who deployed the task")
+	cli.Must(cmd.Flags().MarkHidden("commitHash")) // internal use only
+	cli.Must(cmd.Flags().MarkHidden("branch"))     // internal use only
+	cli.Must(cmd.Flags().MarkHidden("repository")) // internal use only
+	cli.Must(cmd.Flags().MarkHidden("gitUser"))    // internal use only
```
leeweisberger, line 74: Trying not to blow up our flag options plus wouldn't expect anyone to manually set this.

pkg/api/types.go
```diff
@@ -450,9 +450,18 @@ type GetBuildResponse struct {
...
-	TaskID         string  `json:"taskID"`
-	SourceUploadID string  `json:"sourceUploadID"`
-	Env            TaskEnv `json:"env"`
+	TaskID         string         `json:"taskID"`
+	SourceUploadID string         `json:"sourceUploadID"`
+	Env            TaskEnv        `json:"env"`
+	GitConfig      BuildGitConfig `json:"gitConfig"`
+}
+
+type BuildGitConfig struct {
```
joshma, line 459: maybe also include `Uncommitted boolean` to know if there are "dirty" changes?

pkg/api/types.go
```diff
@@ -450,9 +450,18 @@ type GetBuildResponse struct {
 }
 
 type CreateBuildRequest struct {
-	TaskID         string  `json:"taskID"`
-	SourceUploadID string  `json:"sourceUploadID"`
-	Env            TaskEnv `json:"env"`
+	TaskID         string         `json:"taskID"`
+	SourceUploadID string         `json:"sourceUploadID"`
+	Env            TaskEnv        `json:"env"`
+	GitConfig      BuildGitConfig `json:"gitConfig"`
```
joshma, line 456: "git config" reminds me of .gitconfig, maybe `GitMeta`?

pkg/api/types.go
```diff
@@ -450,9 +450,18 @@ type GetBuildResponse struct {
...
+}
+
+type BuildGitConfig struct {
+	CommitHash    string `json:"commitHash"`
+	Branch        string `json:"branch"`
```
joshma, line 461: could do `Ref` instead, which would be a branch OR tag, in case we want to build on tags in the future

joshma (commented):
NB golang calls theirs gitrevision and gituncommitted https://utcc.utoronto.ca/~cks/space/blog/programming/GoVersionOfYourSource

pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -57,6 +65,16 @@ func New(c *cli.Config) *cobra.Command {
...
+	cmd.Flags().StringVar(&cfg.git.commitHash, "commitHash", "", "The commit hash of the source code of the deployed task")
+	cmd.Flags().StringVar(&cfg.git.commitMessage, "commitMessage", "", "The commit message the deployed task was pushed with")
+	cmd.Flags().StringVar(&cfg.git.branch, "branch", "", "The branch containing the source code of the deployed task")
+	cmd.Flags().StringVar(&cfg.git.repository, "repository", "", "The repository containing the source code of the deployed task")
+	cmd.Flags().StringVar(&cfg.git.user, "gitUser", "", "The git user who deployed the task")
+	cli.Must(cmd.Flags().MarkHidden("commitHash"))    // internal use only
+	cli.Must(cmd.Flags().MarkHidden("commitMessage")) // internal use only
+	cli.Must(cmd.Flags().MarkHidden("branch"))        // internal use only
+	cli.Must(cmd.Flags().MarkHidden("repository"))    // internal use only
+	cli.Must(cmd.Flags().MarkHidden("gitUser"))       // internal use only
```
joshma, line 77: can this just read from, well, VCS? use https://github.com/go-git/go-git and figure out commit, HEAD, etc? should be the same deal if running in CI.

leeweisberger: Man you mentioned this and I forgot. I'll try this next week. I'm not sure that this will work in CI but it might? Will just have to test it.

And then if their repo is dirty might just not write the fields b/c they could have completely deleted and changed all of the code

pkg/api/types.go
leeweisberger: If I can get this reading from version control, I'm tempted to just not write git info if there are dirty changes. We can't tell how dirty the changes are - they could have completely re-written the entire task

pkg/api/types.go
leeweisberger: Yeah this is actually going to be branch or tag. I'll rename to reflect

pkg/cmd/tasks/deploy/deploy.go
joshma: > And then if their repo is dirty might just not write the fields b/c they could have completely deleted and changed all of the code

Meh IMO the dirty flag _allows_ you to store the other info, it's like an asterisk. I can see that being helpful if you have a mostly clean state but e.g. you have a new temp file that you don't want to check in.

pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -57,6 +74,10 @@ func New(c *cli.Config) *cobra.Command {
 
 	cmd.Flags().BoolVarP(&cfg.local, "local", "L", false, "use a local Docker daemon (instead of an Airplane-hosted builder)")
 	cmd.Flags().BoolVar(&cfg.upgradeInterpolation, "jst", false, "Upgrade interpolation to JST")
+	cmd.Flags().StringVar(&cfg.gitMeta.Repository, "repository", "", "The repository containing the source code of the deployed task")
+	cmd.Flags().StringVar(&cfg.gitMeta.User, "gitUser", "", "The git user who deployed the task")
```
joshma, line 78: these are flags right? should be `git-user`?

pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -57,6 +74,10 @@ func New(c *cli.Config) *cobra.Command {
 
 	cmd.Flags().BoolVarP(&cfg.local, "local", "L", false, "use a local Docker daemon (instead of an Airplane-hosted builder)")
 	cmd.Flags().BoolVar(&cfg.upgradeInterpolation, "jst", false, "Upgrade interpolation to JST")
+	cmd.Flags().StringVar(&cfg.gitMeta.Repository, "repository", "", "The repository containing the source code of the deployed task")
```
joshma, line 77: hrm so there's no way to detect this? nb in ci/cd I think we should set some env var that the CLI can pick up, feels cleaner than managing flags. do we still need the flag if that's the case?

pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -85,3 +106,47 @@ func run(ctx context.Context, cfg config) error {
...
 	return NewDeployer().deployFromScript(ctx, cfg)
 }
+
+func addGitMeta(meta *api.BuildGitMeta) error {
+	dir, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+
+	repo, err := git.PlainOpen(dir)
```
joshma, line 116: does this work if you're in a subdir?

joshma (commented):
overall looks fine to me, but let's let colin/francesca approve!

pkg/cmd/tasks/deploy/deploy.go
leeweisberger: We can "detect" it by looking at the remotes you have set up:
1. You might have multiple remotes
2. You might never have pushed your changes the the remote so it isn't even on the remote.
3. We would have to translate from remote -> url. e.g. for github, your remote can be ssh and then we have to translate from ssh -> url. Probably doable, but a bit annoying.

The combination of those 3 made me think it's easier to pass the remote in.

We can set an env var and detect it.

pkg/cmd/tasks/deploy/script.go
```diff
@@ -307,13 +319,64 @@ func getTaskConfigFromScript(ctx context.Context, client api.Client, script scri
...
 		kind:             kind,
 		kindOptions:      kindOptions,
 		task:             task,
 	}, nil
 }
 
+func getGitMetadata(taskFilePath string) (api.BuildGitMeta, error) {
+	meta := api.BuildGitMeta{}
+
+	repo, err := git.PlainOpenWithOptions(filepath.Dir(taskFilePath), &git.PlainOpenOptions{
```
colinking, line 333: Does this error if run outside of a git repo? Ideally it shouldn't.

pkg/cmd/tasks/deploy/script.go
```diff
@@ -215,7 +219,13 @@ More information: https://apn.sh/jst-upgrade`)
 		}
 	}
 
-	tp.buildLocal = cfg.local
+	gitMeta, err := getGitMetadata(tc.taskFilePath)
+	if err != nil {
+		logger.Debug("failed to gather git metadata: %v", err)
```
colinking, line 224: Are you just being defensive by not pushing this error up? If so, can you at least send it to Sentry?

pkg/cmd/tasks/deploy/script.go
```diff
@@ -215,7 +219,13 @@ More information: https://apn.sh/jst-upgrade`)
...
+	}
+	gitMeta.User = conf.GetGitUser()
```
colinking, line 226: What exactly is git user trying to track? The user who created the last commit (if not dirty) else the user who likely produced the dirty changes? Seems complicated. When would we want to use this over the user who created the deploy?

pkg/cmd/tasks/deploy/script.go
```diff
@@ -307,13 +319,64 @@ func getTaskConfigFromScript(ctx context.Context, client api.Client, script scri
...
+	}
+	meta.CommitHash = commit.Hash.String()
+	meta.CommitMessage = commit.Message
+	if meta.User != "" {
+		meta.User = commit.Author.Name
+	}
+
+	ref := h.Name().String()
+	if h.Name().IsBranch() {
+		ref = strings.TrimPrefix(ref, "refs/heads/")
```
colinking, line 373: So if you checked out a tag and ran this, we'll get `refs/tags/v1.55.0` (or whatever the tag is). Maybe we leave in `refs/...` and let downstream folks parse it?

pkg/cmd/tasks/deploy/script.go
```diff
@@ -307,13 +319,64 @@ func getTaskConfigFromScript(ctx context.Context, client api.Client, script scri
...
+	if err != nil {
+		return meta, err
+	}
+
+	commit, err := repo.CommitObject(h.Hash())
+	if err != nil {
+		return meta, err
+	}
+	meta.CommitHash = commit.Hash.String()
+	meta.CommitMessage = commit.Message
```
colinking, line 366: Wdyt of just storing the title (iow, the first line)?

colinking (approved):

pkg/api/types.go
```diff
@@ -450,9 +450,20 @@ type GetBuildResponse struct {
...
+}
+
+type BuildGitMeta struct {
+	CommitHash    string `json:"commitHash"`
+	Ref           string `json:"gitRef"`
+	User          string `json:"gitUser"`
+	Repository    string `json:"repository"`
+	CommitMessage string `json:"commitMessage"`
+	FilePath      string `json:"filePath"`
+	IsDirty       bool   `json:"isDirty"`
```
colinking, line 466: Does this give us enough information to tie a build back to a specific PR? Or do we need additional GitHub/GitLab/etc. specific fields?

pkg/cmd/tasks/deploy/script.go
joshma: hmmm "downstream" would be ourselves typically, i guess it depends on how precise we want to be here. I feel usually it's just to produce a human-friendly label, a lot of CI will just be the branch or tag name, maybe we can strip off the ref?

pkg/cmd/tasks/deploy/script.go
leeweisberger: Updated to ignore that error.

pkg/cmd/tasks/deploy/script.go
leeweisberger: Yeah just being defensive. Added a sentry log

pkg/cmd/tasks/deploy/script.go
leeweisberger: The only thing this is tracking is the git user who created the deploy.

If you deploy directly from the CLI, this will be empty. An Airplane user created the deploy, not a git user.

If you deploy from CI and include the correct env var (our gh action does this and it's documented), it will be the git user who triggered the deploy.

pkg/cmd/tasks/deploy/script.go
leeweisberger: I think I'd prefer to send the whole thing and let the backend decide what to store and what to display rather than dropping the rest of the commit. e.g. maybe we store the whole thing but just display the title to the user? Maybe just the title in a list view and then whoel thing in a detail? fwiw vercel stores the entire thing

pkg/api/types.go
leeweisberger: It doesn't tie to a pr, but it ties to the commit that contains the code change that led to the deploy.

We are recommending that a new deploy gets built on a push event. e.g. every time a commit is pushed or a branch is merged (also triggers a push) run a deploy. The same fields make sense for a pull request event, but we might have to pull the data from CI and pass it in if we want to support that fully. Pull request info doesn't exist in local git config as far as I understand

pkg/cmd/tasks/deploy/script.go
colinking: It could be nice if we can differentiate between a tag and branch (f.e. different iconography, possibly different URLs, etc.).

PR #330, opened 2021-11-29
Loader fixes

leeweisberger:
1. Support not showing the loader (will use this to hide the loader in debug mode)
2. Loader handles concurrency
3. Loader does not show in non terminals like the log output of github actions

---

Closing https://github.com/airplanedev/lib/pull/9 in favor of this.

At some point we'll get to a better solution where the loader is built into the logger and the logger is not a singleton - don't want to invest in those changes now


colinking (approved):

colinking:
Is the `logger` the only package duplicated between `cli` and `build`? How much effort would it be to remove it from `build`?

PR #332, opened 2021-11-30
Add Definition_0_3 with JSON schema

francescaleung:
Adds the Go struct with JSON tags + JSON schema for the new task definition. Does not include unmarshalling, validation, or attempts at trying to fit old task definitions in. This also does not update the canonical Definition that we're using everywhere else, which I'm sure is going to cause problems later.

No YAML tags, because I've read the rationale behind https://github.com/kubernetes-sigs/yaml & I'm convinced it's the right way to go.

Resolves AIR-2456, AIR-2457.


joshma:
> No YAML tags, because I've read the rationale behind https://github.com/kubernetes-sigs/yaml & I'm convinced it's the right way to go.

I regret the YAML validator we use right now—can we switch to a jsonschema validator once that's ready?

joshma:
lol nevermind I just read the PR 😂

pkg/taskdir/definitions/schema_0_3.json
```diff
@@ -0,0 +1,353 @@
...
+	  "type": "string",
+	  "pattern": "^[a-z0-9_]+$",
+	  "maxLength": 50
+	},
+    "baseDefinition": {
+      "type": "object",
+      "properties": {
+        "name": { "type": "string" },
+        "slug": { "$ref": "#/$defs/slug" },
+        "folder": { "type": "string" },
```
leeweisberger, line 297: Didn't we decide we were going to leave this out to start?

pkg/taskdir/definitions/schema_0_3.json
```diff
@@ -0,0 +1,353 @@
+{
```
leeweisberger, line 1: None of these have titles / descriptions - fine for now, but we're going to want to add to provide documentation

https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.9.1

leeweisberger: Same with examples
https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.9.5

leeweisberger (approved):

pkg/taskdir/definitions/schema_0_3.json
francescaleung: right, my bad

pkg/taskdir/definitions/schema_0_3.json
francescaleung: yep, those should get added, will file a ticket in linear.

PR #333, opened 2021-12-01
Add launch darkly to CLI

leeweisberger:
Add launch darkly client and flagger interface so we can use feature flags in the CLI.


.github/workflows/release.yml
```diff
@@ -38,3 +38,4 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GORELEASER_GITHUB_TOKEN }}
           SEGMENT_WRITE_KEY: ${{ secrets.SEGMENT_WRITE_KEY }}
           SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
+          LAUNCHDARKLY_SDK_KEY: ${{ secrets.LAUNCHDARKLY_SDK_KEY }}
```
colinking, line 41: FYI, we should not use an SDK key here: https://docs.launchdarkly.com/sdk/concepts/client-side-server-side#sdk-key

It's considered confidential and could be extracted from the CLI binary. Looks like a mobile SDK is what we want here: https://docs.launchdarkly.com/sdk/concepts/client-side-server-side#mobile-key

pkg/cmd/root/root.go
```diff
@@ -53,6 +61,14 @@ func New() *cobra.Command {
 				logger.Debug("error in analytics.Init: %v", err)
 			}
 
+			// LaunchDarkly
+			var err error
+			ldf, err = launchdarkly.NewClient(cfg)
+			if err != nil {
+				logger.Debug("error in launchdarkly.NewClient: %v", err)
```
colinking, line 68: Could you report to Sentry, too?

pkg/flags/launchdarkly/launchdarkly.go
```diff
@@ -0,0 +1,99 @@
...
+	// inject it. We do not offer a pre-initialized singleton instance because
+	// doing so would make it difficult to toggle flags during tests.
+	if !atomic.CompareAndSwapInt32(&initialized, 0, 1) {
+		return nil, errors.New("unable to initialize LaunchDarkly: NewClient called more than once")
+	}
+
+	sdkKey := launchdarklySDKKey
+	if sdkKey == "" {
+		// Fall back on env var.
+		sdkKey = viper.GetString("launchdarkly_sdk_key")
```
colinking, line 49: Could you make this a field on `NewClient`? f.e. `fallbackSDKKey` I generally prefer that we don't use globals like `viper`.

pkg/flags/launchdarkly/launchdarkly.go
```diff
@@ -0,0 +1,99 @@
+package launchdarkly
```
colinking, line 1: Wonder if we should extract this into `airplanedev/lib` and share it with `airport`? We'd just need to generalize the user configuration logic and support both sdk and mobile keys.

colinking (approved):

leeweisberger:
launch darkly doesn't support clientside mode in go. We would have to plumb this through airport via API

PR #334, opened 2021-12-01
Add rough `init` pass for brand new tasks

francescaleung:
This is a little rough, but it fleshes out `airplane init --dev` for new tasks. Not all runtimes exist, so some task types aren't fully fleshed out yet, but Node/Python/shell tasks should work perfectly fine. It also handles having a file specified (entrypoint or task defn) as well as switching on `--def-format`.

Some rougher points that I'm aware of:
* YAML generation has not been implemented; only JSON generation is currently supported.
* Does not handle existing tasks yet. There's another ticket for that.
* There's an explicit check for REST & Image task types, so we don't ask for entrypoints. I'm thinking of bundling it into the `runtime.Interface`; at some point it stops being runtime-specific & is more like... task kind specifics. This could also just ask for other kind-specific things (e.g., we require a node version for node tasks, a method/path for REST tasks, etc.).

Resolves AIR-2459.


pkg/cmd/tasks/initcmd/init.go
```diff
@@ -125,13 +151,73 @@ func initWithTaskDef(ctx context.Context, cfg config) error {
...
+	if cfg.slug != "" {
+		task, err := client.GetTask(ctx, cfg.slug)
+		if err != nil {
+			return err
+		}
+
+		name = task.Name
+		kind = task.Kind
+		slug = task.Slug
+		// TODO: handle this properly
```
leeweisberger, line 168: Can you expand on what properly means in the comment? In case you aren't working on this or something.

pkg/cmd/tasks/initcmd/init.go
```diff
@@ -125,13 +151,73 @@ func initWithTaskDef(ctx context.Context, cfg config) error {
...
+		entrypoint = task.KindOptions["entrypoint"].(string)
+	} else {
+		name = cfg.newTaskInfo.name
+		kind = cfg.newTaskInfo.kind
+		entrypoint = cfg.newTaskInfo.entrypoint
```
leeweisberger, line 173: should we check that this info exists?

pkg/cmd/tasks/initcmd/init.go
```diff
@@ -125,13 +151,73 @@ func initWithTaskDef(ctx context.Context, cfg config) error {
...
+		slug = utils.MakeSlug(name)
+	}
+
+	// Create entrypoint, without comment link.
```
leeweisberger, line 177: Can yiou move this comment to where you are creating the entrypoint? or update it to reflect that we check to see if there is an entrypoint and create it if does not exist

pkg/cmd/tasks/initcmd/init.go
```diff
@@ -125,13 +151,73 @@ func initWithTaskDef(ctx context.Context, cfg config) error {
...
+		if err := createEntrypoint(r, entrypoint, nil); err != nil {
+			return errors.Wrapf(err, "unable to create entrypoint")
+		}
+		logger.Step("Created %s", entrypoint)
+	} else {
+		logger.Step("%s already exists", entrypoint)
+	}
+
+	// Create task defn file.
+	// TODO: check for existence of this file before overwriting it.
```
leeweisberger, line 192: Can you just resolve this? Should be like one line right?

pkg/cmd/tasks/initcmd/init.go
```diff
@@ -125,13 +151,73 @@ func initWithTaskDef(ctx context.Context, cfg config) error {
...
+	defFn := slug + ".task.yaml"
+	if cfg.defFormat == "json" {
+		defFn = slug + ".task.json"
+	}
```
leeweisberger, line 196: Can we do something like

ext := cfg.defFormat
defFn := fmt.Sprintf(%s.task.%s,slug, ext)

pkg/cmd/tasks/initcmd/init.go
```diff
@@ -282,6 +358,111 @@ func promptForNewFileName(task api.Task) (string, error) {
...
+			Default: base,
+		},
+		&info.name,
+	); err != nil {
+		return err
+	}
+
+	// Ask for a kind.
+	kindNames := []string{}
+	for _, kind := range orderedKinds {
```
leeweisberger, line 408: can you just store the orderedKindNames? Why do this conversion?

pkg/cmd/tasks/initcmd/init.go
```diff
@@ -282,6 +358,111 @@ func promptForNewFileName(task api.Task) (string, error) {
...
+			Message: "What kind of task should this be?",
+			Options: kindNames,
+			Default: defaultKind,
+		},
+		&selectedKindName,
+	); err != nil {
+		return err
+	}
+	for kind, name := range namesByKind {
+		if name == selectedKindName {
```
leeweisberger, line 432: what if we don't find it? Can that happen?

pkg/cmd/tasks/initcmd/init.go
```diff
@@ -282,6 +358,111 @@ func promptForNewFileName(task api.Task) (string, error) {
...
+			info.kind = kind
+			break
+		}
+	}
+
+	// Ask for an entrypoint, maybe.
+	if info.kind != build.TaskKindREST && info.kind != build.TaskKindImage {
```
leeweisberger, line 439: make a constant set of kindsWithoutEntrypoints

pkg/cmd/tasks/initcmd/init.go
```diff
@@ -282,6 +358,111 @@ func promptForNewFileName(task api.Task) (string, error) {
...
+		if file != "" && !definitions.IsTaskDef(file) {
+			info.entrypoint = file
+		} else {
+			fileName := utils.MakeSlug(info.name) + runtime.SuggestExt(info.kind)
+			if cwdIsHome, err := cwdIsHome(); err != nil {
+				return err
+			} else if cwdIsHome {
+				// Suggest a subdirectory to avoid putting a file directly into home directory.
```
leeweisberger, line 447: mmm ok but why not just leave it up to them?

pkg/cmd/tasks/initcmd/init.go
```diff
@@ -282,6 +358,111 @@ func promptForNewFileName(task api.Task) (string, error) {
...
+				fileName = filepath.Join("airplane", fileName)
+			}
+
+			if err := survey.AskOne(
+				&survey.Input{
+					Message: "Where should the script be created?",
```
leeweisberger, line 453: Unsure, just asking, do we use `script` to describe this?

pkg/cmd/tasks/initcmd/init.go
```diff
@@ -293,3 +474,28 @@ func cwdIsHome() (bool, error) {
 	}
 	return cwd == home, nil
 }
+
+func createEntrypoint(r runtime.Interface, entrypoint string, task *api.Task) error {
+	var code []byte
+	var fileMode os.FileMode
+	var err error
+
+	if task != nil {
```
leeweisberger, line 483: A bit confused why this is what you are switching on. Can you explain?

pkg/runtime/javascript/javascript.go
```diff
@@ -55,10 +57,17 @@ type Runtime struct{}
 
 // Generate implementation.
 func (r Runtime) Generate(t api.Task) ([]byte, fs.FileMode, error) {
-	var args = data{Comment: runtime.Comment(r, t)}
+	return r.generate(data{Comment: runtime.Comment(r, t)})
+}
+
+func (r Runtime) GenerateWithoutComment() ([]byte, fs.FileMode, error) {
```
leeweisberger, line 63: we could also just make this a param on Generate instead of a new function? Like options

pkg/taskdir/definitions/def_0_3.go
```diff
@@ -5,6 +5,9 @@ import (
 	"encoding/json"
 
 	"github.com/airplanedev/cli/pkg/api"
+	"github.com/airplanedev/lib/pkg/build"
+	"github.com/pkg/errors"
+	"sigs.k8s.io/yaml"
```
leeweisberger, line 10: why use this lib if it is unordered? Isn't this the standard lib: https://github.com/go-yaml/yaml

leeweisberger (approved):

pkg/runtime/javascript/javascript.go
colinking: +1 -- essentially a toggle between "lite" vs "full" format

pkg/runtime/typescript/typescript.go
```diff
@@ -57,18 +59,26 @@ type Runtime struct {
...
+		d.Params = append(d.Params, param{
 			Name: p.Slug,
 			Type: typeof(p.Type),
 		})
 	}
+	return r.generate(d)
+}
+
+func (r Runtime) GenerateWithoutComment() ([]byte, fs.FileMode, error) {
+	return r.generate(data{})
```
colinking, line 75: This is missing parameter types which are still in `Generate`

pkg/taskdir/definitions/def_0_3.go
colinking: +1 since we're already using `go-yaml`: https://github.com/airplanedev/cli/blob/073c9c7fe9089e30098899d482503faebc3f8704/pkg/taskdir/definitions/validate.go#L7

pkg/cmd/tasks/initcmd/init.go
francescaleung: This is AIR-2460, I'll probably remember it :)

pkg/cmd/tasks/initcmd/init.go
francescaleung: only if `orderedKindNames` and `kindNames` aren't in sync. I'll add a check anyway.

pkg/cmd/tasks/initcmd/init.go
francescaleung: I want to build this into `Runtime.Interface`, punted for now.

pkg/cmd/tasks/initcmd/init.go
francescaleung: This mirrors what we currently do when you run `airplane init` for a new file (existing task).

pkg/cmd/tasks/initcmd/init.go
francescaleung: This is the same question we ask right now for `airplane init`.

pkg/cmd/tasks/initcmd/init.go
francescaleung: combining `Generate` & `GenerateWithoutComment` fixes this

pkg/runtime/typescript/typescript.go
francescaleung: Unless we ask for parameters in the CLI wizard as well, it's gonna have to be empty.

pkg/taskdir/definitions/def_0_3.go
francescaleung: `go-yaml` requires `yaml` tags & does unmarshalling slightly different than `json`. See the [original blog post](https://web.archive.org/web/20190603050330/http://ghodss.com/2014/the-right-way-to-handle-yaml-in-golang/).

Ordering may be a deal breaker, though.

francescaleung: Removed this for now, so YAML generation is disabled for now.

PR #339, opened 2021-12-06
Add YAML support to Definition_0_3

francescaleung:
Uses `github.com/goccy/go-yaml` to convert between JSON & YAML, preserving ordering of keys. This means we only need to define JSON tags & JSON marshalling/unmarshalling functions, instead of also supporting separate YAML tags & YAML marshalling/unmarshalling functions.

Some small tests for good measure.

Resolves AIR-2533.


pkg/taskdir/definitions/def_0_3_test.go
```diff
@@ -0,0 +1,104 @@
...
+		},
+	},
+	Python: &PythonDefinition_0_3{
+		Entrypoint: "hello_world.py",
+		Arguments:  []string{"{{JSON.stringify(params)}}"},
+	},
+	Timeout: 3600,
+}
+
+func TestDefinition_0_3(t *testing.T) {
```
leeweisberger, line 72: In the future, would recommend using table driven tests for this kind of testing: https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

Can be generated in vscode by typing `tdt` and selecting the autocomplete option

francescaleung: Will do when there's more to test here (i.e., non-zero defaults). The bodies for these are all kinds of different, so I'll keep it as is for now.

PR #340, opened 2021-12-06
Handle deploying from a task definition file

francescaleung:
I punted on handling explicit permissions and new tasks, will address in later PRs. There's some thought towards handling multiple task definition files, but it's not fully fleshed out. This patch works for both local + remote builds.

I don't know how well this works for complicated Root determination, & what exactly should happen if the task defn file is not in the working directory & the entrypoint is not in the same directory as the task defn file. I suspect this doesn't handle that well & I'm not sure I understand the concept of the Root well enough to figure out exactly what should happen here. This works for the case where the working directory is where the task defn file + entrypoint are. Filed a ticket for examining Root complications in more detail.

Resolves AIR-2462, AIR-2546, AIR-2463.


pkg/build/remote.go
```diff
@@ -45,7 +45,11 @@ func NewDeployer() *Deployer {
 }
 
 func (d *Deployer) remote(ctx context.Context, req Request) (*libBuild.Response, error) {
-	ctx = context.WithValue(ctx, taskSlugContextKey, req.Def.Slug)
+	if req.Def_0_3 != nil {
```
leeweisberger, line 48: We're going to delete the whole def v2 path right? So all of these if/elses are going to go away?

leeweisberger: Might want to use `def` for the new version and then `def_deprecated` or `dev_0_2` for the old? That way when we clean this up, we just delete all of the things dealing with the deprecated def and we don't have to rename anything. Up to you

pkg/taskdir/directory.go
```diff
@@ -34,7 +34,7 @@ func New(file string) (TaskDirectory, error) {
 // Open creates a TaskDirectory struct from a file argument
 // Supports file in the form of github.com/path/to/repo/example and will download from GitHub
 // Supports file in the form of local_file.yml and will read it to determine the full details
-func Open(file string) (TaskDirectory, error) {
+func Open(file string, use_0_3 ...bool) (TaskDirectory, error) {
```
leeweisberger, line 37: What's up with the multiple bools? What's going on here?

pkg/taskdir/definitions/definitions.go
```diff
@@ -257,17 +257,21 @@ func (def Definition) Validate() (Definition, error) {
...
 }
 
-var jsonRegex = regexp.MustCompile(`{{ *JSON *}}`)
-
 // Upgrades this task definition for JST interpolation.
 // Assumes only usage of expressions is {{JSON}}.
 func (def *Definition) UpgradeJST() {
-	args := []string{}
-	for _, arg := range def.Arguments {
+	def.Arguments = upgradeArguments(def.Arguments)
```
leeweisberger, line 263: what's the point of this refactor?

pkg/taskdir/definitions/def_0_3.go
```diff
@@ -252,3 +259,274 @@ func (d *Definition_0_3) Unmarshal(format TaskDefFormat, buf []byte) error {
...
+		Name:        d.Name,
+		Description: d.Description,
+		Timeout:     d.Timeout,
+	}
+
+	if image != nil {
+		req.Image = image
+	}
+
+	// Convert parameters.
```
leeweisberger, line 299: Just for readability, can you break this into a helper function

pkg/taskdir/definitions/def_0_3.go
```diff
@@ -252,3 +259,274 @@ func (d *Definition_0_3) Unmarshal(format TaskDefFormat, buf []byte) error {
...
+		req.RequireExplicitPermissions = true
+		// TODO: convert permissions.
+	}
+
+	if d.Constraints != nil {
+		req.Constraints = *d.Constraints
+	}
+
+	resourcesByName := map[string]api.Resource{}
+	if d.SQL != nil || d.REST != nil {
```
leeweisberger, line 349: break this into a constant map - types that have resources

pkg/taskdir/definitions/def_0_3.go
```diff
@@ -252,3 +259,274 @@ func (d *Definition_0_3) Unmarshal(format TaskDefFormat, buf []byte) error {
...
+		resp, err := client.ListResources(ctx)
+		if err != nil {
+			return api.UpdateTaskRequest{}, errors.Wrap(err, "fetching resources")
+		}
+		for _, resource := range resp.Resources {
+			resourcesByName[resource.Name] = resource
+		}
+	}
+
+	// Convert kind-specific things.
```
leeweisberger, line 360: No way around this switch statement? In the old code it looks like we used mapstructure.Decode to do this automatically. 

How many steps do you have to take to add a new field to a python task e.g.? If it obvious where you have to add these things? Where are we going to mess up in the future? How about adding a new task type?

pkg/cmd/tasks/deploy/defn.go
```diff
@@ -0,0 +1,190 @@
...
+	"github.com/airplanedev/cli/pkg/logger"
+	"github.com/airplanedev/cli/pkg/taskdir"
+	"github.com/airplanedev/cli/pkg/taskdir/definitions"
+	"github.com/airplanedev/cli/pkg/utils/pointers"
+	libBuild "github.com/airplanedev/lib/pkg/build"
+	"github.com/pkg/errors"
+)
+
+// deployFromTaskDefn deploys from a task definition file.
+func deployFromTaskDefn(ctx context.Context, cfg config) error {
```
leeweisberger, line 21: We still need to do some sort of discovery step to discover all task def tasks in a package.

pkg/cmd/tasks/deploy/defn.go
```diff
@@ -0,0 +1,190 @@
...
+		return errors.Wrap(err, "getting task")
+	}
+
+	tc, err := getTaskConfigFromDefn(ctx, *cfg.client, def, task, dir.DefinitionRootPath())
+	if err != nil {
+		return err
+	}
+
+	if err := deploySingleTaskFromTaskDefn(ctx, cfg, tc); err != nil {
+		logger.Log("\n" + logger.Bold(tc.def_0_3.Slug))
```
leeweisberger, line 47: All these logs should really be in one place. 

We might want to change around some architecture here.

This is how I think it should go:
Rather than having N ways to deploy a task, we should have N ways to discover a task. We have one master discover function that goes through all files and discovers defn tasks, yaml tasks, and script tasks. Then when a task is discovered, it is transformed into a single abstraction (task config). Then we have one code path that deploys a task config.

leeweisberger (changes requested):
One main comment - we really want this to flow:
- Walk through filesystem and discover tasks (we can discover script tasks, yaml tasks, defnv3 tasks).
- Transform the task to a single abstraction (task config) from the given task type
- Once we have N task configs, deploy them using a single codebase.

This is basically the visitor pattern (https://refactoring.guru/design-patterns/visitor) where each type of task config has the ability to recognize a task and convert it to a task config

pkg/build/remote.go
francescaleung: Yes--we should be upgrading Definition_0_2 structs to Definition_0_3 structs. There's a ticket filed for this.

pkg/taskdir/definitions/def_0_3.go
francescaleung: Optionally: maybe this is a `lib/build` or `runtime` concern, does this type need resources? what resources does this type need?

pkg/taskdir/directory.go
francescaleung: default false. I'll just change it to explicit true/false, no defaults.

pkg/taskdir/definitions/definitions.go
francescaleung: Definition_0_3 (which is _not_ aliased to Definition) also has an UpgradeJST method.

pkg/taskdir/definitions/def_0_3.go
francescaleung: Old code is using a structure a lot more similar to how we keep things in the database, which is why we can just use mapstructure.Decode.

The part that's missing here is taking an api.Task & deriving a Definition from it. For adding a new field to a kind-specific thing: add it in the schema.json, add it in the Definition -> UpdateTaskRequest method, add it in the (currently nonexistent) api.Task -> Definition method. Adding a new task type means all of that, plus also adding it to main Kind switch & all the other switches on kind (currently: Entrypoint, Root, UpgradeJST).

Instead of switching on Kind, we could have a TaskKind interface instead. Then for new task types, we'd need to add it to the schema.json, implement the TaskKind interface, add it to the main Kind / TaskKind switches. Does that sound better?

pkg/cmd/tasks/deploy/defn.go
francescaleung: Agree--as discussed, punting on this for now.

pkg/cmd/tasks/deploy/defn.go
francescaleung: Agree--as discussed, punting on this for now.

pkg/build/remote.go
francescaleung: The other possibility here for cleaner user code is a DefinitionInterface that abstracts away the underlying def implementation. I think this is probably a better way to go.

pkg/taskdir/definitions/def_0_3.go
```diff
@@ -34,13 +37,52 @@ type Definition_0_3 struct {
 	Timeout int `json:"timeout,omitempty"`
 }
 
+type taskKind_0_3 interface {
+	updateTaskRequest(context.Context, *api.Client, *api.UpdateTaskRequest) error
```
leeweisberger, line 41: I don't love this name - it sounds like this is actively updating something while it's really just hydrating. Maybe `hydrateTaskRequest`?

pkg/taskdir/definitions/def_0_3.go
```diff
@@ -252,3 +564,244 @@ func (d *Definition_0_3) Unmarshal(format TaskDefFormat, buf []byte) error {
...
+	} else if d.SQL != nil {
+		return d.SQL, nil
+	} else if d.REST != nil {
+		return d.REST, nil
+	} else {
+		return nil, errors.New("incomplete task definition")
+	}
+}
+
+func (d Definition_0_3) UpdateTaskRequest(ctx context.Context, client *api.Client, image *string) (api.UpdateTaskRequest, error) {
```
leeweisberger, line 616: This isn't updating anything right? Make this `getUpdateTaskRequest`?

pkg/taskdir/definitions/def_0_3.go
```diff
@@ -252,3 +564,244 @@ func (d *Definition_0_3) Unmarshal(format TaskDefFormat, buf []byte) error {
...
+	}
+
+	if err := d.convertKindSpecifics(ctx, client, &req); err != nil {
+		return api.UpdateTaskRequest{}, err
+	}
+
+	return req, nil
+}
+
+func (d Definition_0_3) convertParameters(ctx context.Context, client *api.Client, req *api.UpdateTaskRequest) error {
```
leeweisberger, line 647: this func is strangely named to. `convert` makes me think it should return something, but it returns nothing. It's not obvious that it's updating the req. Maybe `addParamsToReq` or `hydrateReqWithParams`

pkg/taskdir/definitions/def_0_3.go
```diff
@@ -252,3 +564,244 @@ func (d *Definition_0_3) Unmarshal(format TaskDefFormat, buf []byte) error {
...
+				param.Constraints.Options[j].Value = od.Value
+			}
+		}
+
+		req.Parameters[i] = param
+	}
+	return nil
+}
+
+func (d Definition_0_3) convertPermissions(ctx context.Context, client *api.Client, req *api.UpdateTaskRequest) error {
```
leeweisberger, line 690: same here. Not really converting? and below

leeweisberger (approved):
A few naming issues. otherwise lgtm

PR #341, opened 2021-12-07
Fix `python dev` to check for python if python3 is not found

joshma:
On Windows (e.g. Ancaonda and the normal Python installer), python3.exe
is not available and you have to run python.exe. We therefore attempt to
use python3, but if python3 isn't available and python is, we use that
instead. We still check that `python --version` returns `Python 3`.


joshma:
Was able to test and verify on a windows VM!

leeweisberger (approved):

PR #342, opened 2021-12-07
Supress errors from checking for versions and hide prerelease tags

joshma:
Two unrelated changes: (1) filter out draft and prerelease tags, (2) if
the version check fails, log a debug message but otherwise supporess the
message.

This way, our optimistic version warnings don't block workflows if they
break.

The root cause might be that GH API is 500-ing and we're not handling
those correctly. I'll take a pass at that later.


colinking (approved):
Could you add a comment [here](https://github.com/airplanedev/cli/blob/main/.github/CONTRIBUTING.md#releases) explaining that pre-releases must be marked as pre-releases in the GitHub Releases UI?

pkg/version/version.go
```diff
@@ -25,14 +25,17 @@ func Date() string {
...
+	Draft      bool   `json:"draft"`
+	Prerelease bool   `json:"prerelease"`
 }
 
 // CheckLatest queries the GitHub API for newer releases and prints a warning if the CLI is outdated.
 func CheckLatest(ctx context.Context) error {
 	latest, err := getLatest(ctx)
 	if err != nil {
-		return err
+		logger.Debug("An error ocurred checking for the latest version: %s", err)
```
leeweisberger, line 37: can we also send this to our logging service so we can track how often it's happening?

leeweisberger (approved):

joshma:
FYI @colinking I made this a lot more aggressive, see the last few commits. had to pull out version/latest to `analytics.ReportError` without a circular dep, and made `GetLatest()` not return an error since all the returns were `nil`.

PR #343, opened 2021-12-07
Use runtime from lib

leeweisberger:
Use runtime from the `lib` library.

Remove runtime and examples from CLI


pkg/cmd/tasks/initcmd/init.go
```diff
@@ -269,7 +269,7 @@ func initCodeOnly(ctx context.Context, cfg config) error {
 		if err != nil {
 			return err
 		}
-		code := prependComment(buf, runtime.Comment(r, task))
+		code := prependComment(buf, runtime.Comment(r, task.URL))
```
leeweisberger, line 272: This was one change I made - `Comment` takes a task url instead of the whole task

pkg/cmd/tasks/initcmd/init.go
```diff
@@ -483,7 +483,7 @@ func cwdIsHome() (bool, error) {
 }
 
 func createEntrypoint(r runtime.Interface, entrypoint string, task *api.Task) error {
-	code, fileMode, err := r.Generate(task)
+	code, fileMode, err := r.Generate(apiTaskToRuntimeTask(task))
```
leeweisberger, line 486: The other change - `Generate` takes a slimmed down version of task with only the info the runtime needs

pkg/logger/logger.go
```diff
@@ -16,6 +16,39 @@ var (
 	EnableDebug bool
 )
 
+type Logger interface {
```
leeweisberger, line 19: `PrepareRun` in `lib` takes a logger interface so that we can log from within the `lib`

PR #345, opened 2021-12-09
Discover & bulk deploy tasks regardless of type.

leeweisberger:
When a user runs airplane deploy, recursively search all files for either script (lite) tasks or yaml tasks. Convert found airplane tasks into task configs. Batch deploy task configs.

---

There's a lot of code here - definitely some refactoring mixed in. Most of the refactoring supports the goal of condensing discover / deploy code and other is to support testing (we really have to start creating interfaces!!!).

Might be useful to review in person @francescaleung. I've heavily commented changes. Hopefully easy to follow!


pkg/api/client.go
```diff
@@ -74,6 +74,23 @@ type Client struct {
 	TeamID string
 }
 
+type APIClient interface {
```
leeweisberger, line 77: Allow client to be mocked in tests

pkg/build/build.go
```diff
@@ -8,10 +8,13 @@ import (
 	"github.com/airplanedev/lib/pkg/build"
 )
 
+type BuildCreator interface {
```
leeweisberger, line 11: Interface so the remote / local build creator's (much better name than deployer!) can be mocked out

pkg/build/build.go
```diff
@@ -26,11 +29,3 @@ type Response struct {
 	// Optional, only if applicable
 	BuildID string
 }
-
-// Run runs the build and returns an image URL.
-func Run(ctx context.Context, deployer *Deployer, req Request) (*build.Response, error) {
```
leeweisberger, line 31: This was a dumb way of doing this. Now the caller creates a build creator (either local or remote) and just calls `CreateBuild`

pkg/build/local.go
```diff
@@ -10,8 +10,19 @@ import (
 	"github.com/pkg/errors"
 )
 
-func (d *Deployer) local(ctx context.Context, req Request) (*build.Response, error) {
+type localBuildCreator struct {
+	registryTokenGetter
```
leeweisberger, line 14: This allows the local/remote build creators to share token getting code

pkg/cmd/tasks/deploy/defn.go
```diff
@@ -1,190 +0,0 @@
-package deploy
```
leeweisberger, line 1: this code has been moved into
* pkg/cmd/tasks/deploy/discover/defn_discover.go (discovering tasks and converting to task_configs)
* pkg/cmd/tasks/deploy/deploy_tasks.go (actually running the deploy/update)

pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -102,14 +102,29 @@ func run(ctx context.Context, cfg config) error {
 		return errors.New("Cannot specify both --yes and --no")
 	}
 
-	if cfg.dev && definitions.IsTaskDef(cfg.paths[0]) {
-		return deployFromTaskDefn(ctx, cfg)
-	}
-
 	ext := filepath.Ext(cfg.paths[0])
-	if ext == ".yml" || ext == ".yaml" {
+	if !cfg.dev && (ext == ".yml" || ext == ".yaml") && len(cfg.paths) == 1 {
```
leeweisberger, line 106: You can only get legacy yaml by not including the dev flag and only including one file. I think we should remove this code after we launch yaml and move our outstanding customer onto it

pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -102,14 +102,29 @@ func run(ctx context.Context, cfg config) error {
...
+		// Legacy YAML.
 		return deployFromYaml(ctx, cfg)
 	}
 
-	return NewDeployer().deployFromScript(ctx, cfg)
+	d := &discover.Discoverer{
```
leeweisberger, line 111: This is the entrypoint. Pretty easy to follow:
1. Discover tasks
2. Deploy discovered tasks

pkg/cmd/tasks/deploy/deploy_tasks.go
```diff
@@ -136,67 +127,21 @@ func (d *scriptDeployer) deployFromScript(ctx context.Context, cfg config) error
...
 	return groupErr
 }
 
-type script struct {
-	file     string
-	taskSlug string
-}
-
-// discoverScripts recursively discovers Airplane task scripts.
-func (d *scriptDeployer) discoverScripts(ctx context.Context, paths ...string) ([]script, error) {
```
leeweisberger, line 151: This was moved to the discovery phase!

pkg/cmd/tasks/deploy/deploy_tasks.go
```diff
@@ -240,128 +177,56 @@ More information: https://apn.sh/jst-upgrade`)
...
-		TaskID:  task.ID,
-		Root:    tc.taskRoot,
-		Def:     tc.def,
-		TaskEnv: env,
-		Shim:    true,
-		GitMeta: gitMeta,
-	})
-	if err != nil {
+	var image *string
+	if ok, err := libBuild.NeedsBuilding(kind); err != nil {
```
leeweisberger, line 185: I took this code from the yaml code.

We're going to need to update this to support sql / rest (currently errors)

pkg/cmd/tasks/deploy/discover/defn_discover.go
```diff
@@ -0,0 +1,75 @@
+package discover
```
leeweisberger, line 1: All of this code is copy pasted.

pkg/cmd/tasks/deploy/discover/discover.go
```diff
@@ -0,0 +1,103 @@
+package discover
```
leeweisberger, line 1: All of this code is copy pasted except where I call out below

pkg/cmd/tasks/deploy/discover/discover.go
```diff
@@ -0,0 +1,103 @@
...
+			task, err := d.Client.GetTask(ctx, slug)
+			if err != nil {
+				return nil, err
+			}
+			taskConfig, err := td.GetTaskConfig(ctx, task, p)
+			if err != nil {
+				return nil, err
+			}
+			taskConfig.From = td.TaskConfigSource()
+			taskConfigs = append(taskConfigs, taskConfig)
```
leeweisberger, line 98: This is the only change. Once we find a file, see if it is an airplane task using the task discovers. If it is, convert to a task config

pkg/cmd/tasks/deploy/discover/script_discover.go
```diff
@@ -0,0 +1,71 @@
+package discover
```
leeweisberger, line 1: This code is copy pasted. This is how we discover script (lite) airplane tasks

pkg/taskdir/definitions/validate.go
```diff
@@ -20,7 +22,7 @@ type ErrSchemaValidation struct {
 }
 
 func (err ErrSchemaValidation) Error() string {
-	return "invalid YAML format"
+	return fmt.Sprintf("invalid YAML format: %v", err.Errors)
```
leeweisberger, line 25: Including the json schema error. We might want to format differently, but this is more useful than just barfing. I used this when writing tests to figure out what was wrong with my yaml.

pkg/cmd/tasks/deploy/predeploy.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 // ensureConfigsExist checks for config references in env and asks users to create any missing ones
-func ensureConfigsExist(ctx context.Context, client *api.Client, def definitions.Definition) error {
+func ensureConfigsExist(ctx context.Context, client api.APIClient, def definitions.Definition) error {
```
francescaleung, line 18: hmm this should get changed to a DefinitionInterface, make sure everyone's calling it

francescaleung (changes requested):
We should also make sure that discovery detects duplicate tasks (e.g., your script is still comment-linked but you also have a task defn file). Your choice whether to handle that in this PR or a separate one.

pkg/cmd/tasks/deploy/discover/discover.go
```diff
@@ -0,0 +1,103 @@
...
+			if err != nil {
+				return nil, err
+			}
+			if slug == "" {
+				// The file is not an Airplane task.
+				continue
+			}
+			task, err := d.Client.GetTask(ctx, slug)
+			if err != nil {
+				return nil, err
```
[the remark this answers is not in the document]
leeweisberger: Yeah I have to merge in your changes here

pkg/cmd/tasks/deploy/deploy_tasks.go
```diff
@@ -0,0 +1,314 @@
...
+			GitMeta: gitMeta,
+		})
+		if err != nil {
+			return err
+		}
+		tp.buildID = resp.BuildID
+		image = &resp.ImageURL
+	}
+
+	utr, err := tc.Def.UpdateTaskRequest(ctx, client, image)
```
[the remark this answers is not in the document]
leeweisberger: Are you saying I should remove it and change image -> nil... and that's it?

pkg/cmd/tasks/deploy/deploy_tasks.go
```diff
@@ -0,0 +1,314 @@
...
+	if err != nil {
+		return err
+	}
+
+	utr.BuildID = pointers.String(tp.buildID)
+	utr.InterpolationMode = interpolationMode
+	utr.RequireExplicitPermissions = tc.Task.RequireExplicitPermissions
+	utr.Permissions = tc.Task.Permissions
```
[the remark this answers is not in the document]
leeweisberger: Can you explain? Are you saying that the taskConfig won't have these values and we shouldn't overwrite the values in the utr? I don't understand

pkg/cmd/tasks/deploy/predeploy.go
leeweisberger: I didn't change this - you're saying that it was just wrong in the first place? It should be an interface and we should call it before deploying? What's the purpose of this (I'm not really following)

leeweisberger:
I added back the "task doesn't exist behavior" here: https://github.com/airplanedev/cli/pull/345/commits/40a278702b3273bea2276f7be89151cd4c2fd77c

pkg/cmd/tasks/deploy/deploy_tasks.go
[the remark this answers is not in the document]
francescaleung: I'm saying that the UpdateTaskRequest object coming out of Definition_0_3.GetUpdateTaskRequest will fill in these fields (because Definition_0_3 defines them) but the UpdateTaskRequest object coming out of Definition.GetUpdateTaskRequest won't. These lines mean any changes to permissions in a task defn won't get propagated, but not having these lines means that you lose any explicit permissions on tasks deployed by script.

pkg/cmd/tasks/deploy/predeploy.go
francescaleung: This function appears to be called only from deploy-from-yaml, which kind of makes sense since deploy-from-script doesn't have the ability to specify configs. But deploy-from-defn does, so maybe it should get called at least for deploy-from-defn.

pkg/cmd/tasks/deploy/deploy_tasks.go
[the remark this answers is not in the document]
francescaleung:
```suggestion
	utr, err := tc.Def.GetUpdateTaskRequest(ctx, client)
	...
	if image != nil {
		utr.Image = image
	}
```

pkg/cmd/tasks/deploy/discover/discover.go
```diff
@@ -0,0 +1,117 @@
...
+				continue
+			}
+			task, err := d.Client.GetTask(ctx, slug)
+			if err != nil {
+				var missingErr *api.TaskMissingError
+				if errors.As(err, &missingErr) {
+					task, err = td.HandleMissingTask(ctx, p)
+					if err != nil {
+						return nil, err
+					} else if task.ID == "" {
```
francescaleung, line 99: maybe HandleMissingTask should return a pointer instead, so this can be a nil check

francescaleung (changes requested):
Looking good, concerns from first review still applicable though

pkg/cmd/tasks/deploy/predeploy.go
leeweisberger: Updated. I didn't understand what the function was doing - I gave it a slight rename

pkg/cmd/tasks/deploy/discover/discover.go
leeweisberger: Good call. Done

pkg/cmd/tasks/deploy/deploy_tasks.go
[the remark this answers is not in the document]
francescaleung: This still needs to be addressed!

francescaleung (changes requested):

pkg/cmd/tasks/deploy/deploy_tasks.go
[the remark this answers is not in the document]
leeweisberger: Missed this.

How's this: https://github.com/airplanedev/cli/pull/345/commits/dba6976678f51831078181102686f2850f1c07a4

When you finish the permissions todo, you can write a test for this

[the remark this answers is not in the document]
francescaleung: Doesn't this fail if you're moving from explicit permissions to no explicit permissions?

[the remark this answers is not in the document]
leeweisberger: Ugh you're right. I think the only way to do it then is to pass in the current state and make the call in the update fn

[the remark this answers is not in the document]
leeweisberger: I don't think it's that bad: https://github.com/airplanedev/cli/pull/345/commits/a9a6fee57efad7a85a9af6edc3aa3bd1b3136d77

francescaleung (approved):

PR #348, opened 2021-12-10
Remove kind and options from task config

leeweisberger:
We can already get these values from the def, so no need to also have them on the config.

I also renamed TaskFilePath -> TaskEntryPoint to be more consistent with our conventions


pkg/cmd/tasks/deploy/discover/discover_test.go
```diff
@@ -39,18 +39,11 @@ func TestDiscoverTasks(t *testing.T) {
 				{
 					TaskRoot:         fixturesPath,
 					WorkingDirectory: fixturesPath,
-					TaskFilePath:     fixturesPath + "/single_task.js",
-					Kind:             build.TaskKindNode,
+					TaskEntryPoint:     fixturesPath + "/single_task.js",
```
francescaleung, line 42: gofmt?

francescaleung (approved):

PR #359, opened 2022-01-06
Use shared deployed code in lib

leeweisberger:
Relies on moving shared deployed code to lib https://github.com/airplanedev/lib/pull/16/files

---

Other than code deletion, the only real change here is to move shared api types from here -> lib. This ensures that there is only one source of truth for structs


pkg/cmd/tasks/deploy/client.go
```diff
@@ -0,0 +1,209 @@
+package deploy
+
+import (
+	"context"
+
+	"github.com/airplanedev/cli/pkg/api"
+	deployapi "github.com/airplanedev/lib/pkg/deploy/api"
+)
+
+type deployClient struct {
```
leeweisberger, line 10: This is the struct that satisfies the lib's deploy api client interface. It just uses the api client and converts structs to their identical counterparts :crying_cat_face: again open to other suggestions here!

pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -136,3 +142,52 @@ func run(ctx context.Context, cfg config) error {
 
 	return NewDeployer(cfg, l, DeployerOpts{}).DeployTasks(ctx, taskConfigs)
 }
+
+func HandleMissingTask(cfg config, l logger.Logger) func(ctx context.Context, def definitions.DefinitionInterface) (*deployapi.Task, error) {
```
leeweisberger, line 146: And then related to my comment in lib, here the cli provides the missing task handler that is cli specific (prompts the user to create a new task)

pkg/cmd/tasks/deploy/client.go
francescaleung: Is this really less work than using the structs defined in `lib`?

leeweisberger: We discussed offline. I changed to storing source of truth for shared structs in lib

pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -136,3 +140,52 @@ func run(ctx context.Context, cfg config) error {
 
 	return NewDeployer(cfg, l, DeployerOpts{}).DeployTasks(ctx, taskConfigs)
 }
+
+func HandleMissingTask(cfg config, l logger.Logger) func(ctx context.Context, def definitions.DefinitionInterface) (*libapi.Task, error) {
```
leeweisberger, line 144: This is the only bit of new code. This is cli specific, so the cli passes it into the task discoverer

pkg/api/types.go
```diff
@@ -5,8 +5,10 @@ import (
 	"time"
 
 	"github.com/airplanedev/lib/pkg/build"
+	// Some types are imported from lib. Eventually we might want all of these types to live in lib. For now,
+	// we can move tasks from here -> lib on an as-needed basis.
+	. "github.com/airplanedev/lib/pkg/deploy/api"
```
francescaleung, line 10: ooh, did not know about this. Having said that, seems like it's [not recommended to be used outside of test packages](https://github.com/golang/go/wiki/CodeReviewComments#import-dot). There's also a [two-year-old issue](https://github.com/golang/go/issues/29326) proposing the removal of dot imports (though no consensus there). I'd be leaning towards not using dot imports, I think.

francescaleung (approved):

pkg/api/types.go
leeweisberger: ah fooey will revert. It was actually really nice because it provided a check that we weren't duping the same structs, but I'll change

leeweisberger: I learned this from @colinking he let me down

PR #362, opened 2022-01-13
Handle init for existing tasks

francescaleung:
Properly handles `init` for existing tasks. Also does not generate a generic entrypoint for existing tasks, but informs the user where the entrypoint is set. This is probably not a _great_ solution--if you run `airplane init --dev --from hello_world hello_world.py`, the file argument is completely ignored. Maybe the `entrypoint` should be overwritten in that case?

Also fixes a bug where if a definition file was supplied to `init` it would not get used & instead default to `<slug>.task.<format>`.

Depends on https://github.com/airplanedev/lib/pull/24. Resolves AIR-2460.


leeweisberger (approved):

francescaleung:
> > dles `init` for existing tasks. Also does not generate a generic entrypoint for existing task
> 
> I guess our options if the task already exists are
> 
>     1. Do what you do and ignore
> 
>     2. Overwrite the entrypoint - seems kinda weird?
> 
>     3. Copy over the entrypoint file from old place to new place... also kinda weird
> 
>     4. Error
> 
> 
> I think 1 makes the most sense?

I think I've waffled on this a little--I'm going to do generate the entrypoint if the entrypoint doesn't already exist. It'll be similar to current `airplane init` & future `airplane init --code-only` runs.

PR #364, opened 2022-01-14
Bump lib

leeweisberger:
Bump lib with fix for permission roles.


francescaleung:
I think you should run a `go mod tidy`

francescaleung (approved):

PR #368, opened 2022-01-19
Pass kind and build config when creating build

leeweisberger:


pkg/api/types.go
```diff
@@ -13,20 +13,20 @@ import (
 
 // CreateTaskRequest creates a new task.
 type CreateTaskRequest struct {
-	Slug             string            `json:"slug"`
```
leeweisberger, line 16: just formatting

francescaleung (approved):

PR #369, opened 2022-01-19
Fix execute hanging on pipe

stepzhou:
## Summary

The `survey` module doesn't seem to handle non-terminal i/o very well. See https://github.com/AlecAivazis/survey/issues/363 and https://github.com/AlecAivazis/survey/issues/328.

However, I thought it was strange that it was only the "Execute" prompt that was hanging while all the param prompts were fine. I pattern-matched the `survey.WithStdio(os.Stdin, os.Stderr, os.Stderr)` option over to the "Execute" prompt and now it works. I'm not sure exactly what's going on here but my guess is that `survey.WithStdio(os.Stdin, os.Stderr, os.Stderr)` is setting the survey context back to terminal stdio. 🤷 

Resolves AIR-2733

## Testing

* Verify that executing a task with multiple params does not hang and executes the task
* Verify that the output of the task is properly piped

```
go run cmd/airplane/main.go --host api.airplane.so:5000 execute copy_of_ping > output.txt
Executing Copy of Ping task: https://app.airplane.so:5000/t/copy_of_ping
? Value (--value): test
? test (--test): test
? Execute? Yes
Queued run: https://app.airplane.so:5000/runs/run20220119zn8wfyi
[log] airplane_output_set "test"
Starting run...
Pulling image ubuntu:latest
```


joshma:
> Fixes https://linear.app/airplane/issue/AIR-2733/fix-cli-execute-to-avoid-hanging-on-pipe

You can do `Resolves AIR-2733` instead to link it!

joshma (approved):

PR #371, opened 2022-01-25
Fix handling for empty entrypoint in init

francescaleung:
Previously, if an entrypoint wasn't defined, the user wouldn't be prompted for a new entrypoint even if the task required one. This is a regression; now we prompt the user for an entrypoint if one isn't defined in the task yet.

Depends on https://github.com/airplanedev/lib/pull/36.


pkg/cmd/tasks/initcmd/init.go
```diff
@@ -355,8 +365,8 @@ func patch(slug, file string) (ok bool, err error) {
 	return
 }
 
-func promptForNewFileName(task libapi.Task) (string, error) {
-	fileName := task.Slug + runtime.SuggestExt(task.Kind)
+func promptForNewFileName(slug string, kind build.TaskKind) (string, error) {
```
leeweisberger, line 368: I like splitting this up when we really only need some metadata from the struct

PR #374, opened 2022-01-28
Add `--env` flag to task commands

colinking:
This PR adds an `--env` flag to:
- `airplane tasks deploy`
- `airplane tasks dev`
- `airplane tasks execute`
- `airplane tasks get`
- `airplane tasks init`
- `airplane tasks list`

In other words, you can now deploy tasks into different environments, view tasks in different environments, and execute tasks in different environments.

A future PR will add it to:
- `airplane tasks open`: once we have UI support for envs

This is a hidden parameter that will be made visible once we release environments.

Depends on:
- https://github.com/airplanedev/airport/pull/1487
- https://github.com/airplanedev/lib/pull/38
- https://github.com/airplanedev/cli/pull/368(which removes the `UpdateTask` call when building)

Resolves: AIR-2791

cc @ameyakhare


go.mod
```diff
@@ -38,7 +38,7 @@ require (
 )
 
 require (
-	github.com/airplanedev/lib v0.2.11
+	github.com/airplanedev/lib v0.2.12-0.20220128193213-a8a6a1c0a496
```
colinking, line 41: Will update with a tag once https://github.com/airplanedev/lib/pull/38 merges.

pkg/params/cli.go
```diff
@@ -19,7 +19,7 @@ import (
 //
 // A flag.ErrHelp error will be returned if a -h or --help was provided, in which case
 // this function will print out help text on how to pass this task's parameters as flags.
-func CLI(args []string, client *api.Client, task libapi.Task) (api.Values, error) {
+func CLI(args []string, task libapi.Task) (api.Values, error) {
```
colinking, line 22: Noticed this wasn't used while tracing where we used `client`.

pkg/cmd/tasks/dev/dev.go
```diff
@@ -62,6 +64,12 @@ func New(c *cli.Config) *cobra.Command {
 		},
 	}
 
+	// Unhide this flag once we release environments.
+	cmd.Flags().StringVar(&cfg.envSlug, "env", "", "The slug of the environment to query. Defaults to your team's default environment.")
+	if err := cmd.Flags().MarkHidden("env"); err != nil {
+		logger.Debug("unable to hide --env: %s", err)
+	}
```
francescaleung, line 71: Seems a _little_ odd that local execution would require an `--env` flag

pkg/build/remote.go
```diff
@@ -118,54 +118,24 @@ func (d *registryTokenGetter) getRegistryToken(ctx context.Context, client api.A
 	return registryToken, nil
 }
 
-func updateKindAndOptions(ctx context.Context, client api.APIClient, def definitions.DefinitionInterface, shim bool) error {
-	task, err := client.GetTask(ctx, def.GetSlug())
+func getKindAndConfig(ctx context.Context, def definitions.DefinitionInterface, shim bool) (build.TaskKind, build.KindOptions, error) {
```
francescaleung, line 121: `getKindAndConfig` -> `getKindAndBuildConfig`?

francescaleung (approved):

pkg/cmd/tasks/dev/dev.go
colinking: Yep. I would like for `airplane dev` to work offline eventually. It won't need `--env` for deving on scripts (once we add support for that) since all the task config is in the definition file. But for deving on comment-based scripts, it needs to fetch that config from the API.

pkg/build/remote.go
colinking: This was from a different PR. Think you saw this when there were new changes on `main` that had not been pulled into this PR yet.

PR #377, opened 2022-02-01
Fix field casing in YAML formatted runs and tasks

colinking:
We were missing YAML struct tags on a few types. The YAML encoder was falling back on lowercasing the field name.

There is additionally a breaking change to our JSON format for tasks/runs to fix their ID format. I changed from `taskID -> id` and `runID -> id`.


zhan-airplane:
When you say breaking change to our JSON format, what exactly do you mean? Like, what might break?

colinking:
> When you say breaking change to our JSON format, what exactly do you mean? Like, what might break?

This is a breaking change for folks who use `airplane runs list` (or `airplane tasks list`) since we no longer produce the ID as `runID/taskID`. There's one customer we have in mind who is doing that that we'll notify.

zhan-airplane (approved):

PR #378, opened 2022-02-02
Run deploys from CLI

leeweisberger:
Rather than creating builds and updating tasks, we now:
1. Discover airplane tasks.
2. Tar and upload tasks that need to be built.
3. Collect git metadata about the deploy.
4. Create a deploy.
5. Log deploy logs and wait for the deploy to complete.

---
There is one piece of complexity here - technically someone can deploy tasks that are part of different git repos. We store git metadata on the deploy and I think it generally makes sense here for the product. I decided, at least for now, to detect if there are tasks from multiple git repos in one deploy, log a message to sentry, and not include git metadata on the deploy. If we have to, we could move git metadata onto individual tasks. I imagine this will be very uncommon.

This will officially launch deploys, so I have to make sure everything is working downstream before I merge this.


pkg/analytics/analytics.go
```diff
@@ -141,6 +141,14 @@ func ReportError(err error) {
 	}
 }
 
+// CaptureMessage sends a message to Sentry.
+func CaptureMessage(msg string) {
```
leeweisberger, line 145: @colinking I don't think we use this anywhere - I just want a log so I can track how often unexpected things happen. We don't have any other logging system hooked up (just sentry and segment). Thoughts on this?

pkg/api/client.go
```diff
@@ -87,15 +87,15 @@ type APIClient interface {
 	SetConfig(ctx context.Context, req SetConfigRequest) (err error)
 	GetConfig(ctx context.Context, req GetConfigRequest) (res GetConfigResponse, err error)
 
-	CreateBuild(ctx context.Context, req CreateBuildRequest) (res CreateBuildResponse, err error)
 	GetRegistryToken(ctx context.Context) (res RegistryTokenResponse, err error)
 	CreateBuildUpload(ctx context.Context, req libapi.CreateBuildUploadRequest) (res libapi.CreateBuildUploadResponse, err error)
-	GetBuildLogs(ctx context.Context, buildID string, prevToken string) (res GetBuildLogsResponse, err error)
-	GetBuild(ctx context.Context, id string) (res GetBuildResponse, err error)
+	GetDeployLogs(ctx context.Context, deploymentID string, prevToken string) (res GetDeployLogsResponse, err error)
```
leeweisberger, line 92: We are now dealing with deploys, not builds!

pkg/build/local.go
```diff
@@ -97,3 +104,18 @@ func getBuildEnv(ctx context.Context, client api.APIClient, taskEnv libapi.TaskE
 	}
 	return buildEnv, nil
 }
+
+func (d *registryTokenGetter) getRegistryToken(ctx context.Context, client api.APIClient) (registryToken api.RegistryTokenResponse, err error) {
```
leeweisberger, line 108: I moved this from build/remote.go (which is now deleted). No other change

pkg/build/mocks.go
```diff
@@ -12,7 +12,6 @@ var _ BuildCreator = &MockBuildCreator{}
 
 func (mbc *MockBuildCreator) CreateBuild(ctx context.Context, req Request) (*build.Response, error) {
 	return &build.Response{
-		BuildID:  "buildID",
```
leeweisberger, line 15: We are only creating builds locally, and no build id is produced

pkg/cmd/tasks/deploy/git.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"path/filepath"
+	"strings"
+
+	"github.com/airplanedev/cli/pkg/api"
+	"github.com/go-git/go-git/v5"
+	"github.com/pkg/errors"
+)
+
+type GitRepoGetter interface {
```
leeweisberger, line 12: I moved git functions here and put GetGitRepo behind an interface so we can mock it out for testing

pkg/cmd/tasks/deploy/git.go
```diff
@@ -0,0 +1,85 @@
...
+
+func GetEntrypointRelativeToGitRoot(repo *git.Repository, taskFilePath string) (string, error) {
+	w, err := repo.Worktree()
+	if err != nil {
+		return "", err
+	}
+	return filepath.Rel(w.Filesystem.Root(), taskFilePath)
+}
+
+func GetGitMetadata(repo *git.Repository) (api.GitMetadata, error) {
```
leeweisberger, line 49: No changes to this function. Just moved

pkg/cmd/tasks/deploy/deploy_tasks.go
```diff
@@ -99,60 +112,95 @@ func (d *deployer) DeployTasks(ctx context.Context, taskConfigs []discover.TaskC
...
+		var gitRoot string
+		if repo != nil {
+			w, err := repo.Worktree()
+			if err == nil {
+				gitRoot = w.Filesystem.Root()
 			}
-			return nil
-		})
+		}
+		gitRoots = append(gitRoots, gitRoot)
```
leeweisberger, line 146: This only exists for the case I talked about in the pr description - if tasks from multiple git repos are deployed together, drop git metadata

pkg/cmd/tasks/deploy/deploy_tasks.go
```diff
@@ -99,60 +112,95 @@ func (d *deployer) DeployTasks(ctx context.Context, taskConfigs []discover.TaskC
...
+	resp, err := d.cfg.client.CreateDeploy(ctx, api.CreateDeployRequest{
+		Tasks:       tasksToDeploy,
+		GitMetadata: gitMeta,
+	})
+	if err != nil {
+		return err
 	}
 
-	return groupErr
+	// TODO log URL to deploy
```
leeweisberger, line 182: We're going to log the URL to the deploys page that @rose-huang is building when that is ready

pkg/cmd/tasks/deploy/deploy_tasks.go
```diff
@@ -187,108 +235,205 @@ More information: https://apn.sh/jst-upgrade`)
...
+}
 
-	repo, err := git.PlainOpenWithOptions(filepath.Dir(taskFilePath), &git.PlainOpenOptions{
-		DetectDotGit: true,
-	})
-	if err != nil {
-		if errors.Is(err, git.ErrRepositoryNotExists) {
-			return meta, nil
+// tarAndUploadTasks concurrently tars and uploads tasks that need building.
+func (d *deployer) tarAndUploadTasks(ctx context.Context, taskConfigs []discover.TaskConfig) (map[string]string, error) {
```
leeweisberger, line 303: None of this tarring and uploading stuff is new - It's just taken from remote.go (which was deleted). I copy and pasted it and made slight modifications

pkg/cmd/tasks/deploy/deploy_tasks.go
```diff
@@ -187,108 +235,205 @@ More information: https://apn.sh/jst-upgrade`)
...
+		)})
 	}
-	meta.IsDirty = !status.IsClean()
+	return uploadID, nil
+}
 
-	h, err := repo.Head()
-	if err != nil {
-		return meta, err
+func waitForDeploy(ctx context.Context, loader logger.Loader, client api.APIClient, deploymentID string) error {
```
leeweisberger, line 352: This was also copied from remote.go, just updated from `waitForBuild` to `waitForDeploy`

pkg/cmd/tasks/deploy/deploy_tasks.go
```diff
@@ -187,108 +235,205 @@ More information: https://apn.sh/jst-upgrade`)
...
 
-	commit, err := repo.CommitObject(h.Hash())
-	if err != nil {
-		return meta, err
+type deployLogReq struct {
+	taskSlug string
+	msg      string
+}
+
+func deployLog(ctx context.Context, level api.LogLevel, loader logger.Loader, req deployLogReq, args ...interface{}) {
```
leeweisberger, line 406: Also copied from remote.go.

pkg/cmd/tasks/deploy/deploy_tasks.go
```diff
@@ -187,108 +235,205 @@ More information: https://apn.sh/jst-upgrade`)
...
-	if h.Name().IsBranch() {
-		ref = strings.TrimPrefix(ref, "refs/heads/")
+	if loaderActive {
+		loader.Start()
 	}
-	meta.Ref = ref
+}
 
-	return meta, nil
+func confirmBuildRoot(root string) error {
```
leeweisberger, line 423: Copied from remote.go. Didn't make any changes to this

francescaleung (approved):

pkg/analytics/analytics.go
leeweisberger: We actually don't initialize a sentry client if the user hasn't opted in, so this will be a no-op

pkg/api/client.go
```diff
@@ -87,11 +87,12 @@ type APIClient interface {
...
 
-	CreateBuild(ctx context.Context, req CreateBuildRequest) (res CreateBuildResponse, err error)
 	GetRegistryToken(ctx context.Context) (res RegistryTokenResponse, err error)
 	CreateBuildUpload(ctx context.Context, req libapi.CreateBuildUploadRequest) (res libapi.CreateBuildUploadResponse, err error)
-	GetBuildLogs(ctx context.Context, buildID string, prevToken string) (res GetBuildLogsResponse, err error)
-	GetBuild(ctx context.Context, id string) (res GetBuildResponse, err error)
+
+	GetDeployLogs(ctx context.Context, deploymentID string, prevToken string) (res GetDeployLogsResponse, err error)
+	GetDeployment(ctx context.Context, id string) (res Deployment, err error)
+	CreateDeploy(ctx context.Context, req CreateDeployRequest) (CreateDeployResponse, error)
```
colinking, line 95: `CreateDeploy / GetDeployment` are inconsistent. IIRC `Deployment` is the noun here? Let's update the API before releasing this.

pkg/api/client.go
```diff
@@ -351,17 +346,22 @@ func (c Client) DeleteAPIKey(ctx context.Context, req DeleteAPIKeyRequest) (err
...
+		"deploymentID": []string{deploymentID},
 	}
 	if logger.EnableDebug {
 		q.Set("level", "debug")
 	}
 	if prevToken != "" {
 		q.Set("prev_token", prevToken)
 	}
-	err = c.do(ctx, "GET", "/builds/getLogs?"+q.Encode(), nil, &res)
+	err = c.do(ctx, "GET", "/deploys/getLogs?"+q.Encode(), nil, &res)
```
colinking, line 364: There's a helper for query params (`encodeQueryString`) that filters out empty string values.

pkg/analytics/analytics.go
colinking: Neat, didn't know about Sentry messages.

Minor: `ReportMessage` to match `ReportError`? 😇

pkg/api/types.go
```diff
@@ -316,3 +264,46 @@ type APIKey struct {
...
 }
+
+type DeployTask struct {
+	TaskID            string                   `json:"taskID"`
+	InterpolationMode string                   `json:"interpolationMode"`
+	Kind              build.TaskKind           `json:"kind"`
+	BuildConfig       build.KindOptions        `json:"buildConfig"`
+	UploadID          string                   `json:"uploadID"`
+	UpdateTaskRequest libapi.UpdateTaskRequest `json:"updateTaskRequest"`
+	Env               libapi.TaskEnv           `json:"env"`
```
colinking, line 275: Can you rename this to `EnvVars`? Since we're introducing environments, it no longer is clear this refers to environment variables. We'll make this change in our existing APIs down the road.

pkg/cmd/tasks/deploy/yaml.go
```diff
@@ -154,8 +153,6 @@ More information: https://apn.sh/jst-upgrade`)
 		var bc build.BuildCreator
 		if cfg.local {
 			bc = build.NewLocalBuildCreator()
-		} else {
-			bc = build.NewRemoteBuildCreator(archive.NewAPIArchiver(&logger.StdErrLogger{}, cfg.client, &archive.HttpUploader{}))
```
colinking, line 158: This is a breaking change -- anyone using the legacy YAML format will no longer be able to do their builds remotely. Was that intended?

pkg/cmd/tasks/deploy/deploy_tasks.go
```diff
@@ -36,28 +39,38 @@ type deployer struct {
...
+		loader:       logger.NewLoader(logger.LoaderOpts{HideLoader: logger.EnableDebug}),
+		archiver:     a,
+		repoGetter:   rg,
 	}
 }
 
 // DeployTasks deploys all taskConfigs as Airplane tasks.
-// It concurrently builds (if needed) and updates tasks.
 func (d *deployer) DeployTasks(ctx context.Context, taskConfigs []discover.TaskConfig) error {
 	if len(d.cfg.changedFiles) > 0 {
```
colinking, line 75: This function is rather large -- consider splitting out some of the logic into separate functions, e.g. `filterTasks()`. Should make this easier to skim.

colinking (changes requested):
Let's chat about how this impacts folks on the legacy YAML format.

pkg/cmd/tasks/deploy/yaml.go
leeweisberger: Gar this was unintentional. I added back the remote build creator b/c I don't want to put in the work to convert legacy yaml -> deploys... Once we launch our other task yaml I think we should push to get rid of this cc @francescaleung

pkg/api/client.go
leeweisberger: My thinking here was that Deploy is the process and Deployment is the resource. CreateDeploy kicks off the deploy process (and as part of that, creates a deployment resource) while GetDeployment just gets the resource. 

If that's confusing we can either
a) call everything Deployment
b) Change the endpoint to TriggerDeploy or something like that?

thoughts?

leeweisberger: Yeah the more I think of it this is probably confusing and we should probably move to a)

leeweisberger: I moved over to a

colinking (approved):

PR #384, opened 2022-02-04
Add env flag to configs commands

ameyakhare:
Resolves AIR-2830


pkg/cmd/configs/get/get.go
```diff
@@ -6,38 +6,54 @@ import (
...
 
 // Run runs the get command.
-func run(ctx context.Context, c *cli.Config, name string) error {
-	var client = c.Client
+func run(ctx context.Context, cfg config) error {
+	var client = cfg.root.Client
 
-	nt, err := configs.ParseName(name)
+	nt, err := configs.ParseName(cfg.name)
 	if err == configs.ErrInvalidConfigName {
```
colinking, line 49: If this function could return other errors, we would be silently ignoring them. Seems fine to remove this `ErrInvalidConfigName` check and treat all errors as invalid config names. Maybe could clarify that by returning a boolean instead of an error?

pkg/cmd/tasks/deploy/predeploy.go
```diff
@@ -15,23 +15,23 @@ import (
...
+			if err := ensureConfigVarExists(ctx, client, k, *v.Config, envSlug); err != nil {
 				return err
 			}
 		}
 	}
 	return nil
 }
 
-func ensureConfigVarExists(ctx context.Context, client api.APIClient, envName, configName string) error {
+func ensureConfigVarExists(ctx context.Context, client api.APIClient, envName, configName, envSlug string) error {
```
colinking, line 34: Can you update this with a struct? Since there are three string types in a row.

colinking (approved):

pkg/cmd/configs/get/get.go
ameyakhare: just changed it to a != nil check. I looked over the function and it does only return that type of error or nil, but better safe than sorry

PR #386, opened 2022-02-09
Upgrade lib + add DefnDiscoverer before ScriptDiscoverer

francescaleung:
For https://github.com/airplanedev/lib/pull/45.


leeweisberger (approved):
FYI we're also going to have to add the defn discoverer to airport https://github.com/airplanedev/airport/blob/main/pkg/deploys/deploys.go#L246

PR #387, opened 2022-02-09
Hide loader when prompting user for input in deploy

francescaleung:
Resolves AIR-2801.


leeweisberger (approved):
Looks good - I have an outstanding pr I've been meaning to get in that will take care of all of the loader hiding automatically at least as related to logs

PR #390, opened 2022-02-10
Look for a defn file if a task config is provided via script

francescaleung:
Deliberately only for CLI.

Resolves AIR-2484.


leeweisberger:
General question - what if you have two task entrypoints, foo.py with slug bar and bar.py with slug foo and task defintion files foo.task.yaml and bar.task.yaml. How do we resolve this?

leeweisberger (changes requested):
Before I look at the code, can you tell me what the user flow is here? What setup / action would lead to this codepath being used?

francescaleung:
> Before I look at the code, can you tell me what the user flow is here? What setup / action would lead to this codepath being used?

User has a script file (foo.py) linked to a task (foo). User also has a definition file (foo.task.yaml) linked to the same task (foo). User tries to deploy the script file (foo.py). This patch prompts the user to switch to the definition file instead of just deploying from the script file, reason being that the defn file contains more information than what's in the script file (e.g., execute rules, task description, constraints).

francescaleung:
> General question - what if you have two task entrypoints, foo.py with slug bar and bar.py with slug foo and task defintion files foo.task.yaml and bar.task.yaml. How do we resolve this?

We match on slug. If foo.py has slug bar, it matches bar.task.yaml. If bar.py has slug foo, it matches foo.task.yaml (assuming <slug>.task.yaml).

pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -214,3 +229,52 @@ func HandleMissingTask(cfg config, l logger.Logger, loader logger.Loader) func(c
...
+				continue
+			}
+
+			path := filepath.Join(dir, fileInfo.Name())
+			if slug, err := defnDiscoverer.IsAirplaneTask(ctx, path); err != nil {
+				return nil, err
+			} else if slug != taskConfig.Task.Slug {
+				continue
+			}
+			question := fmt.Sprintf("A definition file for task %s exists. Would you like to use it?", taskConfig.Task.Slug)
```
leeweisberger, line 262: Can we log where it exists so they know that we are going to use it instead of the script file?

pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -214,3 +229,52 @@ func HandleMissingTask(cfg config, l logger.Logger, loader logger.Loader) func(c
 		return &task, nil
 	}
 }
+
+// Look for a defn file that matches this task config, in the directory where the entrypoint is
+// located & also in the current directory.
+func findDefinition(ctx context.Context, cfg config, defnDiscoverer *discover.DefnDiscoverer, taskConfig discover.TaskConfig) (*discover.TaskConfig, error) {
```
leeweisberger, line 235: I think we can name this better. Maybe `findDefinitionInSameDirectory`. Might also want to include in the name that it only applies to script files?

Can you also update the comment with the usecase - I don't think it's obvious. i.e. we are doing this in case the user accidentally chose to deploy the script file rather than the definition file in the same directory

leeweisberger (approved):

PR #400, opened 2022-02-17
Release SRC / tasks-as-code

francescaleung:
This patch releases SRC / tasks-as-code **as a non-default path** & deprecates legacy YAML deploys. It removes `--dev` flags & unhides new flags (though `--env` remains hidden). Most explicit references to Def03 are removed. ScriptDiscoverer is now being initialized with an IAPIClient. Users wishing to use tasks-as-code should run `airplane init --code-only=false`.

Resolves AIR-2741. Depends on https://github.com/airplanedev/lib/pull/55.

Additional PRs that were merged in:
* Fix local execution (#404)
* Check if local execution is supported (#405)
* Prompt user for where definition file should go (#406)
* Suggest next steps based on task kind (#408)
* Always ask for an entrypoint (#416)
* Look for a defn file if you're trying to deploy a sql file (#407)
* Upgrade lib + use GenerateCommentedFile in init (#418)
* Smoother experience for `init` (#420)
* Prep SRC for release (#435)

---

cc @joshma


pkg/cmd/tasks/initcmd/init.go
```diff
@@ -61,9 +60,10 @@ func New(c *cli.Config) *cobra.Command {
 		Use:   "init",
 		Short: "Initialize a task definition",
 		Example: heredoc.Doc(`
-			$ airplane tasks init --slug task-slug
-			$ airplane tasks init --slug task-slug ./my/task.js
-			$ airplane tasks init --slug task-slug ./my/task.ts
+			$ airplane tasks init --from task_slug
+			$ airplane tasks init --from task_slug ./folder/my_task.js
```
leeweisberger, line 64: To be clear - if they do this, we prompt them to create a .task.yaml right?

We might want to look at other CLIs, but would be nice to inlude one liners with these examples. What do we actually do if you run this command?

leeweisberger (approved):
Can you also delete:
- pkg/build/remote.go
- GetBuildLogs from client
- GetBuild from client
- CreateBuild from client
- All build related types from pkg/api/types.go

pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -50,9 +49,10 @@ func New(c *cli.Config) *cobra.Command {
 		Example: heredoc.Doc(`
 			airplane tasks deploy ./task.ts
 			airplane tasks deploy --local ./task.js
```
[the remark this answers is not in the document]
francescaleung: It does with https://github.com/airplanedev/cli/pull/404!

pkg/cmd/tasks/initcmd/init.go
francescaleung: Yes, a defn file does get created. There should probably some thought going into what we want to recommend, rather than just blindly listing all possible use cases. I might revisit this in a later PR.

PR #401, opened 2022-02-17
Rewrite defn file after upgrading to JST

francescaleung:
Resolves AIR-2974. Depends on https://github.com/airplanedev/lib/pull/58.


pkg/cmd/tasks/deploy/deploy_tasks.go
```diff
@@ -209,6 +209,11 @@ More information: https://apn.sh/jst-upgrade`)
 			if err := tc.Def.UpgradeJST(); err != nil {
 				return api.DeployTask{}, err
 			}
+			if tc.From == discover.TaskConfigSourceDefn && tc.FilePath != "" {
+				if err := tc.Def.Write(tc.FilePath); err != nil {
+					return api.DeployTask{}, err
+				}
+			}
```
leeweisberger, line 216: What if the filepath is non empty? Then we log that the task is being migrated but it actually isn't right? Maybe it's ok to always assume the filepath will be there.

It's also unfortunate that we still have to support this...

leeweisberger (approved):

pkg/cmd/tasks/deploy/deploy_tasks.go
francescaleung: It's still being migrated, it's just not being saved to disk.

PR #404, opened 2022-02-23
Fix local execution

francescaleung:
Local execution can now happen via defn file.

Resolves AIR-3058.


pkg/cmd/tasks/dev/dev.go
```diff
@@ -245,12 +251,26 @@ func getDevEnv(r runtime.Interface, path string) (map[string]string, error) {
...
-	slug, ok := runtime.Slug(file)
-	if !ok {
-		return "", runtime.ErrNotLinked{Path: file}
+func entrypointFrom(file string) (string, error) {
+	switch ext := filepath.Ext(file); ext {
+	case ".yml", ".yaml", ".json":
+		return entrypointFromDefn(file)
+	default:
+		return file, nil
+	}
```
leeweisberger, line 260: Maybe break this out into a helper in utils. We are doing this in another place as well. utils.IsDefintion

leeweisberger (approved):
Would be awesome if we had a test for this that stubbed out the actual run. Up to you

PR #407, opened 2022-02-24
Look for a defn file if you're trying to deploy a sql file

francescaleung:
If you're trying to deploy `subdir/foo.sql`, looks for a definition file `{,subdir/}foo.task.{yaml,yml,json}` that is already linked to a task.

Resolves AIR-3057.


leeweisberger (commented):
Why is a sql file different from a node file?

If I run `airplane deploy myNodeTask.js` and there is a defn file next to it, we want to deploy that defn file right? I feel like I've already seen a pr that implements this... can you clarify?

francescaleung:
We do this for node files that are linked to a task. SQL files are never linked, so we never pick them up via script discovery.

leeweisberger (changes requested):
I think we should add this to the discoverer - if the file is sql, run this code. We can make a new discoverer or we can add a one off that if there is only a single file specified, we run this code.  I'm not super worried about the perf of running it on every sql file. And then if we discover the task twice, we dedupe using your algo you already wrote.

I also don't think we need to task the user if they want to use the defn file - I feel like this is the only thing they could want by explicitly selecting a sql file?

Just kind of weird to have this code in deploy.go... I would think the discoverer would handle this. We can make

Let me know if you disagree

francescaleung:
Can you write a discoverer for this? There's no TaskConfig to return (we return a path that the defn discoverer should pick up), I only want it run on explicitly specified .sql files, & I want to prompt the user that they're running `airplane deploy` on the wrong file, because maybe we found the wrong task definition. Those things sound incompatible with the discovery pattern, but maybe I'm just looking at it wrong?

pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -262,3 +279,77 @@ func findDefinitionForScript(ctx context.Context, cfg config, defnDiscoverer *di
...
+			}); err != nil {
+				switch errors.Cause(err).(type) {
+				case *libapi.TaskMissingError:
+					continue
+				default:
+					return "", err
+				}
+			}
+
+			question := fmt.Sprintf("File %s is not linked to a task.\nFound definition file %s linked to task %s instead.\nWould you like to use it?", path, p, slug)
```
leeweisberger, line 343: I feel like we can shorten this to something like "Found definition file %s. Would you like to use it?"

leeweisberger (approved):

pkg/cmd/tasks/deploy/deploy.go
francescaleung: I like indicating which file we're trying to match to a definition file, in case you're deploying multiple things,

PR #412, opened 2022-03-01
Support deploying tasks to new environments

colinking:
Depends on: https://github.com/airplanedev/lib/pull/75


pkg/cmd/tasks/deploy/deploy_tasks.go
```diff
@@ -179,17 +180,17 @@ func (d *deployer) getDeployTask(ctx context.Context, tc discover.TaskConfig, up
...
 	tp := taskDeployedProps{
-		from:       string(tc.From),
+		source:     string(tc.Source),
 		kind:       kind,
-		taskID:     tc.Task.ID,
-		taskSlug:   tc.Task.Slug,
-		taskName:   tc.Task.Name,
+		taskID:     tc.TaskID,
+		taskSlug:   tc.Def.GetSlug(),
+		taskName:   tc.Def.GetName(),
```
francescaleung, line 187: Seems wrong somehow that we're using this in analytics...

francescaleung (approved):

pkg/cmd/tasks/deploy/deploy_tasks.go
colinking: In what way? As in, we shouldn't be exposing this in the definition interface just for analytics? Or that we shouldn't be sending the task name to our analytics?

pkg/cmd/tasks/deploy/deploy_tasks.go
```diff
@@ -31,10 +32,6 @@ type deployer struct {
 	loader       logger.Loader
 	archiver     archive.Archiver
 	repoGetter   GitRepoGetter
-
-	erroredTaskSlugs  map[string]error
-	deployedTaskSlugs []string
-	mu                sync.Mutex
```
colinking, line 37: This was dead code.

PR #413, opened 2022-03-02
Disable upgrade text for pre-releases

colinking:
For CLI pre-releases, we will no longer prompt the user to upgrade.

Ideally, we prompt the user if there is a newer pre-release, but we can't get the latest prerelease from GitHub without paging through the list of releases.


francescaleung (approved):
nice

PR #416, opened 2022-03-03
Always ask for an entrypoint

francescaleung:
Previously we were only prompting the user for an entrypoint if we didn't have one to use. Now we always prompt the user for an entrypoint. In order of decreasing priority, the default entrypoint is (a) the file provided to `airplane init`, if the file is not a task definition file, (b) the entrypoint specified on the existing task, or (c) `<slug>.<ext>` based on the task kind. For usages of `airplane init --from <slug>`, this gives users the ability to double-check the entrypoint; it also means that the given file is not ignored for usages of `airplane init --from <slug> <entrypoint>`.

The entrypoint prompt was tweaked so as to not imply that we're always going to create the entrypoint.

Also resolves AIR-3190.


pkg/cmd/tasks/initcmd/init.go
```diff
@@ -516,7 +520,7 @@ var orderedKindNames = []string{
 func promptForNewTask(file string, info *newTaskInfo) error {
 	defFormat := definitions.GetTaskDefFormat(file)
 	ext := filepath.Ext(file)
-	base := strings.TrimSuffix(file, ext)
+	base := strings.TrimSuffix(filepath.Base(file), ext)
```
leeweisberger, line 523: What's the point of this change?

leeweisberger (approved):

joshma:
1 - Does the user know what an "entrypoint" is?
2 - Are we OK with this breaking for non-interactive modes? IIUC `airplane init --slug=foo foo.js` allows you to init in a script whereas this would break that.

pkg/cmd/tasks/initcmd/init.go
```diff
@@ -449,26 +450,29 @@ func patch(slug, file string) (ok bool, err error) {
 	return
 }
 
-func promptForNewEntrypoint(slug string, kind build.TaskKind) (string, error) {
-	fileName := slug + runtime.SuggestExt(kind)
+func promptForEntrypoint(slug string, kind build.TaskKind, defaultEntrypoint string) (string, error) {
+	if defaultEntrypoint == "" {
```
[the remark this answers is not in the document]
francescaleung: I don't think `init` has ever worked in non-interactive terminals. which is to say--there's a path where there are no prompts to the user, & that case will obviously work, but I don't think `init` was ever specifically designed to work in non-interactive terminals.

I should check that `deploy` prompts work in non-interactive terminals though...

pkg/cmd/tasks/initcmd/init.go
francescaleung: This is for AIR-3190.

Before:
```
$ airplane init folder/main.ts
? What should this task be called? (folder/main)
```

After:
```
$ airplane init folder/main.ts
? What should this task be called? (main)
```

leeweisberger: yeah missed the linked issue!

PR #418, opened 2022-03-04
Upgrade lib + use GenerateCommentedFile in init

francescaleung:
Resolves AIR-3174.


leeweisberger (commented):
Why are we on a weird version of lib? Shouldn't we be on v0.3.9-0?

francescaleung:
we're on the fl-use-def03 branch of lib, because beta

leeweisberger (approved):

PR #420, opened 2022-03-07
Smoother experience for `init`

francescaleung:
Adds basic tab completion, validates for the right extension, & creates any necessary folders.

Resolves AIR-3224.


leeweisberger (changes requested):
We need to figure out the .ts and .js thing. We probably need to change SuggestExt to return a list of extensions in order of precedence

pkg/cmd/tasks/initcmd/init.go
```diff
@@ -567,31 +624,6 @@ func promptForNewTask(file string, info *newTaskInfo) error {
 		return errors.Errorf("Unknown kind selected: %s", selectedKindName)
 	}
 
-	// Ask for an entrypoint, maybe.
```
francescaleung, line 570: Since we now always prompt for an entrypoint when init'ing with a task definition, this is no longer needed here. This path doesn't get used when init'ing with `--code-only` because `--code-only` requires an existing task.

pkg/cmd/tasks/initcmd/init.go
```diff
@@ -525,6 +579,9 @@ func promptForNewTask(file string, info *newTaskInfo) error {
 		// Trim off the .task part, too
 		base = strings.TrimSuffix(base, ".task")
 	}
+	if base == "." {
+		base = ""
+	}
```
francescaleung, line 584: Noticed that `airplane init` was suggesting a default name of `.`.

leeweisberger (approved):

PR #422, opened 2022-03-08
Skip slack for Task Deployed segment event

rose-huang:
https://linear.app/airplane/issue/AIR-2926#comment-8351509b


pkg/analytics/analytics.go
```diff
@@ -99,7 +99,7 @@ type TrackOpts struct {
 
 // Track sends a track event to Segment.
 // event should match "[event] by [user]" - e.g. "[Invite Sent] by [Alice]"
-func Track(c *cli.Config, event string, properties map[string]interface{}) {
+func Track(c *cli.Config, event string, properties map[string]interface{}, opts ...TrackOpts) {
```
leeweisberger, line 102: This param isn't being used...

leeweisberger (changes requested):

pkg/analytics/analytics.go
rose-huang: 🤦🏻‍♀️

leeweisberger (approved):

PR #424, opened 2022-03-10
Implement airplane runtime httpd

ericliuche:
Adds the following functionality
```
  airplane httpdexec cmd [--port] [--host] [cmd_args...] [flags]
```
The `httpdexec` starts an http server that implements the airplane runtime. The airplane runtime is a lightweight server that communicates with a user command. The server supports the following
- a POST to `/` will execute the command, can optionally provide arguments for the command to run, and stream outputs e.g. `curl -i localhost:6000/ -X POST -d '{"args":["new echo", "1"], "env": {"TEST_ENV_VARIABLE":"hi"}}'`
- a POST to `/cancel` will SIGKILL the existing process if there is one

To see example behavior, start server with `go run ./cmd/airplane httpdexec --httpdhost localhost --debug pkg/cmd/httpd/test.sh`

Run ` curl -i localhost:6000/ -X POST -d '{}'` to see outputs with default arguments

Run ` curl -i localhost:6000/ -X POST -d '{"args":["new echo", "1"], "env": {"TEST_ENV_VARIABLE":"hi"}}'` to see outputs with provided arguments

Run ` curl -i localhost:6000/ -X POST -d '{}'` and then `curl -i localhost:6000/cancel -X POST` to see cancelled run


pkg/cmd/httpd/httpd.go
```diff
@@ -0,0 +1,81 @@
...
+	root *cli.Config
+	host string
+	port int
+	cmd  string
+	args []string
+}
+
+const (
+	defaultPort             = 6000
+	shutdownTimeoutDuration = 10 * time.Second
```
zhan-airplane, line 32: nit: I have a slight preference for keeping this `const` in the same file as where it's used, if possible

pkg/cmd/httpd/httpd.go
```diff
@@ -0,0 +1,81 @@
...
+
+	cmd := &cobra.Command{
+		Use:   "httpdexec cmd [--port] [--host] [cmd_args...]",
+		Short: "Start the Airplane runtime.",
+		Long:  "Start an http server that implements the Airplane runtime.",
+		Example: heredoc.Doc(`
+			airplane httpdexec ./helloworld.py [-- <parameters...>]
+			airplane httpdexec ./helloworld.py --port 5000 --host localhost [-- <parameters...>]
+		`),
+		PersistentPreRunE: utils.WithParentPersistentPreRunE(func(cmd *cobra.Command, args []string) error {
```
zhan-airplane, line 47: We don't actually need to be logged in for this right?

pkg/cmd/httpd/cmdexecutor.go
```diff
@@ -0,0 +1,131 @@
...
+	"os/exec"
+
+	"github.com/airplanedev/cli/pkg/logger"
+)
+
+// CmdExecutor handles the lifetime of an executed command.
+// The current active command, if there is one, is stored in `ActiveCmd`.
+// You are only allowed to run one command at a time. If the previous command is finished,
+// you are able to rerun.
+// NOCOMMIT should we add a unique exec id used to track command executions?
```
zhan-airplane, line 16: Eh we can add it when we need it right? Unless you have a use case for it right now

pkg/cmd/httpd/server.go
```diff
@@ -0,0 +1,178 @@
...
+		Msg:        msg,
+		Code:       code,
+		OutputType: OutputTypeStatus,
+	}); err != nil {
+		WriteHTTPError(w, r, err)
+	}
+}
+
+// GetOutputsHandler hangs forever and streams stdout/stderr/process statuses to the client.
+// NOCOMMIT check that hanging forever the right behavior
```
zhan-airplane, line 110: Hmm actually I was thinking that maybe we should just just combine the `getOutputs` endpoint with the `execute` endpoint - i.e. the execute endpoint doesn't return immediately, but sticks around and writes contents back until the process exits

pkg/cmd/httpd/server.go
```diff
@@ -0,0 +1,178 @@
...
+			select {
+			case stdout := <-cmdExecutor.Stdout:
+				WriteOutput(w, r, stdout, OutputTypeStdout)
+				flusher.Flush()
+			case stderr := <-cmdExecutor.Stderr:
+				WriteOutput(w, r, stderr, OutputTypeStderr)
+				flusher.Flush()
+			case exitStatus := <-cmdExecutor.ExitStatus:
+				WriteExit(w, r, exitStatus)
+				flusher.Flush()
```
zhan-airplane, line 129: Also, maybe we can return right after we receive the exit status? No point keeping things around after that right?

pkg/cmd/httpd/cmdexecutor.go
```diff
@@ -0,0 +1,131 @@
...
+type CmdExecutor struct {
+	ActiveCmd *exec.Cmd
+	// NOCOMMIT do we want to close / open channels depending on the state? e.g. close stdout/stderr channels
+	// when there isn't an active command?
+	Signal chan os.Signal
+	// NOCOMMIT should we combine stdout / stderr?
+	Stdout     chan string
+	Stderr     chan string
+	ExitStatus chan error
```
ericliuche, line 25: This could just all be combined into an Output message. I realized that right now the big issue with splitting them up is that the order of the lines can be mixed up

pkg/cmd/httpd/cmdexecutor.go
```diff
@@ -0,0 +1,131 @@
...
+
+	logger.Log("starting up: %s", cmd)
+	if err := cmd.Start(); err != nil {
+		return &StartCmdError{
+			Cmd: cmd,
+			Err: err,
+		}
+	}
+	c.ActiveCmd = cmd
+	go c.Wait()
```
zhan-airplane, line 115: If we merge the execute and get-outputs endpoints, we can do this synchronously instead of asynchronously (i.e. then we don't need to run this in a goroutine)

pkg/cmd/httpd/cmdexecutor.go
```diff
@@ -0,0 +1,131 @@
...
+			return
+		}
+	}
+}
+
+func readPipe(pipe io.ReadCloser, c chan string) {
+	scanner := bufio.NewScanner(pipe)
+	for scanner.Scan() {
+		// NOCOMMIT what happens if this channel is full?
+		c <- scanner.Text()
```
zhan-airplane, line 84: If this channel is full, the outputs will be left waiting on the pipe itself right? which I don't think should be a huge problem haha

pkg/cmd/httpd/cmdexecutor.go
```diff
@@ -0,0 +1,131 @@
...
+		done <- c.ActiveCmd.Wait()
+	}()
+
+	for {
+		select {
+		case signal := <-c.Signal:
+			logger.Log("recieved signal: %s", signal)
+			if c.ActiveCmd == nil || c.ActiveCmd.Process == nil {
+				// NOCOMMIT where should these errors go?
+				logger.Log("unable to signal, command already exited")
```
zhan-airplane, line 58: I think if the execute endpoint is synchronous it also answers this question - these errors can pass all the way back into the execute handler and we can deal with it then

pkg/cmd/httpd/cmdexecutor.go
```diff
@@ -0,0 +1,131 @@
...
+		case err := <-done:
+			// Clear the active command so we can run again.
+			c.ActiveCmd = nil
+			if err != nil {
+				logger.Log("finished command with error: %s", err.Error())
+			} else {
+				logger.Log("finished command with no error")
+			}
+			// NOCOMMIT what if this channel is full
+			c.ExitStatus <- err
```
zhan-airplane, line 74: Exit status should only pass back once per execution right? would be spooky if the channel gets full haha

pkg/cmd/httpd/cmdexecutor.go
```diff
@@ -0,0 +1,131 @@
...
+
+// CmdExecutor handles the lifetime of an executed command.
+// The current active command, if there is one, is stored in `ActiveCmd`.
+// You are only allowed to run one command at a time. If the previous command is finished,
+// you are able to rerun.
+// NOCOMMIT should we add a unique exec id used to track command executions?
+type CmdExecutor struct {
+	ActiveCmd *exec.Cmd
+	// NOCOMMIT do we want to close / open channels depending on the state? e.g. close stdout/stderr channels
+	// when there isn't an active command?
```
zhan-airplane, line 20: Hmm IMO if we combine the get-output and execute endpoints, then that answers this question more cleanly - we can just pass these channels in the args and not keep them around this global struct. We still need to keep `ActiveCmd` around for cancellation, but that's much more straightforward to think about

zhan-airplane (changes requested):
Overall, I think my main impression is that the code should end up being neater if we combine the execute and getoutputs endpoints (i.e. execute is long lasting and returns outputs over time, closing the connection when the command exits) - the execute can spin up goroutines for reading stderr/stdout maybe, but exit status can go in the main loop and also we keep a lot less strange global state around. I think it's OK to keep this long-lived connection around because it's intended to only go through localhost anyway

LMK if there's a reason why you didn't go with that route haha

pkg/cmd/httpd/cmdexecutor.go
zhan-airplane: In general you can't really determine the order of `stdout` and `stderr` reliably IIRC. Exit status should definitely go at the end though haha

pkg/cmd/httpd/cmdexecutor.go
ericliuche: this was in the case of being able to have multiple executions

zhan-airplane: Even if we have multiple executions, we should have different channels for each execution anyway

pkg/cmd/httpd/server.go
```diff
@@ -0,0 +1,162 @@
...
+		flusher, ok := w.(http.Flusher)
+		if !ok {
+			panic("expected http.ResponseWriter to be an http.Flusher")
+		}
+		w.Header().Set("Connection", "Keep-Alive")
+		w.Header().Set("X-Content-Type-Options", "nosniff")
+		go WriteOutput(w, r, flusher, stdoutPipe, OutputTypeStdout)
+		go WriteOutput(w, r, flusher, stderrPipe, OutputTypeStderr)
+
+		exitStatus, err := cmdExecutor.Wait()
```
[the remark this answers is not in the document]
zhan-airplane: Yeah I'm ok with it personally

pkg/cmd/httpd/cmdexecutor.go
```diff
@@ -0,0 +1,120 @@
...
+	if c.ActiveCmd == nil {
+		return &NoExistingCmdError{}
+	}
+	select {
+	case c.Signal <- os.Kill:
+	default:
+		logger.Log("already processing signal, discarding interrupt")
+	}
+	// NOCOMMIT: check if cmd is actually killed
+	// NOCOMMIT: should we try to kill if interrupt doesn't take down the process?
```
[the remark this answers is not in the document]
zhan-airplane: Also OK with punting checking if the cmd is actually killed to another PR

pkg/cmd/httpd/cmdexecutor.go
```diff
@@ -0,0 +1,120 @@
...
+			Cmd: cmd,
+			Err: err,
+		}
+	}
+	c.ActiveCmd = cmd
+	return nil
+}
+
+func (c *CmdExecutor) Cancel() error {
+	if c.ActiveCmd == nil {
```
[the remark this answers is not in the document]
ericliuche: ah yeah you're right. man i haven't used my multithreaded brain in so long lol

pkg/cmd/httpd/server.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+func WriteOutput(w http.ResponseWriter, r *http.Request, flusher http.Flusher, pipe io.ReadCloser, outputType OutputType) {
+	scanner := bufio.NewScanner(pipe)
+	for scanner.Scan() {
+		output := Output{
+			Msg:  scanner.Text(),
+			Type: outputType,
+		}
+		if err := json.NewEncoder(w).Encode(output); err != nil {
```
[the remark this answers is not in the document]
ericliuche: added a mutex to be sure

pkg/cmd/httpd/server.go
```diff
@@ -0,0 +1,169 @@
...
+
+func WriteOutput(w http.ResponseWriter, flusher http.Flusher, pipe io.ReadCloser, outputType OutputType, writeMutex *sync.Mutex) {
+	scanner := bufio.NewScanner(pipe)
+	for scanner.Scan() {
+		output := Output{
+			Msg:  scanner.Text(),
+			Type: outputType,
+		}
+		writeMutex.Lock()
+		if err := json.NewEncoder(w).Encode(output); err != nil {
```
ericliuche, line 113: So here, `WriteOutput` is called once for stderr, and once for stdout. The behavior i want to achieve is 
1. If either `WriteOutput` fails, e.g. `Encode` or `scanner` have an error, i want the other `WriteOutput` to also stop. I tried doing this with a `context` but had trouble writing this function in such a way that a `<-ctx.Done()` can stop it.
2. if either `WriteOutput` fails, I want to surface that error back to the client. right now the exist status is coming from `exitStatus := cmdExecutor.Wait()`, so somehow i need to get this error there, or approach it a different way. one way i tried doing this was creating a new `chan error` that `WriteOutput` could write to and `Wait` would consume. is that the right way?
3. it would be cool if anytime the request context `r.Context` finishes or `WriteOutput` errors, `Wait` would call `Cancel()` to terminate the child process and then pass on the `r.Context`/`WriteOutput` error in the ExitStatus, not sure how to accomplish that though :/

pkg/cmd/httpd/cmdexecutor.go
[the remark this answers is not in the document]
ericliuche: its also not too obvious to me the right way to do that, the naive way of 

```
1. c.Signal <- os.Interrupt
2. sleep(1 second)
3. return if process exists
4. c.Signal <- os.Kill
```

has the main flaw that (1) and (4) could possibly kill two different calls to `Execute` if timed poorly

pkg/cmd/httpd/server.go
```diff
@@ -0,0 +1,169 @@
...
+			http.Error(w, "invalid body", http.StatusBadRequest)
+			return
+		}
+
+		cmd := exec.Command(cmd, append(args, req.Args...)...)
+		cmd.Env = os.Environ()
+		for k, v := range req.Env {
+			cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", k, v))
+		}
+		stdoutPipe, err := cmd.StdoutPipe()
```
ericliuche, line 62: actually i think the right way to do this might just be to implement our own `io.Writer` and se it to `cmd.Stdout`

pkg/cmd/httpd/cmdexecutor.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+	}
+	c.ActiveCmd = cmd
+	return nil
+}
+
+func (c *CmdExecutor) Cancel() error {
+	c.activeCmdMutex.Lock()
+	if c.ActiveCmd == nil {
+		return &NoExistingCmdError{}
```
zhan-airplane, line 123: You need to unlock here when returning :P

pkg/cmd/httpd/cmdexecutor.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	c.activeCmdMutex.Unlock()
+	logger.Log("cancelling")
+	select {
+	case c.Signal <- os.Kill:
```
zhan-airplane, line 128: Actually now you mention it, there's a race condition here right? Like, we could accidentally execute an entirely new command and then have it be cancelled by a slow cancellation...

Maybe execution IDs is a good idea after all haha, like we'd generate one when we first call execute and return that, then if it's provided to the cancel handler we only cancel if it's the correct run. That might solve the broader problem too, where instead of having a global `CmdExecutor`, we have a map of ID -> `CmdExecutor` and then we don't have this strangeness around multiple runs (if the cancel handler doesn't receive an execution ID we just grab the first one in the map I guess)

pkg/cmd/httpd/server.go
```diff
@@ -0,0 +1,169 @@
...
+
+		w.Header().Set("Connection", "Keep-Alive")
+		w.Header().Set("X-Content-Type-Options", "nosniff")
+		w.Header().Set("Transfer-Encoding", "chunked")
+		var mutex sync.Mutex
+
+		go WriteOutput(w, flusher, stdoutPipe, OutputTypeStdout, &mutex)
+		go WriteOutput(w, flusher, stderrPipe, OutputTypeStderr, &mutex)
+
+		exitStatus := cmdExecutor.Wait()
```
zhan-airplane, line 92: I'm not entirely sure this is doing the right thing actually? According to the `exec` documentation (https://pkg.go.dev/os/exec#example-Cmd.StdoutPipe), `It is thus incorrect to call Wait before all reads from the pipe have completed`.

Looks like maybe you should be doing something like https://github.com/golang/go/issues/38268#issuecomment-609562062, where you have a waitgroup on the two `WriteOutput` goroutines, and only wait on the process once the two `WriteOutput` calls exit

pkg/cmd/httpd/server.go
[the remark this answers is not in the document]
zhan-airplane: Actually just remembered, Colin linked this to me:

> [Don't communicate by sharing memory, share memory by communicating.](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=2m48s)

Instead of having a mutex here, maybe the two sides should write to a shared channel and we should have a third goroutine reading off the end and sending outputs that way.

Would be nice if the cancellation thing could also work the same way haha, but I think if we go with a map of ID->executor approach we'd need a mutex on the top level, so oh well

zhan-airplane (changes requested):
Hmm meta comment: I realize we probably should have taken time to talk this through and sketched out how we want the different goroutines, etc. to work before we started implementing it haha, let's take note for next time we have something fiddly like this (I didn't realize there were this many pitfalls up front, my bad)

Anyway, I think there are two main high-level changes I'd like to make. LMK if you see any concurrency/etc issues with these, and then we can talk it through before actually coding haha to iron out bugs more quickly

1) I think the way the execute endpoint should be set up is, you set up three channels (`doneC`, `errC`, `logsC`), which are shared between two scanner goroutines. 
* The scanner goroutines call `scanner.Scan()` in a loop and write strings to `logsC` (this handles any messiness between `stderr` and `stdout` racing). If any of them throw an error, it writes to `errC` and exits. If they exit correctly, they write to `doneC`. 
* The main thread listens on all 3 channels simultaneously, and 1) if it receives something on `logsC`, it writes it via http, 2) if it receives something on `errC`, it kills the process and closes the pipes, 3) if it receives two messages on `doneC` (one from each scanner), it calls `Wait()` on the main `Cmd` and then pulls the exit code.

2) Let's assign an ID to each execution after all (good callout about this earlier, I was wrong haha; just generate it via UUID or something I guess), and store a map of ID to `CmdExecutor` (we should guard this map with a mutex).
* Cancellations that receive an ID cancel only that execution
* Cancellations that don't receive an ID pull the first execution off the map and cancel that.
* Let's not have restrictions on being only allowed to run one command at a time on this level - we'll handle this restriction on the runner level instead
* Signals we catch can be passed directly to every executor in the map. Cancellation signals just pass to one specific executor
* We can do the interrupt -> kill cancellation this way cleanly without any weird races.

pkg/cmd/httpd/server.go
```diff
@@ -0,0 +1,292 @@
...
+		w.Header().Set("Transfer-Encoding", "chunked")
+
+		outputC := make(chan Output, 4096)
+		errC := make(chan error)
+		outputDoneC := make(chan interface{}, 2)
+
+		// Write an output so the client knows the executionID.
+		outputC <- Output{
+			Msg:         "",
+			Type:        OutputTypeStdout,
```
ericliuche, line 143: should this be a different OutputType?

zhan-airplane: Yeah let's make it something different, like `OutputTypeMetadata` or `OutputTypeSystem` or something

pkg/cmd/httpd/server.go
```diff
@@ -0,0 +1,292 @@
...
+
+		// After setting these headers, we can't call `WriteHTTPError` anymore to write an error.
+		// Instead, we should write the error as a StatusOutput message.
+		w.Header().Set("Connection", "Keep-Alive")
+		w.Header().Set("X-Content-Type-Options", "nosniff")
+		w.Header().Set("Transfer-Encoding", "chunked")
+
+		outputC := make(chan Output, 4096)
+		errC := make(chan error)
+		outputDoneC := make(chan interface{}, 2)
```
zhan-airplane, line 138: Let's add some comments for what these channels are

pkg/cmd/httpd/server.go
```diff
@@ -0,0 +1,292 @@
...
+					}
+					return
+				}
+			}
+		}()
+
+		if err := cmdExecutor.Run(outputC, outputDoneC, errC, &ChunkedEncoder{
+			Encoder: json.NewEncoder(w),
+			Flusher: flusher,
+		}); err != nil {
```
zhan-airplane, line 164: Let's log this error too

zhan-airplane: Actually, can we return these errors via http too? So that our runner can deal with them

pkg/cmd/httpd/cmdexecutor.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// CmdExecutor handles the lifetime of an executed command.
+// The current active command, if there is one, is stored in `ActiveCmd`.
+// You are only allowed to run one command at a time. If the previous command is finished,
+// you are able to rerun.
+type CmdExecutor struct {
+	ActiveCmd      *exec.Cmd
+	SignalC        chan os.Signal
+	activeCmdMutex *sync.Mutex
```
zhan-airplane, line 26: We don't need this mutex any more now right?

zhan-airplane: Since it's not global any more, we don't even need to check if `ActiveCmd` exists when cancelling, we can just yolo and send the kill signal (I guess it won't return an error if it's not running, but I think that's ok - the end result state is the same anyway)

pkg/cmd/httpd/cmdexecutor.go
```diff
@@ -0,0 +1,121 @@
...
+			if err := c.ActiveCmd.Process.Signal(signal); err != nil {
+				return errors.Wrap(err, "unable to signal, command already exited")
+			}
+		case output := <-outputC:
+			err := encoder.Encode(output)
+			if err != nil {
+				return err
+			}
+		case <-outputDoneC:
+			logger.Log("closed output")
```
zhan-airplane, line 61: We probably can get rid of some of these log statements haha

pkg/cmd/httpd/cmdexecutor.go
```diff
@@ -0,0 +1,121 @@
...
+			outputDoneCount++
+			if outputDoneCount == cap(outputDoneC) {
```
zhan-airplane, line 63: We should document somewhere (maybe on the comment for this channel?) that we're expecting the cap to be the number of dones we're looking for

zhan-airplane (approved):

PR #426, opened 2022-03-14
Add source header from env var

leeweisberger:
Add source header so that we inform the API what context the CLI is running in. e.g. the GitHub action. We will add more sources in the future as we support running the CLI in more contexts (often different CI environments)


francescaleung (approved):
so normal cli will have ClientKind=cli, ClientSource=nil; github action will have ClientKind=cli, ClientSource=githubAction?

leeweisberger:
> so normal cli will have ClientKind=cli, ClientSource=nil; github action will have ClientKind=cli, ClientSource=githubAction?

yes. We could also do ClientSource=cli for normal cli, but that will be a no-op so decided not to do it

PR #428, opened 2022-03-17
Combine cmd and cmd_args into just cmd

ericliuche:
Having them separate makes it confusing with regards to `run.command` and `run.arguments`. 

@zhan-airplane we could also just pass everything in runtime, wdyt?


pkg/cmd/httpdexec/httpdexec.go
```diff
@@ -31,18 +30,15 @@ func New(c *cli.Config) *cobra.Command {
...
 	cmd := &cobra.Command{
-		Use:   "httpdexec cmd [--port] [--host] [cmd_args...]",
+		Use:   "httpdexec [--port] [--host] cmd...",
 		Short: "Start the Airplane runtime.",
 		Long:  "Start an http server that implements the Airplane runtime.",
 		Example: heredoc.Doc(`
-			airplane httpdexec ./helloworld.py [-- <parameters...>]
-			airplane httpdexec ./helloworld.py --port 5000 --host localhost [-- <parameters...>]
+			airplane httpdexec ./helloworld.py [-- <cmd...>]
+			airplane httpdexec ./helloworld.py --port 5000 --host localhost [-- <cmd...>]
```
zhan-airplane, line 38: Wait is this correct? Shouldn't this be like `airplane httpdexec --port 5000 --host localhost ./helloworld.py [parameters...]`?

zhan-airplane (approved):

joshma:
> @zhan-airplane we could also just pass everything in runtime, wdyt?

where did you two land on this? is the idea that run.command is hard-coded into the runner and run.arguments changes at runtime?

ericliuche:
@joshma 
` is the idea that run.command is hard-coded into the runner and run.arguments changes at runtime?`
yup!

we ended up deciding against passing everything at runtime due to possible security issues. having everything passed at runtime means the runner pod essentially has an open shell that can execute arbitrary commands from a `POST` to the execute endpoint. while it's true that 
1. a malicious actor that has access to the runner pod can already do much harm, e.g. making db queries
2. a malicious actor could create a task that is just `/bin/bash` which essentially turns the httpdexec process into a shell
we thought that having an extra layer of indirection is a bit safer. if ever users self host the runner pods, having an open shell accessible seems a bit scary, though if a malicious actor has access to the runner pods, then the situation is already quite bad

joshma:
got it, sounds good + thanks for the context!

PR #430, opened 2022-03-17
Add loader to stderrorlogger

leeweisberger:
Add the loader to the logger itself so we don't have to keep manually starting and stopping the loader. The loader automatically stops and restarts every time something is logged. This prevents the loader from appearing during a log.


pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -172,22 +170,15 @@ func run(ctx context.Context, cfg config) error {
...
-		isActive := loader.IsActive()
-		loader.Stop()
-
 		question := fmt.Sprintf("A task with slug %s does not exist. Would you like to create one?", def.GetSlug())
 		ok, err := utils.ConfirmWithAssumptions(question, cfg.assumeYes, cfg.assumeNo)
 
-		if isActive {
-			loader.Start()
-		}
-
```
francescaleung, line 190: The loader was removed here but I think this still needs it, as it's not using the logger

pkg/cmd/tasks/deploy/predeploy.go
```diff
@@ -61,7 +61,7 @@ func ensureConfigVarExists(ctx context.Context, client api.APIClient, params ens
 		if !utils.CanPrompt() {
 			return errors.Errorf("config %s does not exist", params.ConfigName)
 		}
-		logger.Log("Your task definition references config %s, which does not exist", logger.Bold(params.ConfigName))
+		l.Log("Your task definition references config %s, which does not exist", logger.Bold(params.ConfigName))
 		confirmed, errc := utils.Confirm("Create it now?")
```
francescaleung, line 65: I suspect this will also need the loader to be removed

pkg/logger/logger.go
```diff
@@ -25,20 +25,57 @@ type Logger interface {
...
 
 type StdErrLogger struct {
+	loader Loader
+}
+
+type StdErrLoggerOpts struct {
+	WithLoader bool
+}
+
+// NewStdErrLogger creates a new logger taht logs to stderr.
```
francescaleung, line 43:
```suggestion
// NewStdErrLogger creates a new logger that logs to stderr.
```

pkg/cmd/tasks/deploy/yaml.go
```diff
@@ -19,6 +19,8 @@ import (
 // DeployFromYaml deploys from a yaml file.
 func deployFromYaml(ctx context.Context, cfg config) (rErr error) {
 	client := cfg.client
+	l := logger.NewStdErrLogger(logger.StdErrLoggerOpts{WithLoader: true})
+	defer l.StopLoader()
```
francescaleung, line 23: there are other `logger.*` calls in this file that should probably be `l.*` calls instead

francescaleung (changes requested):
Looking good, just a few missed calls

pkg/cmd/tasks/deploy/deploy.go
leeweisberger: Manually stopped the loader before all calls to confirm. We probably also want to add this onto the logger, but didn't do that in this pr

pkg/api/client.go
```diff
@@ -102,10 +102,6 @@ type APIClient interface {
 	CreateDeployment(ctx context.Context, req CreateDeploymentRequest) (CreateDeploymentResponse, error)
 	CancelDeployment(ctx context.Context, req CancelDeploymentRequest) error
 	DeploymentURL(ctx context.Context, deploymentID string) string
-
-	GetBuild(ctx context.Context, id string) (res GetBuildResponse, err error)
-	GetBuildLogs(ctx context.Context, buildID string, prevToken string) (res GetBuildLogsResponse, err error)
-	CreateBuild(ctx context.Context, req CreateBuildRequest) (res CreateBuildResponse, err error)
```
leeweisberger, line 108: This is unrelated - we are no longer using remote.go or any code related to builds. Rather than upgrade it with the new logger loader I just deleted

francescaleung (approved):

PR #432, opened 2022-03-18
Upgrade github.com/containerd/containerd and github.com/xanzy/ssh-agent

joshma:
## Description
Ran `go get -u ...` to upgrade packages plus transitive dependencies

## Test plan
Test that builds still pass


joshma:
@zhan-airplane weirdly enough the snyk checks are now failing:
- license - these were existing license issues (MPL)
- security - this is...interesting because i think it's an issue w/ the version of containerd we're upgrading to. Transitive dependency. But I think it's still right to upgrade to containerd@v1.6.1 and upgrade again when deps have been fixed?

zhan-airplane (approved):
OOC, what are we planning to do with the existing license issues?

joshma:
Going to retry this PR

PR #433, opened 2022-03-18
Remove shell escape from httpdexec

zhan-airplane:
## Description

From my testing, I don't think we need to actually do the shell escape? Seems like go's `os/exec` package executes things via the `exec` system calls, not the shell

## Test plan
Run unit tests in `./pkg/httpd`


ericliuche (approved):

ericliuche:
Do we want to expand environment variables with `ExpandEnv`?

zhan-airplane:
I don't think we should? The args we're passing here generally shouldn't be affected by env variables

PR #437, opened 2022-03-23
Add initialization from examples

francescaleung:
## Description

```
$ airplane init --from-example github.com/airplanedev/examples/python/hello-world/python_hello_world.task.yaml@fl-update-examples
- Copied python_hello_world/main.py
- Copied python_hello_world/python_hello_world.task.yaml
- Copied python_hello_world/requirements.txt

⚡ To execute the task locally:
  airplane dev python_hello_world/python_hello_world.task.yaml

🛫 To deploy your task to Airplane:
  airplane deploy python_hello_world/python_hello_world.task.yaml
```

## Test plan

Test `airplane init --from-example ...` works, & that all examples deploy + execute.


pkg/cmd/tasks/initcmd/init.go
```diff
@@ -80,6 +82,7 @@ func New(c *cli.Config) *cobra.Command {
 
 	cmd.Flags().StringVar(&cfg.slug, "from", "", "Slug of an existing task to initialize.")
```
colinking, line 83: Unrelated, but since we hide `slug` not `from`, can we rename this struct field to `from`?

pkg/cmd/tasks/initcmd/init.go
```diff
@@ -101,10 +104,14 @@ func run(ctx context.Context, cfg config) error {
 	if cfg.assumeYes && cfg.assumeNo {
 		return errors.New("Cannot specify both --yes and --no")
 	}
-	if cfg.codeOnly && cfg.slug == "" {
+	if cfg.codeOnly && cfg.slug == "" && cfg.example == "" {
```
colinking, line 107: Nit: move this into the `cfg == ""` block. Won't need to check `cfg.example`

pkg/cmd/tasks/initcmd/example.go
```diff
@@ -0,0 +1,143 @@
...
+			if ok, err := utils.ConfirmWithAssumptions(question, cfg.assumeYes, cfg.assumeNo); err != nil {
+				return err
+			} else if !ok {
+				// bail here, nothing to do.
+				logger.Step("Initialization aborted.")
+				return nil
+			}
+		}
+
+		if err := copyFile(defPath, "."); err != nil {
```
colinking, line 55: It seems off that these paths have inconsistent copy behavior. This path copies the definition file into the current directory while the other path copies it into a new folder in your current working directory.

How about we merge the copy logic? Instead just split by `ErrNoEntrypoint` to compute a root folder (for `ErrNoEntrypoint` it would be the definition file's parent folder).

Then copy the **contents** of the root folder in the user's current directory.

colinking: > Then copy the contents of the root folder in the user's current directory.

As for this, could we ask the user where they want these files placed? We could default to the current working directory, but then let them provide a directory path. If this path doesn't exist, we can `mkdir -p` it.

Related to that, this command ignores the file path arg:

```
❯ lairlocal init --from-example github.com/airplanedev/examples/shell/hello-world/shell_hello_world.task.yaml ./here
? Folder shell_hello_world exists. Continue anyway? (Some files may get overwritten.) Yes
- Copied shell_hello_world/shell_hello_world.sh
- Copied shell_hello_world/shell_hello_world.task.yaml

⚡ To execute the task locally:
  airplane dev shell_hello_world/shell_hello_world.task.yaml

🛫 To deploy your task to Airplane:
  airplane deploy shell_hello_world/shell_hello_world.task.yaml
```

That could be the default for this prompt?

pkg/cmd/tasks/initcmd/example.go
```diff
@@ -0,0 +1,143 @@
...
+	src, err := ioutil.ReadFile(srcPath)
+	if err != nil {
+		return errors.Wrapf(err, "reading src file %s", srcPath)
+	}
+
+	dstPath := filepath.Join(dstDirectory, filepath.Base(srcPath))
+	if err := ioutil.WriteFile(dstPath, src, 0644); err != nil {
+		return errors.Wrapf(err, "writing dst file %s", dstPath)
+	}
+	logger.Step("Copied %s", dstPath)
```
colinking, line 119: Love it 🔥

pkg/cmd/tasks/initcmd/example.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+		localExecutionSupported = r.SupportsLocalExecution()
+
+		taskroot, err := r.Root(entrypoint)
+		if err != nil {
+			return err
+		}
+
+		dstDirectory := def.GetSlug()
+		if fsx.Exists(dstDirectory) {
```
colinking, line 74: Instead of checking the directory, what if we check for files that clash? For example, if the folder exists but is empty, we don't need to ask them.

This is likely more important if you copy files into the current working directory (from comment above).

colinking (approved):

pkg/cmd/tasks/initcmd/example.go
colinking: Thanks for the context! This makes more sense. I suppose task definition files for non-built tasks don't have nor need a concept of a root folder.

I mostly care that these experiences are consistent. For example, if I initialize a Node task and learn that `init --from-example` places it in a subfolder, then I would be surprised when I initialize a Docker task and it places it in my current directory. If we ask for the destination folder, then this would feel more consistent since the former could default to `./node_hello_world` while the latter could default to `./`. It might help to log if we are creating a file or a folder?

francescaleung: Here's what I'm going with:

* Dst directory defaults to slug if you don't specify a file. If you do specify a file, we try our best to figure out the nearest directory & use that instead.
* If the dst directory doesn't exist, we create it for you & tell you. If it does exist, we check it for collisions & prompt you to confirm whether or not that's okay. If there are no collisions, we don't prompt.
* All task files get copied into dst directory. Sometimes that means you only get one file copied in the directory.

I'm punting on prompting the user for a dst directory because I don't think it's super necessary. I like this being a one-step process in the best case scenario.

PR #438, opened 2022-03-23
Allow local execution via defn files of undeployed tasks

francescaleung:
## Description

Use task information from definition file if provided for local execution. This allows local execution of brand new (undeployed) tasks; it also allows testing changes to task metadata (e.g., parameters) before deploying.

## Test plan

* Test local execution of scripts associated with deployed tasks.
* Test local execution of task defn files with a brand new slug.

---

cc @colinking


pkg/cmd/tasks/dev/task.go
```diff
@@ -0,0 +1,78 @@
...
+		return taskInfo{}, err
+	}
+	defer dir.Close()
+
+	def, err := dir.ReadDefinition()
+	if err != nil {
+		return taskInfo{}, err
+	}
+
+	utr, err := def.GetUpdateTaskRequest(ctx, cfg.root.Client)
```
leeweisberger, line 43: kind of dumb we need to call this function to get the params :/

leeweisberger (approved):

pkg/cmd/tasks/dev/task.go
francescaleung: yeah... the other option is to add a GetParameters to the definition interface & thread that through everywhere, but if we only need it in this single spot, maybe this is fine?

PR #441, opened 2022-03-25
Add telemetry for airplane dev

leeweisberger:
Currently we have no way to know if users are using airplane dev


pkg/cmd/tasks/dev/dev.go
```diff
@@ -205,6 +206,17 @@ func run(ctx context.Context, cfg config) error {
 
 	print.Outputs(api.Outputs(o))
 
+	analytics.Track(cfg.root, "Run Executed Locally", map[string]interface{}{
```
leeweisberger, line 209: Our "task execute" event is `Run Executed` so I stuck with consistency

francescaleung (approved):

PR #451, opened 2022-04-13
Use wget to download airplane if curl isn't found

zhan-airplane:
## Description
Use wget in the install script if curl isn't found. Mainly because our stdapi images contain wget but not curl

## Test plan
Ran locally on my machine


scripts/install.sh
```diff
@@ -27,7 +27,14 @@ if [ ! -d "$bin_dir" ]; then
     mkdir -p "$bin_dir"
 fi
 
-curl --fail --location --progress-bar --output "$exe.tar.gz" "$download_uri"
+if command -v curl &> /dev/null; then
+  curl --fail --location --progress-bar --output "$exe.tar.gz" "$download_uri"
+elif command -v wget &> /dev/null; then
+  wget --show-progress -O "$exe.tar.gz" "$download_uri"
+else
+  echo "Not able to download the Airplane CLI - neither curl nor wget was found"
```
ericliuche, line 35: should we try to automatically install curl here?

ericliuche (approved):

scripts/install.sh
ericliuche: err sry i meant like prompt if they want it or at least link to where they could install it

PR #455, opened 2022-04-14
Add support for deploying task with config attachments

calufornia:
## Description
This PR adds support to the CLI to be able to deploy tasks with config attachments. If a `configs` field is specified in the task definition YAML, then those configs will get attached to the task - if no field is specified, then all configs will be wiped from the task. Previous CLI versions should be backwards compatible with this change since our update tasks endpoint uses patch semantics.

Resolves AIR-3456

## Test plan
`go run ~/airplane/cli/cmd/airplane --host api.airplane.so:5000 deploy ../sql_test/sql_task.task.yaml` with config attachments in YAML - verified that configs are attached through the UI.


pkg/cmd/tasks/deploy/predeploy.go
```diff
@@ -32,6 +32,8 @@ func ensureConfigVarsExist(ctx context.Context, client api.APIClient, def defini
 			}
 		}
 	}
+
+	// TODO(justin): Check that configs in config attachments exist
```
leeweisberger, line 36: Let's make sure we get to this! Not in this pr but in this project

leeweisberger (approved):

PR #456, opened 2022-04-14
Check that config var from config attachment exists before deploying

calufornia:
## Description
We should check that the config attachment in a task definition is a valid config - this PR adds that check.

## Test plan
Unit tests, test from CLI

task definition:
```
name: SQL Task
slug: sql_task
parameters:
- name: test/asdf
  slug: test_asdf
  type: shorttext
sql:
  resource: Airplane DB
  entrypoint: sql_task.sql
  queryArgs:
    sort_col: created_at
  configs:
    - db/dsn
    - fake_config
timeout: 3600
```

```
➜  cli git:(justin/check-config-exists) ✗ aplocal deploy ../sql_test/sql_task.task.yaml
Deploying 1 task:

sql_task
Type: sql
Root directory: ./../sql_test
URL: https://app.airplane.so:5000/t/sql_task

Your task definition references config fake_config, which does not exist
? Create it now? Yes
? Is this config a secret? No
? Value for fake_config test
  Setting fake_config to test...
  Done!
[deploy] Creating deployment...
```


ameyakhare (approved):
lgtm

PR #457, opened 2022-04-20
Enable creation of tasks on deploy without tty

francescaleung:
## Description

Creating tasks on deploy is gated by `utils.CanPrompt`, even with `--yes` specified. Fixed so that we check `cfg.AssumeYes` if `utils.CanPrompt` comes back false.

## Test plan

Run `nohup bash -c 'airplane tasks deploy --yes ./test_deploy.task.yaml'`. Deploy creates a task.


pkg/cmd/tasks/deploy/deploy.go
```diff
@@ -163,24 +163,24 @@ func run(ctx context.Context, cfg config) error {
...
-		if !utils.CanPrompt() {
-			return nil, nil
-		}
-
 		isActive := loader.IsActive()
 		loader.Stop()
 
-		question := fmt.Sprintf("A task with slug %s does not exist. Would you like to create one?", def.GetSlug())
-		ok, err := utils.ConfirmWithAssumptions(question, cfg.assumeYes, cfg.assumeNo)
+		if utils.CanPrompt() {
```
leeweisberger, line 169: If the terminal is not tty and you don't add --yes, this seems like it would auto create the task (which it shouldn't). It would not hit this if or the elseif and continue on to create the task.

Shouldn't this be `else if cfg.assumeNo || !cfg.assumeYes`?

leeweisberger (changes requested):

pkg/cmd/tasks/deploy/deploy.go
francescaleung: ahh good catch. I think it's safe as just `!cfg.AssumeYes`.

leeweisberger (approved):

PR #465, opened 2022-04-27
Fix deploying a task into a new environment for the first time.

ameyakhare:
Resolves AIR-3664

## Description
If you try to deploy a task into an environment that the task does not exist in, we fail on printing the diff between the old task def and the new task def. This catches that error so that we can proceed as normal.

## Test plan
Tested a deploy locally.


pkg/cmd/tasks/deploy/deploy_tasks.go
```diff
@@ -451,6 +451,10 @@ func (d *deployer) getDefinitionDiff(ctx context.Context, taskConfig discover.Ta
 		EnvSlug: d.cfg.envSlug,
 	})
 	if err != nil {
+		if _, ok := err.(*libapi.TaskMissingError); ok {
+			// The task is being promoted into a new environment, proceed as normal.
+			return []string{"(new task)"}, nil
```
francescaleung, line 456: Technically it's new _in this environment_, & not _new_ new. Maybe this wording is fine, just pointing this out.

francescaleung (approved):
Can you add a unit test for this case? & also remove the comment on L448, because apparently it's a lie?

francescaleung:
no unit test? :(

ameyakhare:
oh oops forgot to read that

PR #468, opened 2022-04-28
Fix httpd cancellation for processes that contain children

ericliuche:
## Description
Before, shell scripts were unable to be correctly terminated since they spawn child processes and we were only killing the parent process. This change changes the kill logic to kill the entire process group.

## Test plan
Ran locally with a shell script and confirmed it terminated correctly

### DIFF
--- a/pkg/httpd/cmdexecutor.go
+++ b/pkg/httpd/cmdexecutor.go
@@ -6,7 +6,6 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
-	"os/signal"
 	"sync"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


pkg/httpd/cmdexecutor.go
```diff
@@ -95,13 +95,12 @@ func (c *CmdExecutorManager) DeleteExecutor(executionID *string) error {
...
 	ActiveCmd      *exec.Cmd
 	ActiveCmdMutex *sync.Mutex
-	SignalC        chan os.Signal
+	SignalC        chan syscall.Signal
 	ExecutionID    string
 }
 
 func newCmdExecutor(executionID string) *CmdExecutor {
-	signalC := make(chan os.Signal, 1)
-	signal.Notify(signalC, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
```
ericliuche, line 104: We don't actually want to catch these signals. If the server gets a signal, we want to catch it and wait for the process instead of forwarding the child processes.

zhan-airplane (approved):

PR #470, opened 2022-04-28
Add methods to log with current time and use in httpd

ericliuche:
## Description
When looking at httpd logs, it's much more useful to have the time in the log.

## Test plan
Ran unit tests

### DIFF
--- a/pkg/cmd/serve/serve.go
+++ b/pkg/cmd/serve/serve.go
@@ -14,11 +14,12 @@ import (
 
 // Config is the httpd config.
 type config struct {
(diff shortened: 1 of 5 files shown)


joshma:
we already have time in the full log event and this should just be a formatting change

i don't want so many complicated logging methods

zhan-airplane (approved):

zhan-airplane (commented):
(+1 to josh's comment)

ericliuche:
moving runtime out of cli so this isnt needed

PR #473, opened 2022-05-02
Upgrade lib and clean up go.mod file

yolken-airplane:
## Description
This change updates the lib version used by our cli to pick up the changes made in https://github.com/airplanedev/lib/pull/120. It also cleans up the `go.mod` file so that direct dependencies are in the first `require (...)` block and indirect dependencies are in the second. This is how go does it for new projects, but if the go.mod was created with an earlier version of go, this file can have multiple require blocks over time.

cc: @calufornia 

## Test plan
Testing completed successfully via running locally, verifying that I could deploy tasks with both `runtime: durable` and `runtime` unset.


pkg/cmd/tasks/deploy/deploy_tasks_test.go
```diff
@@ -568,7 +631,8 @@ func TestDeployTasks(t *testing.T) {
 				RepoGetter:   &MockGitRepoGetter{Repo: tC.gitRepo},
 			})
 			for i, absEntrypoint := range tC.absoluteEntrypoints {
-				tC.taskConfigs[i].Def.SetAbsoluteEntrypoint(absEntrypoint)
+				err := tC.taskConfigs[i].Def.SetAbsoluteEntrypoint(absEntrypoint)
+				require.NoError(err)
```
ericliuche, line 635: nice

ericliuche (approved):

PR #421, opened 2022-03-08
Update test workflow and run on all pushes

colinking:
Noticed that tests were not running on stacked PRs. While fixing that, I updated our test workflow to match airport.

PR #474, opened 2022-05-03
Fix status watcher bug

colinking:
## Description

Fixes an issue with `Stopped` calculation caught by @francescaleung here: https://github.com/airplanedev/cli/pull/421/files#r863150865

Also fixes two output issues discovered while testing that bug:
1. Tabular outputs are automatically formatted (but shouldn't be)
2. Output was not printed by the tabular formatter if not a top-level array or object

## Test plan

Tested locally.


colinking:
* main
  * **PR #474** <a href="https://app.graphite.dev/github/pr/airplanedev/cli/474" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/cli/474?utm_source=stack-comment).

pkg/print/table.go
```diff
@@ -165,7 +165,7 @@ func (t Table) run(run api.Run) {
 // print outputs as table
 func (t Table) outputs(outputs api.Outputs) {
 	// Sort the output keys to match the UI.
-	switch t := ojson.Value(outputs).V.(type) {
```
colinking, line 168: Unnecessary cast

PR #480, opened 2022-05-11
Log non-latest on every command. Suggest upgrade.

leeweisberger:
## Description
On every Airplane command, log that the user is not using the latest version if they are not on the latest.

Suggest an upgrade command depending on their OS.

Linux:
image

Mac w/Airplane installed via homebrew:
image

## Test plan
Tested manually

---
I think there's a few smarter things we could do here:

one is [detect](https://apple.stackexchange.com/questions/284379/with-homebrew-how-to-check-if-a-software-package-is-installed) if they have Airplane installed via brew and leave that option out if it isn't installed.

We might also be able to save the installation route, but I'm not sure how to do that or if it's possible (any thoughts?)

Open to suggestions on UI and strategy here.
Re: AIR-3187. If users get rate limited, this won't show up. Not the end of the world, and would be fixed when we fix that issue

Re - this being too annoying, we could limit this to version that are a minor diff away. Pulumi does this on every command and I personally don't find it annoying, but I see how it could be

FIXES AIR-3312


pkg/version/latest/latest.go
```diff
@@ -41,9 +42,23 @@ func CheckLatest(ctx context.Context) {
...
+	}
+}
+
+func getUpgradeCommand() string {
+	os := runtime.GOOS
+	switch os {
+	case "windows":
+		return "iwr https://github.com/airplanedev/cli/releases/latest/download/install.ps1 -useb | iex"
+	case "darwin":
+		return "brew update && brew upgrade airplanedev/tap/airplane\n  --OR--\n  curl -L https://github.com/airplanedev/cli/releases/latest/download/install.sh | sh"
```
colinking, line 59: FYI you can use `heredocs.Docf` for multiline strings. It trims the leading space

pkg/cmd/root/root.go
```diff
@@ -71,6 +72,7 @@ func New() *cobra.Command {
 			// Log the version every time the CLI is run with `--debug`. This aligns
 			// customer debugging output with a specific release of the CLI.
 			logger.Debug(version.Version())
+			latest.CheckLatest(cmd.Context())
```
colinking, line 75: This blocks all CLI calls on a web request to check the latest version. I'm not sure how long this usually takes. Could you add in a debug line that prints how long it took to check the latest version? This would at least gives us context if we need to make this non-blocking / cached.

colinking (approved):
This is great. Thanks!

> one is detect if they have Airplane installed via brew and leave that option out if it isn't installed.

IIRC, deno's CLI persists this during installation

> We might also be able to save the installation route, but I'm not sure how to do that or if it's possible (any thoughts?)

We maintain some config here: `cat ~/.airplane/config`. We could add it there.

> Open to suggestions on UI and strategy here.

Ideally this is as minimal as possible. Could we collapse this down to ~2 lines in most cases? Let's drop the docs URL and merge a few lines. e.g.

```sh
$ airplane tasks list
[warning] A newer CLI version is available (v0.1.2 -> v0.1.3). To upgrade, run:
[warning]  curl -L https://...
```

It'd be nice to make just the `A newer ... v0.1.3).` text orange and the rest gray.

Could we also show the current version? Makes it easier to tell if the upgrade is important (e.g. patch vs. minor).

> Re - this being too annoying, we could limit this to version that are a minor diff away. Pulumi does this on every command and I personally don't find it annoying, but I see how it could be

We could look at the user's CLI's version (which has a timestamp) and use that to throttle the frequency/intensity of these messages? Maybe all gray if <1w and only a patch upgrade? But no strong opinions here, happy to try this and see.

leeweisberger:
> This is great. Thanks!
> 
> > one is detect if they have Airplane installed via brew and leave that option out if it isn't installed.
> 
> IIRC, deno's CLI persists this during installation
> 
> > We might also be able to save the installation route, but I'm not sure how to do that or if it's possible (any thoughts?)
> 
> We maintain some config here: `cat ~/.airplane/config`. We could add it there.
> 
> > Open to suggestions on UI and strategy here.
> 
> Ideally this is as minimal as possible. Could we collapse this down to ~2 lines in most cases? Let's drop the docs URL and merge a few lines. e.g.
> 
> ```shell
> $ airplane tasks list
> [warning] A newer CLI version is available (v0.1.2 -> v0.1.3). To upgrade, run:
> [warning]  curl -L https://...
> ```
> 
> It'd be nice to make just the `A newer ... v0.1.3).` text orange and the rest gray.
> 
> Could we also show the current version? Makes it easier to tell if the upgrade is important (e.g. patch vs. minor).
> 
> > Re - this being too annoying, we could limit this to version that are a minor diff away. Pulumi does this on every command and I personally don't find it annoying, but I see how it could be
> 
> We could look at the user's CLI's version (which has a timestamp) and use that to throttle the frequency/intensity of these messages? Maybe all gray if <1w and only a patch upgrade? But no strong opinions here, happy to try this and see.

Changed to 2 lines. Including the old and new version. I'm now detecting if mac users have Airplane installed via brew and giving brew instructions if that's the case.

PR #482, opened 2022-05-12
Tweak copy if entrypoint already exists

francescaleung:
## Description

If the entrypoint already exists, warns user that only existing Airplane scripts should be linked (if kind != SQL). If the user chooses to not link the file, we give them a chance to choose an alternate script file.

Resolves AIR-3688.

## Test plan

Tested locally.


pkg/cmd/tasks/initcmd/init.go
```diff
@@ -165,10 +165,28 @@ func initWithTaskDef(ctx context.Context, cfg config) error {
...
-			return err
+
+		for {
+			entrypoint, err = promptForEntrypoint(def.GetSlug(), kind, entrypoint)
+			if err != nil {
+				return err
+			}
+
+			if fsx.Exists(entrypoint) {
+				question := fmt.Sprintf("Are you sure you want to link %s? You should only existing Airplane scripts.", entrypoint)
```
leeweisberger, line 176: Does this mean already deployed airplane scripts? Why can't I link a JS file that is in the format of an airplane script? I don't have context - what are we trying to prevent here?

francescaleung:
Context [here](https://airplanedev.slack.com/archives/C01DYGVFUSJ/p1651593390190029?thread_ts=1651032141.562669&cid=C01DYGVFUSJ). You _can_ link any file, this is just tweaking the copy because we think the most likely case for this is already existing Airplane scripts.

pkg/cmd/tasks/initcmd/init.go
leeweisberger: Ok so we just want to make sure that `entrypoint` is an Airplane "script" and not just some arbitrary file. It doesn't have to be deployed - it just has to be of the correct format of an Airplane script file (e.g. for node, single default export, expects params, etc....)

leeweisberger (approved):

PR #485, opened 2022-05-13
Filter tasks with no root

leeweisberger:
## Description
Before, if you pass in `--changed-files`, tasks without roots (e.g. REST) would be included no matter what.

Now, we only include "rootless" tasks if the task definition file changed.

I made a very similar change in airport already. Porting that over here: https://github.com/airplanedev/airport/pull/2059

## Test plan
Unit test + manually tested


pkg/cmd/tasks/deploy/deployer.go
```diff
@@ -674,6 +690,24 @@ func containsFile(dir string, filePaths []string) (bool, error) {
 	return false, nil
 }
 
+// equalsFile returns true if the target file is equal to one of the files.
+func equalsFile(target string, files []string) (bool, error) {
```
francescaleung, line 694: `contains` is better than `equals`, I think.

pkg/cmd/tasks/deploy/deployer.go
```diff
@@ -674,6 +690,24 @@ func containsFile(dir string, filePaths []string) (bool, error) {
...
+	absTarget, err := filepath.Abs(target)
+	if err != nil {
+		return false, errors.Wrapf(err, "calculating absolute path of file %s", target)
+	}
+	for _, cf := range files {
```
francescaleung, line 699: what does `cf` stand for in this context?

francescaleung (approved):

PR #489, opened 2022-05-19
Send task schedules to api during deploy

ameyakhare:
Resolves AIR-3739

## Description
Sends the task schedules to the deployments/create endpoint in api: https://github.com/airplanedev/airport/pull/2357


pkg/cmd/tasks/deploy/deployer.go
```diff
@@ -341,6 +341,11 @@ More information: https://apn.sh/jst-upgrade`)
 		}
 	}
 
+	schedules := tc.Def.GetSchedules()
+	if len(schedules) == 0 {
+		schedules = nil
+	}
```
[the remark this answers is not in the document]
ameyakhare: oh wait I think I was trying to get the tests to pass since it depends on the struct field being nil vs being an empty map lol

francescaleung (approved):

pkg/print/types.go
```diff
@@ -34,6 +34,7 @@ type printTask struct {
 	Timeout                    int                       `json:"timeout" yaml:"timeout"`
 	IsArchived                 bool                      `json:"isArchived" yaml:"isArchived"`
 	InterpolationMode          string                    `json:"-" yaml:"-"`
+	Triggers                   []libapi.Trigger          `json:"triggers" yaml:"-"`
```
colinking, line 37: @ameyakhare @francescaleung -- don't set a value for the json/yaml tags (use `-`) unless you want the value to be printed out when a user specifies `-o json|yaml`. I'm going to revert this in a PR of mine: https://github.com/airplanedev/cli/pull/492/files#diff-25bb5aa5ed94afe01917db042ce914a311938634a4c037af1792c4c0f672b77dR37

And for values we do expose, they should be available to both JSON and YAML.

ameyakhare: sgtm 👍

PR #490, opened 2022-05-23
Prompt user to confirm deploy if there's a diff

leeweisberger:
Now that we are writing a task defn diff, I think it makes sense to prompt the user for confirmation before deploying. This is what I would expect with a system like this. This basically creates a "preview" mode where you can preview changes and then apply them.

If there is no task defn diff or the task is new, we don't prompt for confirmation.

---

## No change, no prompt:
image

## Changes - prompt:
image

Fixes AIR-3447


pkg/version/latest/latest.go
```diff
@@ -45,6 +45,7 @@ func CheckLatest(ctx context.Context) {
 	if latestWithoutPrefix != version.Get() {
 		logger.Warning("A newer CLI version is available (%s -> %s). To upgrade, run", version.Get(), latestWithoutPrefix)
 		logger.Log(logger.Gray("  " + getUpgradeCommand()))
+		logger.Log("")
```
leeweisberger, line 48: Unrelated - this came up in friday demo that we want a newline here

pkg/cmd/tasks/deploy/deployer.go
```diff
@@ -565,7 +578,7 @@ func (d *deployer) getDefinitionDiff(ctx context.Context, taskConfig discover.Ta
 	edits = gotextdiff.LineEdits(oldYAMLStr, edits)
 	diff := fmt.Sprint(gotextdiff.ToUnified(oldLabel, newLabel, oldYAMLStr, edits))
 	if diff == "" {
-		return []string{"(no changes)"}, nil
+		return []string{"(no changes to task definition)"}, nil
```
leeweisberger, line 581: Unrelated - I feel like this is more accurate. I would be confused if I updated my python script file and deployed and it said no changes

francescaleung (approved):

PR #491, opened 2022-05-25
Fix typo in copy

colinking:


pkg/cmd/tasks/initcmd/init.go
```diff
@@ -173,7 +173,7 @@ func initWithTaskDef(ctx context.Context, cfg config) error {
 			}
 
 			if fsx.Exists(entrypoint) {
-				question := fmt.Sprintf("Are you sure you want to link %s? You should only existing Airplane scripts.", entrypoint)
+				question := fmt.Sprintf("Are you sure you want to link %s? This script must follow Airplane's conventions.", entrypoint)
```
francescaleung, line 176: the convenient question here is, what are Airplane conventions?

colinking: Yeah, this phrasing might be a bit much. How about we drop this confirmation? I'll keep it around for SQL since that's destructive (though why is SQL different?).

francescaleung: [Context](https://airplanedev.slack.com/archives/C01DYGVFUSJ/p1651593390190029?thread_ts=1651032141.562669&cid=C01DYGVFUSJ) for adding the initial confirmation here. SQL has always had a confirmation precisely because it's destructive; this confirmation is to warn users that you can't just willy nilly link anything to a task & expect it to work.

francescaleung (approved):

PR #512, opened 2022-06-29
Fill in additional files needed for airplane views dev

zhan-airplane:
## Description
Putting up a PR first just to play with various settings + so it's a little easier to have something to talk about or dev on.

Basically, this expects an `App.tsx` file and optionally a `package.json` file (will patch a whole bunch of dependencies/etc on the `package.json`), and creates a temp dir with a symlink back to the original directory plus writes a whole bunch of other files out, then runs `npm install` and `npm run dev` in that directory. (handling vite config should be easy to tack on later)

There's a `viewdir.ViewDirectory` struct somewhere that when we figure out what the `.view.yaml` should contain, can do all the relevant parsing.

Some of the files here should probably be in `airplanedev/lib`, but I'm putting them here just so it's easy to play with them all in one place.

Also depends on airplanedev/views#61

## Test plan
Ran locally (`/home/zhan/cli/main --host api.airplane.so:5000 views dev --local-views /home/zhan/views`) in a directory containing an `App.tsx` copied from the views repo

---
cc @leeweisberger @stepzhou


zhan-airplane:
@stepzhou @leeweisberger FYI originally put rose as reviewer on this PR but adding the two of you as well since after talking we weren't sure about some of the vite stuff (e.g. whether all the stuff in `main.tsx` makes sense, whether the vite config is set up correctly)

cmd/airplane/views/dev/dev.go
```diff
@@ -107,7 +114,227 @@ const (
...
+		(*packageJSON)["scripts"] = map[string]interface{}{}
+		scripts = (*packageJSON)["scripts"].(map[string]interface{})
+	}
+	scripts["dev"] = "vite"
+
+	return nil
+}
+
+func runNPMInstall(tmpdir string) error {
+	cmd := exec.Command("npm", "install")
```
leeweisberger, line 312: We do want to determine if they're using npm or yarn.

We also don't currently enforce this today in tasks - we expect the user to run this on their own. I think I'm in favor of doing this

cmd/airplane/views/dev/dev.go
```diff
@@ -107,7 +114,227 @@ const (
...
+func patchPackageJSON(packageJSON *map[string]interface{}, cfg Config) error {
+	// TODO(zhan): fill out correct values.
+	deps, ok := (*packageJSON)["dependencies"].(map[string]interface{})
+	if !ok {
+		(*packageJSON)["dependencies"] = map[string]interface{}{}
+		deps = (*packageJSON)["dependencies"].(map[string]interface{})
+	}
+	deps["react"] = "18.0.0"
+	deps["react-dom"] = "18.0.0"
+	deps["@airplane/views"] = "*"
```
leeweisberger, line 284: I feel like it's ok to enforce that the user has this. Especially @airplane/views. We want the user to explicitly choose the version of the views lib as it will change - I don't think we can auto upgrade them

cmd/airplane/views/dev/dev.go
```diff
@@ -107,7 +114,227 @@ const (
...
+	if cfg.LocalViews != "" {
+		deps["@airplane/views"] = fmt.Sprintf("file:/%s", cfg.LocalViews)
+	}
```
leeweisberger, line 287: What is this for? Don't see how this is useful?

cmd/airplane/views/dev/dev.go
```diff
@@ -107,7 +114,227 @@ const (
...
+
+	devDeps, ok := (*packageJSON)["devDependencies"].(map[string]interface{})
+	if !ok {
+		(*packageJSON)["devDependencies"] = map[string]interface{}{}
+		devDeps = (*packageJSON)["devDependencies"].(map[string]interface{})
+	}
+	devDeps["@types/react"] = "^18.0.0"
+	devDeps["@types/react-dom"] = "^18.0.0"
```
leeweisberger, line 295: Adding these at dev time means you don't get ts support when you'r authoring, so I tend to think it's ok to enforce this as well.

cmd/airplane/views/dev/dev.go
```diff
@@ -107,7 +114,227 @@ const (
...
+	devDeps["@vitejs/plugin-react"] = "^1.3.0"
+	devDeps["typescript"] = "^4.6.3"
+	devDeps["vite"] = "^2.9.9"
+	devDeps["vite-tsconfig-paths"] = "^3.5.0"
```
leeweisberger, line 299: I would check to see if these are already in the user's package.json. we only want to add them if they don't already exist

cmd/airplane/views/dev/dev.go
```diff
@@ -107,7 +114,227 @@ const (
...
+
+	scripts, ok := (*packageJSON)["scripts"].(map[string]interface{})
+	if !ok {
+		(*packageJSON)["scripts"] = map[string]interface{}{}
+		scripts = (*packageJSON)["scripts"].(map[string]interface{})
+	}
+	scripts["dev"] = "vite"
```
leeweisberger, line 306: We don't need there to be a script called dev - we could just run `vite` without a script in the package.json. But I kind of like this change IF we only do this if the user doesn't include their own `dev` script. So if the user has a `dev` script, we call that. Else we replace it with the default `vite`

cmd/airplane/views/dev/dev.go
```diff
@@ -107,7 +114,227 @@ const (
...
+		import tsconfigPaths from "vite-tsconfig-paths";
+
+		export default defineConfig({
+			plugins: [react(), tsconfigPaths({ projects: [".", "./node_modules/@airplane/views/lib"] })],
+			envPrefix: "AIRPLANE_",
+			resolve: {
+				preserveSymlinks: true,
+				alias: {
+					"@airplane/views/dist": "@airplane/views/dist",
+					"@airplane/views": "@airplane/views/lib/src/index.ts",
```
leeweisberger, line 261: These guys we aren't going to need. We just have this in `apps` because it contains both the view and the lib in a yarn workspace.

cmd/airplane/views/dev/dev.go
```diff
@@ -107,7 +114,227 @@ const (
...
+	if err := runNPMDev(cfg, tmpdir); err != nil {
+		return errors.Wrap(err, "running npm dev")
+	}
+
+	return nil
+}
+
+func createWrapperTemplates(tmpdir string) error {
+	indexHtmlPath := filepath.Join(tmpdir, "index.html")
+	if err := os.WriteFile(indexHtmlPath, []byte(heredoc.Doc(`
```
leeweisberger, line 197: Generally like this + main.tsx

cmd/airplane/views/dev/dev.go
```diff
@@ -107,7 +114,227 @@ const (
...
+	  import.meta.env.AIRPLANE_API_HOST || "https://api.airplane.dev",
+		import.meta.env.AIRPLANE_TOKEN,
+		import.meta.env.AIRPLANE_API_KEY,
+		import.meta.env.AIRPLANE_ENV_SLUG,
+	);
+	ReactDOM.createRoot(document.getElementById("root")!).render(
+		<React.StrictMode>
+			<ThemeProvider>
+				<ViewProvider>
+					<Container size="xl" py={96}>
```
leeweisberger, line 232: This is pretty cool because we can completely control the chrome of the dev experience. Right now we just place the view in a container - in the future we might add more ux around it

cmd/airplane/views/dev/dev.go
```diff
@@ -60,6 +66,7 @@ func New(c *cli.Config) *cobra.Command {
 	}
 
 	cmd.Flags().StringVar(&cfg.EnvSlug, "env", "", "The slug of the environment to run the view against. Defaults to your team's default environment.")
+	cmd.Flags().StringVar(&cfg.LocalViews, "local-views", "", "If set, overrides the @airplane/views package with a local directory.")
```
leeweisberger, line 69: I don't think we need this - can you explain why you added?

leeweisberger (changes requested):
generally like the approach!

Let's discuss a bit what we want the user to have possible control over and what we want to take from them. I think we should go with a "overwrite only if the user has not specified their own way" approach. e.g. if the user has their own `dev` command, don't override with our own.

cmd/airplane/views/dev/dev.go
leeweisberger, line 312: Another issue here is that we actually need to write a package.json to the user's machine, or else this is going to take a long time every time we run it!

cmd/airplane/views/dev/dev.go
zhan-airplane: Ah it's mainly for our development - I was trying to find a way to use a local version of the views library rather than a downloaded one.

cmd/airplane/views/dev/dev.go
zhan-airplane: Ah I agree with your later comment about not overwriting if the users have provided it, but I don't think we should require users to have some version of `@airplane/views` set inside their `package.json` for this to work - it's quite possible for them to write a view that doesn't import `@airplane/views` at all, and they'd only need this because we're adding our own stuff inside `main.tsx`. In that case, it'd be fairly natural for someone to conclude that they should just delete it from their deps list.

cmd/airplane/views/dev/dev.go
zhan-airplane: I'm just going to take it out, and call vite from `node_modules/.bin/`

cmd/airplane/views/dev/dev.go
leeweisberger: Let's discuss in person! Basically yarn/npm install download a bunch of stuff to node_modules on the users machine and additionally cache stuff in files. We actually have to write files to the users machine, not just a temp dir that will be deleted. I'm actually not positive that's not what we're doing, but it seems like it is?

zhan-airplane:
@leeweisberger I decided to generate a consistent hash of the view directory struct to create a consistent temp dir that isn't deleted, in place of having some kind of `.airplane` cache dir in the original dir itself, to avoid weird symlinking cycle issues. Should still work basically the same way (i.e. as long as you only change code inside the view, rerunning dev will try to reuse the previously-generated temp dir if it exists, which'll effectively cache `node_modules` and everything else there)

cmd/airplane/views/dev/dev.go
```diff
@@ -107,17 +111,246 @@ const (
...
+func createWrapperTemplates(tmpdir string) error {
+	indexHtmlPath := filepath.Join(tmpdir, "index.html")
+	if err := os.WriteFile(indexHtmlPath, []byte(heredoc.Doc(`
+	<!DOCTYPE html>
+	<html lang="en">
+		<head>
+			<meta charset="UTF-8" />
+			<link rel="icon" type="image/svg+xml" href="/src/favicon.svg" />
+			<meta name="viewport" content="width=device-width, initial-scale=1.0" />
+			<title>TODO</title>
```
leeweisberger, line 211: Replace TODO with `Airplane` or something for now. Maybe the view slug if that's easy

cmd/airplane/views/dev/dev.go
```diff
@@ -107,17 +111,246 @@ const (
...
+		import.meta.env.AIRPLANE_TOKEN,
+		import.meta.env.AIRPLANE_API_KEY,
+		import.meta.env.AIRPLANE_ENV_SLUG,
+	);
+	ReactDOM.createRoot(document.getElementById("root")!).render(
+		<React.StrictMode>
+			<ThemeProvider>
+				<ViewProvider>
+					<Container size="xl" py={96}>
+						<Stack direction="column">
```
leeweisberger, line 240: I actually don't really think this is going to do anything b/c there is only one child (App). Let's delete

cmd/airplane/views/dev/dev.go
```diff
@@ -107,17 +111,246 @@ const (
...
+		"react":           "18.0.0",
+		"react-dom":       "18.0.0",
+		"@airplane/views": "*",
+	}
+	defaultDevDeps := map[string]string{
+		"@types/react":         "^18.0.0",
+		"@types/react-dom":     "^18.0.0",
+		"@vitejs/plugin-react": "^1.3.0",
+		"vite":                 "^2.9.9",
+		"vite-tsconfig-paths":  "^3.5.0",
```
leeweisberger, line 297: We aren't using this, delete

leeweisberger (approved):
Let's start getting this checked in so we can iterate
1. We want to move most of this to lib
2. We want to pull out the shim stuff into its own .tsx files so they're easier to develop
3. Is there a good way to manage our dependency versions (e.g. vite). Can we use renovate somehow?
4. We are going to have to make a bunch of changes to ensure this works correctly with yarn workspaces (can probably ignore for now)

I don't loooove hiding files in some tmp directory - would rather be explicit about where we are putting this stuff to a) avoid writing to random parts of the user's system and b) so there's less magic. We can do this for now / open to discuss it.

PR #514, opened 2022-07-01
Add local API server for airplane dev

calufornia:
## Description

Sets up the local API server for `airplane dev` - right now, the API server lasts for the existence of the dev command, but in the future the server will be started/stopped with separate commands. Right now, the API server exposes 3 endpoints required by our SDKs:

- `/v0/tasks/execute`
- `/v0/runs/get`
- `/v0/runs/getOutputs`

as well as a `/shutdown` endpoint that will be used to terminate the API server in the future.

The handlers for these methods currently accept and return the actual request/response objects, respectively, but do not contain any substantive implementation - this will be included in subsequent commits.

Currently, we only start the local API server for workflow tasks while it's still under development - in the future, we don't necessarily need this restriction.

## Test plan
Manually add sleep and start server: `aplocal dev ../node_test/node_test.task.yaml`

```
> curl -X POST "localhost:7190/v0/tasks/execute"
{"runID":"runv4dxzbmnpr"}
> curl "localhost:7190/v0/runs/get?id=run1234"
{"id":"run1234"}
> curl "localhost:7190/v0/runs/getOutputs?id=run1234"
{"output":null}
> curl '127.0.0.1:7190/shutdown'
OK%
> curl '127.0.0.1:7190/shutdown'
curl: (7) Failed to connect to 127.0.0.1 port 7190 after 5 ms: Connection refused
```


pkg/server/server.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/airplanedev/cli/pkg/logger"
+	"github.com/gorilla/mux"
+)
+
+const port = 7190
```
kellymtrinh, line 12: I think this is fine to start- should we use an env variable OR the default so people can override it if they already have something running on that port?

kellymtrinh (approved):

pkg/server/server.go
calufornia: Sounds good, I'll make it a flag you pass to `airplane dev` for now.

PR #517, opened 2022-07-05
Create tasks gentypes command

leeweisberger:
## Description
Create a new command `airplane tasks gentypes` that generates TypeScript types (in the future we could expand to other languages) for Airplane tasks.

If you provide no arguments, we generate a types file for all Node.js tasks. You can also provide a single task and we generate types for that task.

Two improvements we may want to make (I did not in this pr) is to 1) generate types for a task that has not been deployed yet and 2) generate types into the script file rather than an external file.

This command will be very helpful for Views (and probably also Workflows) where we want to know the type of various tasks.

depends on https://github.com/airplanedev/lib/pull/172

## Test plan
I manually ran this and generated a types file. Also pretty solid unit tests.


cmd/airplane/tasks/initcmd/init.go
```diff
@@ -645,8 +647,8 @@ func cwdIsHome() (bool, error) {
 	return cwd == home, nil
 }
 
-func createEntrypoint(r runtime.Interface, entrypoint string, task *libapi.Task) error {
-	code, fileMode, err := r.Generate(apiTaskToRuntimeTask(task))
+func createEntrypoint(r runtime.Interface, entrypoint string, task *libapi.Task, withComment bool) error {
+	code, fileMode, err := r.Generate(apiTaskToRuntimeTask(task), runtime.GenerateOpts{GenerateComment: withComment})
```
colinking, line 651: 🔥

pkg/cmd/tasks/gentypes/gentypes.go
```diff
@@ -0,0 +1,140 @@
...
+	taskSlug string
+	envSlug  string
+	fileName string
+}
+
+func New(c *cli.Config) *cobra.Command {
+	var cfg = config{root: c}
+
+	cmd := &cobra.Command{
+		Use:   "gentypes slug",
```
colinking, line 36: Could we call this `generate` instead? Maybe with an alias `gen` for conciseness.

colinking: FWIW, I expect we'll generate runtime code at some point (e.g. logic to read resources from env vars).

pkg/cmd/tasks/gentypes/gentypes.go
```diff
@@ -0,0 +1,140 @@
...
+		Short: "Generates types for Node.js task parameters",
```
colinking, line 37: Would be nice to support Python as it has types built-in nowadays (cc @ameyakhare). For now, let's remove the reference to Node.js here and assume we'll add a Python implementation at some point.

```suggestion
		Short: "Generates types for task parameters",
```

pkg/cmd/tasks/gentypes/gentypes.go
```diff
@@ -0,0 +1,140 @@
...
+	return cmd
+}
+
+func run(ctx context.Context, cfg config) error {
+	fileName := cfg.fileName
+	if !strings.HasSuffix(fileName, ".ts") {
+		fileName = fmt.Sprintf("%s.ts", fileName)
+	}
+	if fsx.Exists(fileName) {
+		if ok, err := utils.Confirm(fmt.Sprintf("Would you like to overwrite %s?", fileName)); err != nil {
```
colinking, line 69: nice

pkg/cmd/tasks/gentypes/gentypes.go
```diff
@@ -0,0 +1,140 @@
...
+			return err
+		} else if !ok {
+			return nil
+		}
+	}
+	wr, err := os.Create(fileName)
+	if err != nil {
+		errors.Wrapf(err, "creating file %s", fileName)
```
colinking, line 77: ...did our linter not error on this? hmm

```suggestion
		return errors.Wrapf(err, "creating file %s", fileName)
```

pkg/cmd/tasks/gentypes/gentypes.go
```diff
@@ -0,0 +1,140 @@
...
+		taskSlug: cfg.taskSlug,
+		wr:       wr,
+	}); err != nil {
+		return err
+	}
+
+	if fileName != "" {
+		logger.Log("Generated TypeScript types at %s", fileName)
+	} else {
+		logger.Log("Generated TypeScript types")
```
colinking, line 90: Let's still print the file name here. If they didn't provide one, they won't know where the types are going to be produced after all.

pkg/cmd/tasks/gentypes/gentypes.go
```diff
@@ -0,0 +1,140 @@
...
+
+	if err := genTypeScriptTypes(ctx, cfg.root.Client, cfg.envSlug, genTypeScriptTypesOpts{
+		taskSlug: cfg.taskSlug,
+		wr:       wr,
+	}); err != nil {
+		return err
+	}
+
+	if fileName != "" {
+		logger.Log("Generated TypeScript types at %s", fileName)
```
colinking, line 88: Minor, but could you make the file name bold white? So it sticks out.

pkg/cmd/tasks/gentypes/gentypes_test.go
```diff
@@ -0,0 +1,111 @@
...
+					Slug: "my_task",
+					Name: "My Task",
+					Kind: build.TaskKindNode,
+					Parameters: libapi.Parameters{
+						{Slug: "int", Type: libapi.TypeInteger},
+						{Slug: "float", Type: libapi.TypeFloat},
+						{Slug: "string", Type: libapi.TypeString},
+						{Slug: "date", Type: libapi.TypeDate},
+						{Slug: "upload", Type: libapi.TypeUpload},
+						{Slug: "date-time", Type: libapi.TypeDatetime},
```
colinking, line 39: Invalid slug

```suggestion
						{Slug: "date_time", Type: libapi.TypeDatetime},
```

pkg/cmd/tasks/gentypes/fixtures/multipletasks.ts
```diff
@@ -0,0 +1,8 @@
+// Code generated by airplane tasks gentypes. DO NOT EDIT.
+
+export type MyTaskParams = {
+  int: number;
+};
```
colinking, line 5: Can you include a newline between params? 🙏🏻

pkg/cmd/tasks/gentypes/gentypes.go
```diff
@@ -0,0 +1,140 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) > 0 {
+				cfg.taskSlug = args[0]
+			}
+			return run(cmd.Root().Context(), cfg)
+		},
+	}
+
+	cmd.Flags().StringVar(&cfg.envSlug, "env", "", "The slug of the environment to query. Defaults to your team's default environment.")
+	cmd.Flags().StringVar(&cfg.fileName, "file", "types.ts", "The file to save the generated types to.")
```
colinking, line 58: Assuming we eventually add Python support, this default value won't make sense. Maybe we drop the `--file` param entirely and just dump to stdout?

So you would do `airplane tasks gen > types.ts` instead.

pkg/cmd/tasks/gentypes/gentypes.go
```diff
@@ -0,0 +1,140 @@
...
+			Slug:    opts.taskSlug,
+			EnvSlug: envSlug,
+		})
+
+		if err != nil {
+			return err
+		}
+		tasks = append(tasks, task)
+	} else {
+		tasksResp, err := apiClient.ListTasks(ctx, envSlug)
```
colinking, line 113: IMO, it would be more intuitive if this worked like `airplane deploy` and generated for all tasks in the current directory.

Same goes with accepting a YAML file path instead of a slug.

colinking: I guess this has different semantics from `deploy` because users would use this to get type definitions for other tasks they can call?

If that's the primary use-case, I'd start thinking of this CLI command as generating a type-safe Airplane API client. At that point, it'd be nicer if this was more tightly integrated with the SDK. For example, if I call `airplane.execute("my_task")`, I'd want the return type to automatically have the correct types for `my_task`.

I was thinking about something similar within the context of resources. I want to generate runtime code + types that would allow you to access connected resources as `airplane.resources.db` (and TS would know `db` is, say, a PostgreSQL resource because of the types we generate from your task definition). We could do this code generation automatically during `airplane deploy`/`airplane dev` (and expose a `airplane gen` that would do this for you one-off). Possibly we do it similar to Prisma and produce the generated code into `node_modules`.

colinking: If we tie this to the Airplane SDK, then I'd want the code generator to be part of the SDK. Otherwise, the code generator can fall out of line with the user's SDK version.

colinking (changes requested):
LGTM, but I'd like to talk through how this will be used by users. Specifically, see my comment below about how this ties to the SDK.

leeweisberger:
Gonna close. We'll get back to this soon.

PR #519, opened 2022-07-06
Add ability to run child tasks from airplane dev

calufornia:
## Description

This PR allows `airplane dev` on tasks that call `airplane.execute` to actually execute a child task locally. This is done by augmenting the `v0/tasks/execute` endpoint of the local api server to execute the child task's code in another process. Tasks are referenced by slug, and we discover tasks using the same task discovery logic when deploying, except we don't check if the task actually exists or is archived. This PR also moves much of the `dev` execution logic into its own package in `pkg` so that it can be used by both the CLI command as well as the local api server.

TODO: Make logging more clear so that the user knows which logs apply to which task (the child task or the parent task). Also allow the user to configure which tasks are registered locally instead of auto-discovering them.

## Test plan
Created a task that called another task:
```
import airplane from "./index.ts";

// Put the main logic of the task in this function.
export default async function (params) {
  return await airplane.execute("node_task");
}
```
And then ran:
```
> go run ~/airplane/cli/cmd/airplane
aplocal dev ../node-sdk/durable_node_test.task.yaml -- name=justin
Locally running Workflow Node task using airplane args 7 task (http://localhost:7190/t/workflow_node_task)

Locally running standard node task 5 task (http://localhost:7190/t/node_task)

[log] Hello world
[log] airplane_output_set 5
5
[log] airplane_output_set {"id":"run3a8e4u95hs","status":"Succeeded","output":null}

id
run3a8e4u95hs

status
Succeeded

output
```


pkg/api/client.go
```diff
@@ -525,6 +525,13 @@ func (c Client) host() string {
 	return Host
 }
 
+func (c Client) scheme() string {
+	if strings.HasPrefix(c.Host, "localhost") {
```
kellymtrinh, line 529: Would it be easier to store all the urls as `url.URL{}` objects and only  call `u.String()` when it's about to be called? Then we wouldn't need to parse strings as much

calufornia: Just to confirm, is the suggestion to replace the `appUrl` method with field on the Client struct that stores the url? That sounds reasonable to me, could probably be included in a follow-up PR

pkg/dev/dev.go
```diff
@@ -0,0 +1,251 @@
...
+
+	cmd.Env = append(cmd.Env, fmt.Sprintf("AIRPLANE_API_HOST=http://localhost:%d", config.Port))
+	cmd.Env = append(cmd.Env, "AIRPLANE_RUNTIME=dev")
+
+	if err := cmd.Start(); err != nil {
+		return errors.Wrap(err, "starting")
+	}
+
+	// mu guards o and chunks
+	var mu sync.Mutex
```
kellymtrinh, line 117: I realize this isn't your change (?) but maybe we want to defer Unlock ? 
It can be helpful to make this into a struct containing the ojson.Value, chunks, and lock if it's locking on both

calufornia: Ah, good point - could put everything in the for loop into a separate function, and then pass the mutex to it. Will add a TODO

PR #520, opened 2022-07-06
Support deployment of node code-only configured tasks

ameyakhare:
## Description
Adds support for code-only configured tasks by adding the code_task_discoverer to the list of existing discoverers. Also tweaks the discoverer api to allow extracting multiple task configs from the same file.

## Test plan
Tested deployment locally.


parser/node/airplaneTask.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export type UserFunc<TParams extends Params> = (params: ParamValues<TParams>) => any
+
+export type AirplaneFunc<TParams extends Params> = (params: ParamValues<TParams>) => Promise<Awaited<ReturnType<UserFunc<TParams>>>>
+
+export function Task<TParams extends Params>(
+    config: TaskConfig<TParams>,
+    f: UserFunc<TParams>
+): AirplaneFunc<TParams> {
+    const inAirplaneRuntime = !!(process.env.AIRPLANE_RUN_ID ?? "");
```
colinking, line 30: Let's use `AIRPLANE_RUNTIME` to check instead.

ameyakhare: problem is a valid value for AIRPLANE_RUNTIME is empty, wouldn't this fail?

colinking (approved):

PR #521, opened 2022-07-07
Create views entrypoint file during init

rose-huang:
## Description
We want to eventually hook the entrypoint up to a task but this is the first pass at having some basic view. 

The file, when hooked up with the right dependencies, will create a default view that looks like this:
image

## Test plan
```
➜  test ../airplanedev/cli/bin/airplane views init   
? What should this view be called? view with entrypt
- Created view definition at view_with_entrypt/view_with_entrypt.view.yaml
- Created view entrypoint at view_with_entrypt/view_with_entrypt.view.tsx
```

image
Note: need to update the comment around entrypoint in the view def but this is in lib, not cli so will update this later


cmd/airplane/views/initcmd/init.go
```diff
@@ -121,3 +133,58 @@ func createViewDefinition(cfg config) error {
 	logger.Step("Created view definition at %s", defnFilename)
 	return nil
 }
+
+func createEntrypoint(cfg config) error {
+	entrypointPath := generateEntrypointPath(cfg, false)
+
+	if err := ioutil.WriteFile(entrypointPath, []byte(heredoc.Doc(`
+import { Table, Column, Title, Text } from "@airplane/views";
```
rose-huang, line 141: may need to use templates later if we generate any custom comments in the future but this works for now

cmd/airplane/views/initcmd/init.go
```diff
@@ -94,9 +94,21 @@ func createViewScaffolding(cfg *config) error {
 	if err := createViewDefinition(*cfg); err != nil {
 		return err
 	}
+	if err := createEntrypoint(*cfg); err != nil {
+		return err
+	}
 	return nil
 }
 
+func generateEntrypointPath(cfg config, relativeToViewDef bool) string {
```
leeweisberger, line 103: relativeToViewDef is more of an example of when you'd set this param than what the param should be called. Maybe call it `nested` or `inViewDir` or something

cmd/airplane/views/initcmd/init.go
leeweisberger: Pull this into it's own file. Make the extension of the file `.jsx`. Then use [embed](https://pkg.go.dev/embed) to pull it in here.

That will give us syntax highlighting / possible linting / possible testing

leeweisberger (changes requested):

leeweisberger (approved):

PR #524, opened 2022-07-11
Handle latest init in `airplane views dev`

zhan-airplane:
## Description
Some changes to handle the newest init scaffolding:
* You can run both `airplane views dev <folder>` and `airplane views dev <file>` - it'll assume you're talking about the folder containing the file right now (together with the other changes around root dir detection, etc, should be a quick hack to handle the `airplane views dev blah.view.yaml` case, might need to revisit this when we need to handle multiple views in a project)
* Searches for a `package.json` in the path up recursively, uses that as a root dir
* Searches the entire root dir for views, dev only works if there is exactly one view in the whole root dir
* Parses the `.view.yaml` file for the correct entrypoint
* Moves a bunch of files into go embeds, for easier development

## Test plan
Ran `airplane views init` then `airplane views dev`.

---
cc @stepzhou @leeweisberger


cmd/airplane/views/dev/dev.go
```diff
@@ -69,25 +71,35 @@ func New(c *cli.Config) *cobra.Command {
...
+}
+
+const (
+	hostEnvKey    = "AIRPLANE_API_HOST"
+	tokenEnvKey   = "AIRPLANE_TOKEN"
+	apiKeyEnvKey  = "AIRPLANE_API_KEY"
+	envSlugEnvKey = "AIRPLANE_ENV_SLUG"
+)
+
+func findRoot(fileOrDir string) (string, error) {
```
rose-huang, line 88: I was trying out views dev and my directory structure is as follows:

/testtest
  - package.json
  - /test2
    - test2.view.tsx
    - test2.view.yaml
   
Was expecting the root to be testtest for both because that's the directory containing my package.json but it doesn't seem to be the case if I dev by passing in the yaml file, do you know why?


I'm in the test2 directory and calling:
```
➜  test2 ~/airplanedev/cli/bin/airplane views dev 
Using /Users/rosehuang/test/testtest as root directory
temporary dir /var/folders/wb/13trvdqd0qjfbgbpbl2_cq6m0000gn/T/airplane-view-17685415050690606899 exists
...
➜  test2 ~/airplanedev/cli/bin/airplane views dev test2.view.yaml
Using . as root directory
temporary dir /var/folders/wb/13trvdqd0qjfbgbpbl2_cq6m0000gn/T/airplane-view-17685415050690606899 exists

```

rose-huang (changes requested):
> Searches the entire root dir for views, dev only works if there is exactly one view in the whole root dir

Can we make `airplane views dev <VIEW YAML FILE>` work to dev a single view even if there is more than one view in the whole root dir? with the way we have encourage people to share package.json across views, they will probably have a shared root for multiple views


Agree that for `airplane views dev`  (which runs dev in current directory) and `airplane views dev <DIRECTORY>`, we should throw this error

cmd/airplane/views/dev/templates/main.tsx
```diff
@@ -0,0 +1,22 @@
+import { Container, Stack, ThemeProvider, ViewProvider, setEnvVars } from "@airplane/views";
```
rose-huang, line 1: doesn't seem like we're using Stack, let's remove

cmd/airplane/views/dev/dev.go
zhan-airplane: Ugh yeah let me debug this

zhan-airplane:
@rose-huang fixed the root dir bug (`fsx.Find` on non-absolute path didn't work the way I thought) and also if you dev explicitly with a views definition, it'll use that as the view instead of searching

rose-huang (approved):
cool looks great! could you also update the patched in package.json to use the views package that stephen published if you haven't done that in another PR? or feel free to do it in a follow up PR

PR #525, opened 2022-07-12
Add implementation for getOutputs and improve logging

calufornia:
## Description

This PR adds an implementation for the `/v0/runs/getOutputs` endpoint of the `airplane dev` api server. Previously, the endpoint be a no-op, which meant that users wouldn't be able to test utilizing the output from a child task in the current task. We change it so that we actually store the output (in-memory) in the local API server, so that `getOutputs` can return the appropriate output for a given run. Additionally, we improve the logging so it's more clear which task is currently being executed, now that we support locally executing other tasks.

## Test plan
Node task 1:
```
export default async function (params) {
  return await airplane.execute("node_task_2");
}
```

```
aplocal dev ./node_task_1.task.yaml

+------------------------------------------------------------------------+
| Locally running task [Node task 1] http://127.0.0.1:7190/t/node_task_1 |
+------------------------------------------------------------------------+

+------------------------------------------------------------------------+
| Locally running task [Node task 2] http://127.0.0.1:7190/t/node_task_2 |
+------------------------------------------------------------------------+

[Node task 2 log] airplane_output_set 12345

Output for task node_task_2:
12345

+-------------------------------------+
| Finished running task [Node task 2] |
+-------------------------------------+

[Node task 1 log] airplane_output_set {"id":"run5ue07afzib","status":"Succeeded","output":12345}

Output for task node_task_1:

id
run5ue07afzib

status
Succeeded

output
12345

+-------------------------------------+
| Finished running task [Node task 1] |
+-------------------------------------+
```


pkg/print/print.go
```diff
@@ -74,6 +76,19 @@ func Print(obj interface{}, defaultPrintFunc func()) {
 	}
 }
 
+// BoxPrint pretty prints a box around the given string - for example, BoxPrint("hello") would output
+// +-------+
+// | hello |
+// +-------+
+func BoxPrint(s string) {
```
kellymtrinh, line 83: 😍

kellymtrinh (approved):

PR #530, opened 2022-07-14
Add ability to reference local resources for airplane dev

calufornia:
## Description

This PR adds the notion of an airplane dev config file, which will be used to store local resource configuration. The dev config will store a mapping from slug to resource configuration - this is because the task definition file stores resource attachments in the form of a map between alias to slug, and so combining the two we have a mapping from alias to resource configuration, which is what is required by our SDK. We add logic to encode resource configs into an `AIRPLANE_RESOURCES` env var, which will then be used by the Node SDK to get a task's resource attachments and builtins.

This PR also copies over many of the builtins structs - these should eventually go into their own repo but are included here for now. 

## Test plan
Tested with the below dev config and verified that resources were serialized into the `AIRPLANE_RESOURCES` env var properly:
```
env:
  API_KEY: asdf
  IMAGE: 1234
resources:
  db:
    kind: postgres
    slug: db
    username: admin
    password: airplane
  rest:
    kind: rest
    slug: rest
    baseURL: "localhost:1234"
```


go.mod
```diff
@@ -5,7 +5,7 @@ go 1.18
 require (
 	github.com/AlecAivazis/survey/v2 v2.3.5
 	github.com/MakeNowJust/heredoc v1.0.0
-	github.com/airplanedev/lib v0.5.25
+	github.com/airplanedev/lib v0.5.25-0.20220714001423-df759db6bb66
```
calufornia, line 8: Will update this before landing

pkg/conf/dev.go
```diff
@@ -0,0 +1,82 @@
...
+	Resources        map[string]map[string]interface{} `yaml:"resources"`
+	DecodedResources map[string]resource.Resource
+}
+
+func (d *DevConfig) DefaultPath() string {
+	homedir, err := os.UserHomeDir()
+	if err != nil {
+		panic("$HOME environment variable must be set")
+	}
+	return filepath.Join(
```
calufornia, line 25: Default filepath is still TBD

pkg/resource/postgres.go
```diff
@@ -0,0 +1,29 @@
+package resource
```
calufornia, line 1: I've only added postgres for now, but assuming everything else looks good, will add all the other resource structs

calufornia (commented):
Tests to be added

pkg/resource/postgres.go
kellymtrinh: I pretty much need these too for constructing the built in requests :/ maybe we can start with just moving the resource factories and clients out into a built-ins repo

pkg/conf/dev.go
calufornia: Decided offline that there won't be a default filepath - users will specify a path when doing `airplane dev`

kellymtrinh (approved):

PR #532, opened 2022-07-20
Check if we should use yarn for views init

rose-huang:
## Description
We try to use yarn if user has a `yarn.lock` or if they have yarn installed unless the user is creating their view in a directory with a `package-lock.json`

RFC because I'm not sure if we want to do this or just default to using npm all the time if neither lockfile exists even if they have yarn installed. Currently tasks will default to npm if no lockfile is found.

## Test plan
Uninstalled yarn, init in a fresh directory and it uses npm
Re-installed yarn, init in a fresh directory and it uses yarn

With yarn installed, init in a nested directory with a package-lock.json and it uses npm

Is there an easy way to write a test for this?


cmd/airplane/views/initcmd/init.go
```diff
@@ -153,6 +153,25 @@ func createEntrypoint(cfg config) error {
...
+		return true
+	} else if err := fsx.AssertExistsAll(pkglock); err == nil {
+		return false
+	}
+
+	// No lockfiles, so check if yarn is installed by getting yarn version
+	cmd := exec.Command("yarn", "-v")
+	cmd.Dir = filepath.Dir(packageJSONDirPath)
+	err := cmd.Start()
+	return err == nil
```
rose-huang, line 172: Not sure if there's a preferred way to check if this error is because of yarn not being installed. If I try to run yarn -v without yarn it will say `Exec: "yarn": executable file not found in $PATH`

cmd/airplane/views/initcmd/init.go
```diff
@@ -192,33 +212,41 @@ func createPackageJSON(cfg config) error {
...
-			return yarnAddViewsPackage(packageJSONDirPath)
+			return addViewsPackage(packageJSONDirPath, useYarn)
 		}
 	}
+
+	var cmd *exec.Cmd
 	// Create package.json in cwd and install
-	cmd := exec.Command("yarn", "init", "-y")
+	if useYarn {
+		cmd = exec.Command("yarn", "init", "-y")
```
rose-huang, line 222: yarn init and npm init create package.json with slightly different stuff. Not sure if this is a good idea to call yarn/npm init or if we should just copy over our own package.json 
npm: 
```
{
  "name": "testnpm",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "dependencies": {
    "@airplane/views": "^0.0.2"
  }
}
```

yarn:
```
{
  "name": "testyarn",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "dependencies": {
    "@airplane/views": "^0.0.2"
  }
}
```

leeweisberger (approved):
I think we should use our own package.json instead of calling init, but will approve. Preferably we do that

PR #534, opened 2022-07-25
Add error boundary to view when devving

leeweisberger:
## Description
Add error boundary so that errors are caught and displayed rather than crashing the View.


zhan-airplane (approved):
PR look good, but OOC are we adding the same error boundary to deploy/are we going to have any major differences between the dev view and deployed view? I'm opening a PR soon(tm) that's gonna use the lib version of these template files when devving

leeweisberger:
> 

I'm adding the same thing. sgtm for them using the same template file

leeweisberger:
We've moved to using lib for this, and I implemented in lib

PR #539, opened 2022-07-26
Add other packages during airplane views init

rose-huang:
## Description
We also want to install react, @types/react, typescript, etc so that we get the right typescript hints

## Test plan
tested adding more packages to directory with existing package.json
tested init with yarn uninstalled to make sure npm works 

cc @zhan-airplane


leeweisberger:
I'm going to assign this to @zhan-airplane instead. He did something very similar. Can we share code here @zhan-airplane ? Can we share the dependency list that has to be installed? How is this different from what you did?

rose-huang:
> Can we share code here?
I saw Zhan's change in lib but there wasn't that much code to re-use if Zhan's actually editing the contents of the package.json and spitting it out again. I didn't want to put "*" as the version in someone's existing package.json because that's probably not ideal 

there are some overlapping dependencies in this PR and in zhan's `airplane dev` PR but I included them here because I think it will make vscode editor support better. Having dev dependencies for @types/react and @types/react-dom seemed a bit weird if we didn't have react and react-dom so I put those in dependency list too

cmd/airplane/views/initcmd/init.go
```diff
@@ -218,33 +218,85 @@ func createPackageJSON(cfg config) error {
...
+
+	devPackagesToCheck := []string{"@types/react", "@types/react-dom", "typescript"}
+	devPackagesToAdd := getPackagesToAdd(devPackagesToCheck, existingDeps)
+
+	if len(packagesToAdd) > 0 || len(devPackagesToAdd) > 0 {
+		logger.Step("Installing dependencies...")
+	}
+
+	for _, pkg := range packagesToAdd {
+		if err := addPackage(packageJSONDirPath, pkg, false, useYarn); err != nil {
```
zhan-airplane, line 248: I think we can `yarn add` multiple packages in the same command at once right? Might run faster

zhan-airplane: Also do we want to pick versions? Doesn't need to be in this PR (maybe leave a TODO comment), but we probably should have a list somewhere inside `lib` so that we can fix versions correctly in both deploy and dev and this

cmd/airplane/views/initcmd/init.go
```diff
@@ -218,33 +218,85 @@ func createPackageJSON(cfg config) error {
...
 	}
-	return addViewsPackage(cwd, useYarn)
+	return addAllPackages(cwd, useYarn)
 }
 
-func addViewsPackage(packageJSONDirPath string, useYarn bool) error {
-	logger.Step("Installing @airplane/views...")
+func addAllPackages(packageJSONDirPath string, useYarn bool) error {
+	packageJSONPath := filepath.Join(packageJSONDirPath, "package.json")
+	existingDeps, err := build.ListDependencies(packageJSONPath)
```
zhan-airplane, line 233: :O actually I just realized this function exists hahaha I should switch my code over to use it

zhan-airplane (approved):

PR #540, opened 2022-07-26
Implement dev server http api

calufornia:
## Description

Adds the dev server endpoints (`/dev/list` + `/dev/tasks/<task_slug>`) that are needed for the previewer. Also contains general refactoring and adds a `Wrap` function (that we already use internally) to make it easier to return error responses. 

## Test plan
```
aplocal server
Found 1 task and 0 views:
- sunt in tempor eu

Visit https://app.airplane.dev/editor?host=http://localhost:7190 for a development UI.
```
```
> curl "http://localhost:7190/dev/tasks/my_task"
{"name":"sunt in tempor eu","slug":"my_task","description":"ut dolor sit officia ea","node":{"entrypoint":"./single_task.js","nodeVersion":"18"},"allowSelfApprovals":true,"timeout":3600}
> curl "http://localhost:7190/dev/list"
{"tasks":[{"name":"sunt in tempor eu","slug":"my_task"}],"views":[]}
```


calufornia:
Will need to update lib dependency once https://github.com/airplanedev/lib/pull/198 is merged.

pkg/server/errors.go
```diff
@@ -0,0 +1,39 @@
...
+			WriteHTTPError(w, r, err)
+			return
+		}
+	}
+}
+
+// WriteHTTPError writes an error to response and optionally logs it
+func WriteHTTPError(w http.ResponseWriter, r *http.Request, err error) {
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusInternalServerError)
```
kellymtrinh, line 30: are these all internal errors?

kellymtrinh (approved):

PR #544, opened 2022-07-27
Suggest next steps to after views init

rose-huang:
## Description
Add next steps after initializing a view

## Test plan
```
? What should this view be called? view10
- Created view definition at view10/view10.view.yaml
- Created view entrypoint at view10/view10.view.tsx

✅ To complete your view:
- Write your view logic in view10/view10.view.tsx
- Add a description in view10/view10.view.yaml

⚡ To run the view locally:
  airplane views dev view10

🛫 To deploy your view to Airplane:
  airplane deploy view10

```


cmd/airplane/views/initcmd/init.go
```diff
@@ -345,3 +353,23 @@ func createTSConfigFile() error {
...
+	logger.SuggestSteps("✅ To complete your view:", steps...)
+
+	logger.Suggest(
+		"⚡ To run the view locally:",
+		"airplane views dev %s",
+		cfg.slug,
+	)
+	logger.Suggest(
+		"🛫 To deploy your view to Airplane:",
+		"airplane deploy %s",
```
rose-huang, line 372: @leeweisberger do you have an opinion on `airplane views deploy` vs `airplane deploy`? I left it as `airplane deploy` because I figured we would maybe add some tasks soon. 

Also do you have a preference for deploying the folder (slug) or the yaml file (slug/slug.view.yaml)? Same question for dev.
I put just slug in case we get rid of the yaml soon. But we may also get rid of the folder so i dont really know

cmd/airplane/views/initcmd/init.go
```diff
@@ -345,3 +353,23 @@ func createTSConfigFile() error {
 	}
 	return nil
 }
+
+func suggestNextSteps(cfg config) {
+	steps := []string{}
+
+	steps = append(steps, fmt.Sprintf("Write your view logic in %s", generateEntrypointPath(cfg, false)))
+	steps = append(steps, fmt.Sprintf("Add a description in %s", generateDefinitionFilePath(cfg)))
```
leeweisberger, line 361: I would maybe take this out as it's not really important. Not on the golden path to deploying a view

cmd/airplane/views/initcmd/init.go
leeweisberger: sgtm what you've done

leeweisberger (approved):

PR #549, opened 2022-07-29
Add ability to start view from local dev server

calufornia:
## Description
This PR adds a `/dev/startView/<view_slug>` endpoint that runs vite locally. This is used by the previewer to preview any discovered views - we track the vite process in the dev server's state, and kill the existing process (if applicable) before spinning up a new vite server process.

Moves the majority of the views dev logic from `cmd` into `pkg` so that it can be reused by the local dev server.

Also contains some minor changes:
- Changing the default dev server port to 4000
- Renaming DevMetadata -> AppMetadata - apps encompasses both tasks and views.

## Test plan
Tested by running `aplocal server` and verifying that clicking on a view from the app sidebar resulted in the local vite server starting up and the view being properly served.


kellymtrinh (approved):
Looks good- I think some tests are still failing after refactor `getAdditionalEnvs`

pkg/server/dev.go
```diff
@@ -83,3 +98,59 @@ func GetTaskHandler(state *State) http.HandlerFunc {
 		return json.NewEncoder(w).Encode(taskConfig.Def)
 	})
 }
+
+// StartViewHandler handles requests to the /dev/tasks/<task_slug> endpoint.
+func StartViewHandler(state *State) http.HandlerFunc {
+	return Wrap(func(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
+		if state.viteProcess != nil {
```
zhan-airplane, line 105: This is pretty race-condition-y right? If somehow you get two requests to the start view handler

pkg/server/dev.go
```diff
@@ -83,3 +98,59 @@ func GetTaskHandler(state *State) http.HandlerFunc {
...
+			EnvSlug: state.envSlug,
+		})
+		if err != nil {
+			return errors.Wrapf(err, "starting views dev")
+		}
+
+		state.viteProcess = cmd.Process
+
+		for {
+			conn, _ := net.DialTimeout("tcp", net.JoinHostPort("localhost", "3001"), 10*time.Second)
```
zhan-airplane, line 143: Is there a way not to hardcode this port? I'm pretty sure it's not that reliably 3001

zhan-airplane: maybe we should parse the vite process output haha idk

pkg/server/api.go
```diff
@@ -127,6 +128,17 @@ func GetTaskMetadataHandler(state *State) http.HandlerFunc {
 	})
 }
 
+// GetViewHandler handles requests to the /v0/views/get endpoint. It generates a random view ID for each view
+// found locally, and its primary purpose is to ensure that the view discoverer does not error.
+func GetViewHandler(state *State) http.HandlerFunc {
+	return Wrap(func(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
+		return json.NewEncoder(w).Encode(libapi.View{
+			ID:   "vew" + utils.RandomString(10, utils.CharsetLowercaseNumeric),
```
zhan-airplane, line 136: can we generate a string that's e.g. a hash of the slug or something deterministic instead? just in case in the future we ever want to do something that needs a stable ID for the slug

zhan-airplane (approved):

pkg/server/dev.go
kellymtrinh: Hm I just ran into this because I had something running 3001 already and vite chose 3002 (and then 3003 etc)

pkg/server/api.go
kellymtrinh: I've been using task slug everywhere instead of ID for local cli related things- currently working on this. I don't have strong opinions on this though (it's still early stages for task preview)

PR #550, opened 2022-07-29
Support local builtin execution

kellymtrinh:
## Description
Adds support for local builtin execution.
- If a builtin is being called, it will install the builtins package if it isn't already installed and use that
-- we install into the tmp directory
-- if the file already exists, only update the file if the latest checksum on gcs doesn't match 

## Test plan
This is the example I used to test: https://github.com/airplanedev/workflow-examples/pull/5

- Tested builtin execution works for the first time (fresh install from gcs)
- Tested that using the cached version of builtins works
- Tested that an out of date version gets updated and works (I manually edited the checksum file so it doesn't match)

- [x] Get a non mac user to test - Tested with Zhan on his linux computer
(this is debug mode)
 Screenshot from 2022-08-01 12-39-34


pkg/dev/builtins.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	return exec.CommandContext(ctx, binaryPath, req), nil
+}
+
+// Gets the binary file name ie. builtins-linux-amd64
+func (b *BuiltinClient) FileName() string {
+	return fmt.Sprintf("builtins-%s-%s", b.operatingSystem, b.architecture)
+}
+
+func (b *BuiltinClient) TmpDirPath() string {
```
kellymtrinh, line 95: Need to refactor and share this logic with views `.CacheDir()`

pkg/server/api.go
```diff
@@ -49,38 +50,51 @@ func ExecuteTaskHandler(state *State) http.HandlerFunc {
...
 
 		runID := "run" + utils.RandomString(10, utils.CharsetLowercaseNumeric)
 		runStatus := api.RunFailed
 		outputs := api.Outputs{}
-		if config, ok := state.taskConfigs[req.Slug]; ok {
-			kind, kindOptions, err := config.Def.GetKindAndOptions()
-			if err != nil {
-				return errors.Wrap(err, "failed to get kind and options from task config")
+		localTaskConfig, ok := state.taskConfigs[req.Slug]
+		if builtins.IsBuiltinTaskSlug(req.Slug) || ok {
```
calufornia, line 60: let's set the result of `builtins.IsBuiltinTaskSlug` to a variable so we can use it below

pkg/server/api.go
```diff
@@ -49,38 +50,51 @@ func ExecuteTaskHandler(state *State) http.HandlerFunc {
...
+				Port:        state.port,
+				Root:        state.cli,
+				Slug:        req.Slug,
+				EnvSlug:     state.envSlug,
+				IsBuiltin:   builtins.IsBuiltinTaskSlug(req.Slug),
+			}
+			resourceAttachments := map[string]string{}
+			if builtins.IsBuiltinTaskSlug(req.Slug) {
+				// builtins have access to all resources the parent task has
+				for k := range state.devConfig.DecodedResources {
```
calufornia, line 72: nit: call the loop variable `slug` or something similar

pkg/server/api.go
```diff
@@ -49,38 +50,51 @@ func ExecuteTaskHandler(state *State) http.HandlerFunc {
...
+					resourceAttachments[k] = k
+				}
 			}
+			if localTaskConfig.Def != nil {
```
calufornia, line 76: can this be an `else if`? a builtin would never have task definition associated with it, right?

pkg/dev/builtins.go
```diff
@@ -0,0 +1,209 @@
...
+	checksum, err := os.ReadFile(checksumpath)
+	if err != nil {
+		logger.Log("error reading checksum file %v", err)
+		return false
+	}
+	return bytes.Compare(checksum, attrs.MD5) == 0
+}
+
+// Downloads the builtin binary from GCS
+func (b *BuiltinClient) download(fileName string, tmpDir string) (string, error) {
```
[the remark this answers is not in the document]
kellymtrinh: Moved this so we use the directory field from the client

pkg/dev/builtins.go
```diff
@@ -0,0 +1,209 @@
...
+	logger.Debug("Downloading builtin binary...")
+	obj := b.Bucket().Object(b.gcsFileName())
```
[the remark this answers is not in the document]
kellymtrinh: Decided not to do this step because it didn't feel like a lot of reusable stuff - `download` still needs the obj while isLatestVersion only needs the attrs. Didn't feel like it's saving much since both would call `attrs, err := obj.Attrs(context.Background())` or `obj.getLatestAttributes()`

kellymtrinh:
@calufornia I resolved each review comment (or left a comment on why not) for each! Hope I didn't miss anything. Thanks!

pkg/dev/builtins.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	b, err := json.Marshal(req)
+	if err != nil {
+		log.Fatal(err)
+	}
+	return string(b), nil
+}
+
+var supportedOS = map[string]bool{"darwin": true, "linux": true}
```
calufornia, line 42: could we move these top to the top of the file?

pkg/dev/builtins.go
```diff
@@ -0,0 +1,199 @@
...
+	if !b.isLatestVersion() {
+		logger.Debug("Builtins package out of date. Getting latest version.")
+		return b.download()
+	}
+	logger.Debug("Using cached builtins package: %s", b.binaryPath)
+	return b.binaryPath, nil
+}
+
+const (
+	checksumFileName  = "checksum.txt"
```
calufornia, line 128: could we also move these to the top? consts are generally first, followed by vars, followed by methods

pkg/dev/dev.go
```diff
@@ -99,27 +141,27 @@ func (l *LocalExecutor) Execute(ctx context.Context, config LocalRunConfig) (api
...
-	if len(config.Env) > 0 {
-		env = config.Env
-	}
-
 	// cmd.Env defaults to os.Environ _only if empty_. Since we add
 	// to it, we need to also set it to os.Environ.
 	cmd.Env = os.Environ()
-	for k, v := range env {
-		cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", k, v))
+	if r != nil {
```
calufornia, line 147: could you add a comment here on why the runtime might be nil?

calufornia (approved):
nice!

PR #553, opened 2022-08-01
Clean up airplane views dev logs

rose-huang:
## Description
You have to pass in --debug to see all of the logs now. Without --debug, it just shows you the 2 lines (see test plan)

## Test plan
built CLI and tested

```
Started development server at http://localhost:3002/ (^C to quit)
Press ENTER to preview your view in the browser
```

In the future, I can try to make an integration test in case vite changes how they output their server


zhan-airplane (changes requested):
We have a `logger.Debug` right? Could we use that instead of putting `if`s on this layer?

zhan-airplane (approved):

PR #554, opened 2022-08-01
Create demo DB on views init if not exists

zhan-airplane:
## Description
Depends on airplanedev/airport#2917. Lists resources to find something matching the name `[Demo DB]`. If it exists, assumes that's the correct DB. Otherwise, creates a new demo DB

## Test plan
Ran locally (needed to set up `AP_DEMO_DB_CONFIG` in `api.env` as well as run something like `cloud_sql_proxy -instances=airplane-stage:us-central1:demo=tcp:172.20.0.1:17777`, but it works with that)

---
cc @stepzhou @rose-huang


cmd/airplane/views/initcmd/init.go
```diff
@@ -130,6 +136,22 @@ func generateDefinitionFilePath(cfg config) string {
...
+	resp, err := cfg.client.ListResources(ctx)
+	if err != nil {
+		return "", err
+	}
+	for _, r := range resp.Resources {
+		if r.Name == demoDBName {
+			return r.ID, nil
+		}
+	}
+	return cfg.client.CreateDemoDB(ctx, demoDBName)
```
leeweisberger, line 152: Didn't we decide that this is already idempotent? Why are you adding another idempotency check? It's an extra round trip and your idempotency here might drift from what we've done on the server

zhan-airplane: Ah when I was testing, turns out the call on API isn't actually idempotent (it throws an error if you already have a demo DB created with the given name). Wasn't sure if that was the original intended behavior, so decided to do idempotency on this side without changing the original behavior.

On second thoughts though, I don't think it's wrong to make the create demo DB endpoint on API idempotent after all (or at least, not do anything if the name already exists), I'll open another PR to do that I guess

leeweisberger: ok thanks. You can confirm with maybe @francescaleung but yeah if we just make that idempotent everyone gets it. Send me the PR I'll give a quick :+1:

zhan-airplane: PR: https://github.com/airplanedev/airport/pull/2918

leeweisberger (approved):
Going to approve, but like we discussed offline, let's just do this if there is a `getting-started` flag!

PR #556, opened 2022-08-01
Update vite regex

rose-huang:
## Description
vite changed what they print out from v2 to v3 :(
changed from

```
> Local: http://localhost:3001
```
to
```
➜  Local:   http://localhost:5173/
```
so this looks for both

## Test plan
airplane views dev with v1 and v2 vite


zhan-airplane (approved):

zhan-airplane:
Actually I wonder if we can make this a more robust check haha (e.g. pinging the given port to see if it's open maybe?)

rose-huang:
> Actually I wonder if we can make this a more robust check haha (e.g. pinging the given port to see if it's open maybe?)

@zhan-airplane hmm I was thinking of just suggesting users go to 5173 but vite will increment the port by default and try again if 5173 is in use. 
  
We can configure this to be strict:
https://vitejs.dev/config/server-options.html#server-port

However, it seems possible that people may want to dev 2 views at the same time and then it would not work well with a strict port.

Do you think this is ok for now? We can try to make this better once we can dev multiple views at once with the project that team workflows is working on

zhan-airplane:
Yeah it's fine as it is, was just a thought I had

PR #560, opened 2022-08-03
Handle empty slug list from task yaml

ameyakhare:
## Description
Handle parsing an empty slug list from a task yaml file; this can happen if the yaml is malformed or is not an airplane task yaml file.

## Test plan
Doesn't fault anymore.


colinking (approved):
LGTM. Do we have this check in all places we call GetAirplaneTasks now?

ameyakhare:
> LGTM. Do we have this check in all places we call GetAirplaneTasks now?

yup!

PR #562, opened 2022-08-04
Add SSE endpoint for dev server logs

calufornia:
## Description

This PR adds an endpoint, `dev/logs/{run_id}`, that uses server-sent events to stream logs to the previewer during task execution. We add a `LogConfig` to the state - this struct contains a channel that logs are sent to during task execution, and also a logs array that is used to store logs so far (so that e.g. if the user refreshes a page, the logs still persist). Because this channel needs a receiver by the time the task starts streaming logs, we also add an additional `createRun` endpoint that will manually initialize a `LocalRun` struct (and hence the logging channel)  - without this, the `dev/logs/{run_id}` endpoint would need to poll for an entry in `state.runs` until the LocalRun has been created during task execution.

## Test plan
Tested locally by manually executing a task and creating an event source and verified that logs were streamed correctly.


pkg/server/dev.go
```diff
@@ -81,6 +85,24 @@ func ListAppMetadataHandler(state *State) http.HandlerFunc {
 	})
 }
 
+func CreateRunHandler(state *State) http.HandlerFunc {
+	return Wrap(func(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
+		vars := mux.Vars(r)
```
kellymtrinh, line 90: Maybe we can have the backend generate the run_id and return it so it's not all set by the client

kellymtrinh (approved):

PR #567, opened 2022-08-08
Modify deploy command to default to the current directory.

geodn:
## Description
`airplane deploy` with no arguments now defaults to the current directory.

## Test plan
Execute before and after making the change:
```
go install ./cmd/airplane
airplane deploy
```


cmd/airplane/tasks/deploy/deploy.go
```diff
@@ -56,12 +56,13 @@ func New(c *cli.Config) *cobra.Command {
 			airplane tasks deploy my-directory
 			airplane tasks deploy ./my_task1.task.yml ./my_task2.task.json my-directory
 		`),
-		Args: cobra.MinimumNArgs(1),
+		Args: cobra.MinimumNArgs(0),
```
rose-huang, line 59: i don't think we would need this line if minimum is 0, can you try removing it?

rose-huang (changes requested):

cmd/airplane/tasks/deploy/deploy.go
geodn: Done.

rose-huang (approved):

PR #579, opened 2022-08-12
Add support for displays during local dev

colinking:
## Description

Adds support for seeing displays when locally developing on a task. The displays are stored in the run store where they can be accessed by the UI. They're also logged out for the existing `airplane dev` usage, but we can remove that once we fully migrate all local dev to the server.

Resolves AIR-4485

## Test plan

Run a task locally that produced displays. The displays should render in the UI when running https://github.com/airplanedev/web/pull/1639


colinking:
* main
  * **PR #579** <a href="https://app.graphite.dev/github/pr/airplanedev/cli/579" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/cli/579?utm_source=stack-comment).

pkg/print/print.go
```diff
@@ -81,11 +81,24 @@ func Print(obj interface{}, defaultPrintFunc func()) {
 // | hello |
 // +-------+
 func BoxPrint(s string) {
+	BoxPrintWithPrefix(s, "")
+}
+
+func BoxPrintWithPrefix(s, prefix string) {
```
colinking, line 87: Added support for prefixes and for multi-line strings.

pkg/server/api.go
```diff
@@ -30,6 +31,7 @@ type LocalRun struct {
 	Parameters  *libapi.Parameters     `json:"parameters"`
 	LogStore    *dev.LogStore          `json:"-"`
 	TaskName    string                 `json:"taskName"`
+	Displays    []libapi.Display       `json:"displays"`
```
colinking, line 34: Depends on: https://github.com/airplanedev/lib/pull/225

pkg/server/api.go
```diff
@@ -256,3 +263,78 @@ func ListResourcesHandler(state *State) http.HandlerFunc {
...
+		}
+
+		// TODO: remove me
+		// HACK: the root run is not inserted into the run store. For now, we get around
+		// that by inserting a fake version of the run JIT. Remove this code once we start
+		// inserting the root run into the store.
+		if _, ok := state.runs.get(runID); !ok {
+			state.runs.add("tsk123", runID, *NewLocalRun())
+		}
+		// END HACK
```
colinking, line 303: cc @calufornia this is the hack I was talking to you about

kellymtrinh: what was the context?

colinking: Since the root run is not registered in the run store, we can't fetch run metadata during the call to create a display. Once we register the root run in the store, then we can remove this hack.

pkg/server/api.go
```diff
@@ -153,22 +162,22 @@ func ExecuteTaskHandler(ctx context.Context, state *State, req ExecuteTaskReques
...
 		outputs, err := state.executor.Execute(ctx, runConfig)
-		if err != nil {
-			run.Status = api.RunFailed
-		} else if err == nil {
-			run.Status = api.RunSucceeded
-		}
-		run.Outputs = outputs
+
+		completedAt := time.Now()
+		run, _ = state.runs.update(runID, func(run *LocalRun) {
```
colinking, line 167: @kellymtrinh I had to change this to get displays working. The original implementation would overwrite the `run` in the state.

cmd/airplane/tasks/dev/dev.go
```diff
@@ -235,6 +235,7 @@ func run(ctx context.Context, cfg taskDevConfig) error {
 	}
 
 	localRunConfig := dev.LocalRunConfig{
+		ID:          "run" + utils.RandomString(10, utils.CharsetLowercaseNumeric),
```
kellymtrinh, line 238: Maybe move this into a generateRunID function since it's used in two places now?

kellymtrinh (approved):

PR #581, opened 2022-08-12
Use generics for handlers

calufornia:
## Description
Modifies the local dev server to use handlers that accept generic request/response types to make adding new endpoints easier.

## Test plan
Tested with `aplocal dev --editor`


pkg/server/errors.go
```diff
@@ -24,6 +24,39 @@ func Wrap(f func(ctx context.Context, w http.ResponseWriter, r *http.Request) er
 	}
 }
 
+// HandlerWithBody is an API handler that reads a JSON-encoded request body, calls a provided handler,
```
kellymtrinh, line 27: Rename this file to be `handler.go` since it's not errors anymore?

kellymtrinh (approved):

PR #586, opened 2022-08-15
Add support for local dev for config as code

ameyakhare:
Resolves TSK-93

## Description
Adds support for running config as code tasks locally.
Removes a lot of redundant copied discover logic from lib.

Running code-configured task works in the following way (defn files unaffected):

`airplane dev <file>`: run the only task or the only default task in the file (error otherwise if multiple tasks exist, none of which are default)
`airplane dev <file>::<exportName>`: run a specific task based on its exported name

## Test plan
Ran different configurations of code-only files locally.


go.mod
```diff
@@ -2,6 +2,10 @@ module github.com/airplanedev/cli
 
 go 1.19
 
+replace (
+	"github.com/airplanedev/lib" => "/Users/ameya/airplane/lib"
```
calufornia, line 6: Remove this

calufornia (commented):
Does this work from the previewer? i.e. could you test that tasks are discovered when doing `airplane dev --editor` and that you can execute the task from the browser?

ameyakhare:
> Does this work from the previewer? i.e. could you test that tasks are discovered when doing `airplane dev --editor` and that you can execute the task from the browser?

works now, had to update the api a little

cmd/airplane/tasks/dev/dev.go
```diff
@@ -37,6 +37,9 @@ type taskDevConfig struct {
 	// TODO: Remove these fields once launching the dev server is the default behavior of airplane dev
 	fileOrDir string
 	args      []string
+	// If there are multiple tasks a, b in file f (config as code), specifying airplane
+	// dev f:a would set fileOrDir to f and entrypointFunc to a.
```
calufornia, line 41: Could you add an example usage in the command below - maybe replace [this line](https://github.com/airplanedev/cli/blob/dc636c3c8e4fe2206684a146a85909627ca1cb03/cmd/airplane/tasks/dev/dev.go#L55) which seems to be a duplicate

cmd/airplane/tasks/dev/dev.go
```diff
@@ -197,36 +209,40 @@ func run(ctx context.Context, cfg taskDevConfig) error {
...
 	}
+	taskConfig, err := getLocalDevTaskConfig(taskConfigs, cfg)
+	if err != nil {
+		return err
+	}
 
 	// TODO: Allow users to re-register tasks once we move to a long-running local api server
 	apiServer.RegisterTasksAndViews(taskConfigs, viewConfigs)
 
-	taskInfo, err := getTaskInfo(ctx, cfg)
```
calufornia, line 204: 👏

pkg/dev/dev.go
```diff
@@ -228,6 +229,23 @@ func (l *LocalExecutor) Execute(ctx context.Context, config LocalRunConfig) (api
...
+		return "", build.KindOptions{}, errors.Wrap(err, "getting kind and kind options")
+	}
+
+	buildConfig, err := taskConfig.Def.GetBuildConfig()
+	if err != nil {
+		return "", build.KindOptions{}, errors.Wrap(err, "getting build config")
+	}
+	if entrypointFunc, ok := buildConfig["entrypointFunc"]; ok {
+		// Config as code depends on using the correct import name
+		kindOptions["entrypointFunc"] = entrypointFunc
```
calufornia, line 244: Not needed in this PR, but could we just make it so that `Def.GetKindAndOptions()` populates the `"entrypointFunc"` entry in kind options?

calufornia (approved):

PR #587, opened 2022-08-16
Fix: close log channel when local task fails and create run before execution

kellymtrinh:
## Description
Fixes:
- When a task fails, we don't show the output or close the logs channel in the previewer
- We should create the task before executing it, then update the run status at the end so that calls to getTask or getRun while it's still being executed aren't blocked

## Test plan
- [x] `airplane dev` a successful task is unaffected
- [x] `airplane dev` a task that throws an error is unaffected
- [x] Running a failed task via previewer/server correctly closes the log channel


pkg/dev/dev.go
```diff
@@ -225,7 +222,7 @@ func (l *LocalExecutor) Execute(ctx context.Context, config LocalRunConfig) (api
 		config.LogStore.DoneChannel <- true
```
calufornia, line 222: Not needed in this PR, but thinking about it some more, it might be better to do

```
defer func() {
  if config.LogStore.DoneChannel != nil {
    config.LogStore.DoneChannel <- true
  }
}()
```

at the top just so that we're guaranteed that the event stream gets closed even if we don't reach here - I can also add that in a separate PR

calufornia (approved):

pkg/dev/dev.go
kellymtrinh: Ooh yeah that's better!

PR #588, opened 2022-08-17
Add server-side support for modifying resources through the editor

calufornia:
## Description

This PR makes several changes that are required for resource modification through the previewer. In particular, it

- Adds create/get/update endpoints for resources, which use a `InternalResource` struct similar to the one in our API today so that the web app can call our endpoints
- Changes resources in the dev config to be a list instead of a map - it's redundant to include the slug both as a key as well as field within the resource.
- Splits up the api server into an internal api and external api, similar to our prod API - this is again to help facilitate the reuse of components in the web app.
- Makes the dev config file a pointer most places to support modification of the resources slice over the lifetime of the dev server.

Depends on https://github.com/airplanedev/lib/pull/230

## Test plan

`aplocal dev --editor` and modified resources through the previewer.


cmd/airplane/tasks/dev/config/set_resource/reflect.go
```diff
@@ -0,0 +1,85 @@
...
+
+import (
+	"fmt"
+	"os"
+	"reflect"
+
+	"github.com/AlecAivazis/survey/v2"
+)
+
+// Derived from https://gist.github.com/hvoecking/10772475
```
kellymtrinh, line 11: maybe mention it has the MIT license?

kellymtrinh (approved):
Wonder if we need the dev endpoints anymore or if those should be put under internal- doesn't have to be as part of this though

PR #589, opened 2022-08-17
Create new root-level init command

rose-huang:
## Description
Calling this init-dev for now while we polish the content
This will take in a template flag. If no template is provided, it will ask you if you want to init task, view, or from a template and show a list of templates

## Test plan
https://user-images.githubusercontent.com/34196250/185261509-aa6e60a0-1093-46ca-9d87-ae7b17a74a5e.mov


pkg/utils/copy.go
```diff
@@ -45,3 +49,84 @@ func CopyDirectoryContents(srcDirectory string, dstDirectory string) error {
 		return CopyFile(path, filepath.Join(dstDirectory, filepath.Dir(rel)))
 	})
 }
+
```
rose-huang, line 52: stuff in this file is moved over from `views/initcmd/init.go` and changed so it wouldn't need the view cfg param

cmd/airplane/root/initcmd/init.go
```diff
@@ -0,0 +1,162 @@
...
+func initFromTemplate(ctx context.Context, templates []Template, gitPath string) error {
+	template, err := FindTemplate(templates, gitPath)
+	if err != nil {
+		return err
+	}
+	// TODO: add next steps
+	return utils.CopyFromGithubPath(template.GitHubPath)
+}
+
+const docsUrl = "https://docs.airplane.dev/templates/templates.json"
```
rose-huang, line 126: may need to move this to lib later on but i don't think we'll need this outside of cli for rn

cmd/airplane/root/initcmd/init.go
```diff
@@ -0,0 +1,162 @@
...
+		}
+	}
+	return nil
+}
+
+var orderedInitOptions = []string{
+	"From template",
+	"Task",
+	"View",
+}
```
leeweisberger, line 110: Did we discuss this order at all? Depends when we're going to launch this, but right now we only have 1 template. I'm for this in the future, but would prefer to wait until templates are useful

leeweisberger: Also I think it would be good if we included a description of what these things are for newbies. What is a task, view, and template and why would I choose them? Can you add some more helpful content here?

cmd/airplane/root/initcmd/init.go
```diff
@@ -0,0 +1,162 @@
...
+type Template struct {
+	Name          string   `json:"name"`
+	Description   string   `json:"description"`
+	GitHubPath    string   `json:"githubPath"`
+	DemoResources []string `json:"demoResources"`
+	ViewSlugs     []string `json:"viewSlugs"`
+	TaskSlugs     []string `json:"taskSlugs"`
+}
+
+func initFromTemplate(ctx context.Context, templates []Template, gitPath string) error {
```
leeweisberger, line 121: Can you start off a spinner here (using spinner logger or whatever it's called. Look at the deploy code). I think we should get in the habit of adding the spinner to any commands that have latency so we don't just have a blank screen

cmd/airplane/root/initcmd/init.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	defer resp.Body.Close()
+	buf, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return []Template{}, err
+	}
+
+	var t []Template
+	if err = json.Unmarshal(buf, &t); err != nil {
+		return []Template{}, err
```
leeweisberger, line 146: Let's wrap all errors that don't come from our own code

cmd/airplane/root/initcmd/init.go
```diff
@@ -0,0 +1,166 @@
...
+	"View",
+}
+
+type Template struct {
+	Name          string   `json:"name"`
+	Description   string   `json:"description"`
+	GitHubPath    string   `json:"githubPath"`
+	DemoResources []string `json:"demoResources"`
+	ViewSlugs     []string `json:"viewSlugs"`
+	TaskSlugs     []string `json:"taskSlugs"`
```
leeweisberger, line 118: What are these things?

cmd/airplane/root/initcmd/init.go
rose-huang: gonna do this in a separate PR if thats ok! added it to the TODO comment

cmd/airplane/root/initcmd/init.go
rose-huang:
```
➜  test8 ~/airplanedev/cli/bin/airplane init-dev        
? What would you like to initialize?    [Use arrows to move, type to filter]
> Task          Create a SQL query, script, or API call
  View          Build an internal UI
  Template      Clone ready-made code to develop your tools faster
```

I'm going to do a content polish and get feedback in #product slack before switching this over from `airplane init-dev` to `airplane init`

cmd/airplane/root/initcmd/init.go
rose-huang: demoResources are the names of the demo resources used for the template. We'll use this in the future to know if we need to reset the db / fake api.

viewSlugs and taskSlugs are the slugs of the views and tasks in the template. We added these so that we could warn users if they have existing tasks/views with these slugs 

https://www.notion.so/airplanedev/PRD-Cloning-from-templates-2ff1c566c04740d4a270ff3f2f31ec37#411729083bf3499bbfa183f9736ec269

pkg/utils/copy.go
rose-huang: also switched copy to use the non task specific open github directory lib function

cmd/airplane/root/initcmd/init.go
```diff
@@ -0,0 +1,172 @@
...
+		PersistentPreRunE: utils.WithParentPersistentPreRunE(func(cmd *cobra.Command, args []string) error {
+			return login.EnsureLoggedIn(cmd.Root().Context(), c)
+		}),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return run(cmd.Root().Context(), cfg)
+		},
+		Hidden: true,
+	}
+
+	cmd.Flags().StringVarP(&cfg.template, "template", "t", "", "Path of a template to initialize from in the format github.com/org/repo/path/to/template[@REF] or path/to/template (in the airplanedev/templates repository)")
```
leeweisberger, line 48: What is [@REF]?

cmd/airplane/root/initcmd/init.go
```diff
@@ -0,0 +1,172 @@
...
+
+	return cmd
+}
+
+const taskOption = "Task		Create a SQL query, script, or API call"
+const viewOption = "View		Build an internal UI"
```
leeweisberger, line 54: Maybe just `Build a UI`?

cmd/airplane/root/initcmd/init.go
```diff
@@ -0,0 +1,172 @@
...
+const templateOption = "Template	Clone ready-made code to develop your tools faster"
```
leeweisberger, line 55: Maybe indiciate that this can be a task or a view?

Clone a ready-made task or view to start developing faster

leeweisberger (approved):

cmd/airplane/root/initcmd/init.go
rose-huang: you can clone from the folder/file at a specific commit. but this is prob confusing, i'll remove

cmd/airplane/root/initcmd/init.go
rose-huang: I don't want people to think theyre just making either 1 task or 1 view with each template. Gonna go with `Clone ready-made tasks and views to start developing faster`  for now, will polish later... it's technically `and/or` but that sounds kind of clunky

PR #594, opened 2022-08-19
Add command for creating demo db from CLI

zhan-airplane:
## Description
Add `airplane demo create-db` to create a demo DB resource. Note: if a resource with the given name already exists, it doesn't do anything.

## Test plan
Ran locally

---
cc @joshma


leeweisberger:
> ## Description
> 
> Add `airplane demo createdb` to create a demo DB resource. Note: if a resource with the given name already exists, it doesn't do anything.
> ## Test plan
> 
> Ran locally
> 
> cc @joshma

Before I review, what's the usecase? Do we expect users to run this? Are we going to tell them to run this?

cmd/airplane/demo/createdb/createdb.go
```diff
@@ -0,0 +1,50 @@
...
+	Name string
+}
+
+func New(c *cli.Config) *cobra.Command {
+	var cfg = Config{Root: c}
+
+	cmd := &cobra.Command{
+		Use:   "createdb",
+		Short: "Create demo DB",
+		Long:  "Creates a demo SQL DB resource",
```
leeweisberger, line 25: Can you indicate that if there is already a db associated with an account, this won't do anything

leeweisberger (approved):
Code looks fine, still not 100% about usecase though

joshma:
> Code looks fine, still not 100% about usecase though

@leeweisberger maybe we should discuss it first then? Here's why I wanted it: for various tutorials, and templates, we'll likely need them to initialize a demo resource. This'll have to be more composable than the current `--getting-started` flag. So in the near future, I'd like to replace `--getting-started` with

```
airplane demo create-db
airplane init --template=slug
```

does that sound right to you?

cmd/airplane/demo/createdb/createdb.go
```diff
@@ -0,0 +1,50 @@
...
+type Config struct {
+	Root *cli.Config
+	Name string
+}
+
+func New(c *cli.Config) *cobra.Command {
+	var cfg = Config{Root: c}
+
+	cmd := &cobra.Command{
+		Use:   "createdb",
```
joshma, line 23: minor but do we typically combine words or do we kebab case? 

```suggestion
		Use:   "create-db",
```

zhan-airplane: TBH I was trying to figure this out - the only multi-word command I could find in our CLI was `apikeys`, which was combined words (though arguments e.g. `--changed-files` are kebab-case). I'm fine with doing this either way?

joshma: yeah let's do kebab case!

PR #598, opened 2022-08-23
Warn user if missing resources

calufornia:
## Description

Warns the user if any task they're developing has resource attachments that aren't included in the dev config file, and suggests ways to add the resource to the dev config file. Creates a general `RegistrationWarning` struct that contains warnings encountered during task/view registration (e.g. missing resources, unsupported apps, etc.)

## Test plan
```
aplocal dev --editor /Users/justinlu/airplane/workflow-examples/builtins_node_example
Found 1 task:
- Builtins example

Found 0 views:

The following tasks have resource attachments that are not defined in the dev config file. Please add them through the previewer or run `airplane dev config set-resource`.
- Builtins example: [db rest]

```


pkg/server/warnings.go
```diff
@@ -0,0 +1,17 @@
+package server
+
+type UnsupportedApp struct {
```
kellymtrinh, line 3: Wonder if we can make these into some kind of error kind. Or I guess technically using error kinds and logging as we go would be most conventional but then RegisterTasksAndViews logs for both the server path and the `airplane dev` path :/

kellymtrinh (approved):

pkg/server/warnings.go
calufornia: Yeah..same reason why we don't like unsupported apps as we go, I decided to leave it up to the caller. Maybe once we remove the non-server path, we can add this

PR #601, opened 2022-08-24
Add local prompts

kellymtrinh:
## Description
Adds support for local prompts in the previewer. This implements the same internal/external prompt endpoints used in the SDK and web.

I decided to store the prompts on the run itself instead of keeping track of them separately (since updating/submitting prompts seems like the most complicated op here), but open to feedback.

## Test plan
-  Screen Shot 2022-08-25 at 2 28 39 PM

- Screen Shot 2022-08-25 at 2 25 04 PM


kellymtrinh:
Hmm actually might refactor update to allow returning an error

PR #605, opened 2022-08-26
Update local runstore.update to return errors

kellymtrinh:
## Description
I was implementing prompts and it would be easier if state.update was able to return errors instead of just a bool. Then the client can also decide to return errors or not.

## Test plan
- [x] Test that local execution still works (since it exercises that path)


colinking (commented):
🔥

colinking (approved):

PR #607, opened 2022-08-26
Add loading indicators for dependency installation steps in views

zhan-airplane:
## Description
Adds spinny loading indicators for dependency installation in `airplane views init` and `airplane views dev` - those are generally the slowest parts, so this should make it feel not as bad

## Test plan
Ran views init and dev locally

---
cc @stepzhou @leeweisberger @geodn


zhan-airplane:
ping @rose-huang

pkg/utils/copy.go
```diff
@@ -95,6 +95,10 @@ func CopyFromGithubPath(gitPath string) error {
 		if fsx.Exists(filepath.Join(directory, "package.json")) {
 			useYarn := ShouldUseYarn(directory)
 			logger.Step("Installing dependencies...")
+
+			loader := logger.NewLoader()
+			loader.Start()
+			defer loader.Stop()
```
rose-huang, line 101: the loading indicator remains after it finishes installing, could you look into getting rid of it so this looks like it did before?

```
- Copied getting_started/tasks/update_account.task.yaml
- Copied getting_started/tsconfig.json
- Installing dependencies...
⣯- Finished installing dependencies

⚡ To develop your view locally:
  airplane views dev getting_started
```

rose-huang (changes requested):
I think the normal `airplane views init` uses a different function to install dependencies (not CopyFromGithubPath) so can we add a loading indicator to that as well?

rose-huang (approved):
💯

PR #609, opened 2022-08-26
Add local prompts

kellymtrinh:
## Description
Adds support for local prompts in the previewer. This implements the same internal/external prompt endpoints used in the SDK and web. (It's split over ext and int api to match what we do in airport so we can reuse the hooks in web.)

I decided to store the prompts on the run itself instead of keeping track of them separately (since updating/submitting prompts seems like the most complicated op here), but open to feedback.

## Test plan
-  Screen Shot 2022-08-25 at 2 28 39 PM

- Screen Shot 2022-08-25 at 2 25 04 PM


pkg/server/apiext.go
```diff
@@ -81,7 +86,7 @@ type ExecuteTaskRequest struct {
 	Resources   map[string]string `json:"resources"`
 }
 
-func getParentID(r *http.Request) (string, error) {
+func getRunIDFromToken(r *http.Request) (string, error) {
```
colinking, line 89: nice

colinking (approved):

PR #613, opened 2022-08-30
Add required fields for builtin execution summary

kellymtrinh:
## Description
In order to show this modal, which is the execution summary for builtins, these are the required fields
Screen Shot 2022-08-29 at 9 30 36 PM

Depends on https://github.com/airplanedev/lib/pull/246

Runs needs to return an object with alias: resourceIDs
```
"resources": { "db": "res20220111zpqfl6j" },
    "resourceRequests": {},
    "timeout": 3600,
    "isStdAPI": true,
    "stdAPIRequest": {
      "namespace": "sql",
```

and then list resources needs to include more fields like the ID and kindConfig

```
{
  "resources": [
    {
      "id": "res20220103z3wp9eo",
      "slug": "res_airplane_api",
      "teamID": "tea20220103zvy4auu",
      "name": "Airplane API",
      "kind": "rest",
      "kindConfig": {
        "rest": {
          "baseURL": "https://api.airstage.app/",
          "headers": { "foo": "test" }
        }
      },
```

## Test plan
- [x] Tested that builtin execution correctly shows the summary for REST and SQL resources.


pkg/server/apiint.go
```diff
@@ -87,12 +88,26 @@ func GetResourceHandler(ctx context.Context, state *State, r *http.Request) (Get
 func ListResourcesHandler(ctx context.Context, state *State, r *http.Request) (libapi.ListResourcesResponse, error) {
 	resources := make([]libapi.Resource, 0, len(state.devConfig.RawResources))
 	for slug, resource := range state.devConfig.Resources {
-		// It doesn't matter what we include in the resource struct, as long as we include the slug - this handler
-		// is only used so that requests to the local dev api server for this endpoint don't error, in particular:
-		// https://github.com/airplanedev/lib/blob/d4c8ed7d1b30095c5cacac2b5c4da8f3ada6378f/pkg/deploy/taskdir/definitions/def_0_3.go#L1081-L1087
+		internalResource, err := conversion.ConvertToInternalResource(resource, state.logger)
```
kellymtrinh, line 91: Is there a better way to go from the Resource obj to a map[string]interface? 😓

pkg/server/apiint_test.go
```diff
@@ -47,14 +52,29 @@ func TestListResources(t *testing.T) {
...
 		{
 			Slug: "slack",
+			ID:   "r-2",
+			Kind: libapi.ResourceKind(kinds.ResourceKindSlack),
 		},
-	}, resp.Resources)
+	}
+
+	// sort so we can compare- since resources are stored as a map
+	sort.Slice(resp.Resources, func(i, j int) bool {
```
calufornia, line 69: Does it work to just compare `resp.Resources` and `expected` (i.e. `require.Equal(expected, resp.Resources)`)? I think that it should end up calling `reflect.DeepEqual` eventually: https://github.com/stretchr/testify/blob/181cea6eab8b2de7071383eca4be32a424db38dd/assert/assertions.go#L66

pkg/server/apiint.go
calufornia: This seems ok to me!

calufornia (approved):

pkg/server/apiint_test.go
kellymtrinh: It would, but there's a lot of non critical fields that I want to ignore/not update every time a resource struct changes

PR #616, opened 2022-08-31
Create dev config file on-demand and generate resource slug

calufornia:
## Description

If the user doesn't pass in a dev config file, creates one in the current directory on the first "write" operation (e.g. creating a new resource) - previously, the dev server would panic if attempting to create a resource when the dev config file didn't exist. Also generates a resource slug if one isn't provided to the resource creation endpoint - users should ideally be setting this from the UI to match what they have in their resource attachments, but we'll fall back to auto-generation if one isn't provided.

## Test plan
Tested locally by creating a new resource with no `airplane.dev.yaml` present, and verified that the dev config file was created with a slug generated from the name.


pkg/utils/slugs.go
```diff
@@ -71,3 +72,28 @@ func slugFromScript(file string) (string, error) {
...
+
+func GetUniqueSlug(req GetUniqueSlugRequest) (string, error) {
+	if req.MaxAttempts == 0 {
+		req.MaxAttempts = 10
+	}
+
+	// Make sure provided slug is actually a slug.
+	baseSlug := MakeSlug(req.Slug)
+	slug := baseSlug
+	for attempts := 0; attempts < req.MaxAttempts; attempts += 1 {
```
kellymtrinh, line 90: Just curious but does this happen a lot :rip: - could we change the RandomString implementation?

kellymtrinh (approved):

pkg/utils/slugs.go
calufornia: I think it's more likely to happen in our production API (which this is copied over from), where there could be a lot of resources with similar names - I don't expect it to happen that often in the local dev server.

PR #619, opened 2022-09-01
Add .gitignore to ignore .airplane-view folder on init

zhan-airplane:
## Description
Creates gitignore to ignore `.airplane-view` folder in init (both views init and copy-from-github-path init)

## Test plan
Ran locally

---
cc @leeweisberger @stepzhou @geodn


rose-huang (approved):

rose-huang (changes requested):
Oh I guess one thing - if we clone from a template and it doesn't have any views, it will still create this gitignore. Should we only make it in the CopyFromGithubPath command if the directory we clone has a .view.tsx file?

zhan-airplane:
Hmm I'm actually inclined to just create this `.gitignore` in general, just in case in the future users want to suddenly add views to their project... it's going to be semi-annoying to keep track of gitignore types for all the different things we could generate and I think it's relatively safe to just create this even for non-view stuff

On the other hand, if you have strong feelings about it, I think the next-best thing to do is to gitignore correctly specifically in each template in our `templates` github repo instead, though then there's a bit more of an onus on us (haha) to make sure every template we create is gitignoring correctly, wdyt?

rose-huang (approved):
ah okay that makes sense! so we would create this same git ignore during task init as well if we were to have to git ignore any files for tasks? I guess we can hold off on that for now

PR #621, opened 2022-09-01
Add support for resource slug editing from the editor

calufornia:
## Description

Adds the `/i/resources/isSlugAvailable` endpoint to the local dev server, which is now required to create and update resources. cc @kellymtrinh 

## Test plan
Tested updating and creating a new resource through the editor.


pkg/server/apiint.go
```diff
@@ -175,12 +190,41 @@ func UpdateResourceHandler(ctx context.Context, state *State, r *http.Request, r
...
+	if res, ok := state.devConfig.Resources[slug]; ok {
+		if res.ID() == id {
+			available = true
+		}
+	} else {
+		available = true
+	}
+
+	return IsResourceSlugAvailableResponse{
+		Available: available,
```
ericliuche, line 227: just `!ok || res.ID() == r.URL.Query().Get("id")`

ericliuche (approved):

PR #622, opened 2022-09-01
Skip registering tasks that have no registered runtime

kellymtrinh:
## Description
REST tasks currently are the only task type that doesn't have a runtime- they don't have a local entry point. Instead of erroring while registering tasks and views, we should skip REST tasks. Adds a debug statement in case we need to debug this later. The message to users for a task that doesn't support local execution or doesn't have a runtime is the same- since to them they're both unsupported.

Before:
Trying to start the server with some REST tasks would cause it to quit:
```
Found 2 tasks:
- hello world py
- rest task

Found 0 views:

Error: Did not find any runtimes for task type
```

After:
```
 localcli dev --editor /Users/kellytrinh/go/src/github.com/airplanedev/tasks-test
Found 2 tasks:
- test rest
- test  rest1

Found 0 views:

Skipping 2 unsupported tasks or views:
- test rest: rest task does not support local execution
- test  rest1: rest task does not support local execution

```

## Test plan
- [x] start server in a directory that has both rest and supported tasks


pkg/server/server.go
```diff
@@ -128,9 +128,14 @@ func (s *Server) RegisterTasksAndViews(taskConfigs []discover.TaskConfig, viewCo
...
 		r, err := runtime.Lookup(cfg.TaskEntrypoint, kind)
 		if err != nil {
-			return RegistrationWarnings{}, errors.Wrap(err, "looking up runtime")
+			// tasks with unregistered runtimes are unsupported
+			logger.Debug("%s does not support local execution: %v", cfg.Def.GetName(), err)
+			unsupported = append(unsupported, UnsupportedApp{
+				Name:   cfg.Def.GetName(),
+				Kind:   AppKindTask,
+				Reason: fmt.Sprintf("%v task does not support local execution", kind)})
+			continue
```
calufornia, line 137: Could you also add a `continue` under inside the `if r.SupportsLocalExecution()` statement?

calufornia (approved):

calufornia (approved):

PR #626, opened 2022-09-02
Support inline node code config init with inline flag

ameyakhare:
Resolves TSK-467

## Description
Adds the --inline flag to airplane tasks init that works for new and existing tasks.
Also makes the default extension for Node tasks .ts instead of .js.

## Test plan
Able to initialize a new task locally and from a task that was already deployed.


cmd/airplane/tasks/initcmd/init.go
```diff
@@ -226,25 +259,58 @@ func initWithTaskDef(ctx context.Context, cfg config) error {
...
 		question := fmt.Sprintf("Would you like to overwrite %s?", defnFilename)
 		if ok, err := utils.ConfirmWithAssumptions(question, cfg.assumeYes, cfg.assumeNo); err != nil {
-			return err
+			return writeDefnFileResponse{}, err
 		} else if !ok {
 			// User answered "no", so bail here.
-			return nil
+			return writeDefnFileResponse{
+				DefnFile: defnFilename,
+			}, nil
```
francescaleung, line 313: This feels incorrect somehow--like this should return `nil, nil` instead

cmd/airplane/tasks/initcmd/init.go
```diff
@@ -405,9 +467,15 @@ type suggestNextStepsRequest struct {
...
 	isNew              bool
+	inline             bool
 }
 
 func suggestNextSteps(req suggestNextStepsRequest) {
+	if req.inline {
+		req.defnFile = req.entrypoint
+	}
+
+	// Update the next steps for inline code config
```
francescaleung, line 478: this seems like it should get filled in

cmd/airplane/tasks/initcmd/init.go
```diff
@@ -476,6 +547,10 @@ func promptForEntrypoint(slug string, kind build.TaskKind, defaultEntrypoint str
 		}
 	}
 
+	if inline {
+		defaultEntrypoint = getInlineEntrypointName(slug, kind, defaultEntrypoint)
+	}
```
francescaleung, line 552: should this be in the `defaultEntrypoint == ""` case?

cmd/airplane/tasks/initcmd/init.go
```diff
@@ -106,6 +114,9 @@ func Run(ctx context.Context, cfg config) error {
 	if cfg.codeOnly && cfg.from == "" {
 		return errors.New("Required flag(s) \"from\" not set")
 	}
+	if cfg.codeOnly && cfg.inline {
+		return errors.New("Cannot specify both --code-only and --inline")
+	}
```
francescaleung, line 119: is it an error to specify inline with anything other than Node (right now) & Python (in the future)?

francescaleung (changes requested):
some small nits

cmd/airplane/tasks/initcmd/init.go
ameyakhare: yeah, but we don't know whether something is a node task until 1) we pull it (with --from defined) or 2) they select node for the task they want to create

cmd/airplane/tasks/initcmd/init.go
ameyakhare: we want this modification to occur whether there was no entrypoint passed in or if the entrypoint exists and should now be modified (existing node task)

cmd/airplane/tasks/initcmd/init.go
ameyakhare: made it return nil so that parent knows to eject without error

cmd/airplane/tasks/initcmd/init.go
ameyakhare: yeah I realized that we didn't actually need this at all here

ameyakhare: I think artifact of previous implementation

francescaleung (approved):
maybe extract out the `inlineCompatible` kinds? up to you

PR #628, opened 2022-09-02
Add version endpoint for dev server

kellymtrinh:
## Description
We want the editor to start showing a warning if the CLI is out of date. This way we can make breaking changes in the UI that depend on specific CLI versions being > some version.
 
## Test plan
- [x] Check new endpoint returns 
```
curl http://localhost:4000/dev/version | jq
{
  "status": "ok",
  "version": "<unknown>",
  "isLatest": true
}
```

- [x] The regular CLI version command is unaffected: 
```
❯ cli (kelly/TSK-464) ✘ localcli version
Version: <unknown> (<unknown>)
``` 
(it's expected that the local one is unknown)


pkg/version/latest/latest.go
```diff
@@ -24,20 +24,20 @@ type release struct {
 }
 
 // CheckLatest queries the GitHub API for newer releases and prints a warning if the CLI is outdated.
-func CheckLatest(ctx context.Context) {
+func CheckLatest(ctx context.Context) bool {
 	if version.Get() == "<unknown>" || version.Prerelease() {
 		// Pass silently if we don't know the current CLI version or are on a pre-release.
-		return
+		return true
```
calufornia, line 30: nit: Might make more sense to return false here and in error cases below instead

pkg/version/latest/latest.go
```diff
@@ -47,6 +47,7 @@ func CheckLatest(ctx context.Context) {
 		logger.Log(logger.Gray("  " + getUpgradeCommand()))
 		logger.Log("")
 	}
+	return latestWithoutPrefix == version.Get()
```
calufornia, line 50: Already doing the check above^
```suggestion
	return true
```

calufornia (approved):

PR #639, opened 2022-09-09
Generate package.json and tsconfig.json for tasks init

ameyakhare:
Resolves TSK-458

## Description
Creates package.json and tsconfig.json when Node task is initialized. Works fine with views since it merges changes to tsconfig/package.json

## Test plan
Initialized a task then initialized a view in the same directory.


pkg/node/scaffolding/taskTSConfig.json
```diff
@@ -0,0 +1,15 @@
+{
```
rose-huang, line 1: oops @zhan-airplane can you update the tsconfig for your template to use this one instead of the views tsconfig? 😅 I saw this after your template was merged

cmd/airplane/tasks/initcmd/init.go
```diff
@@ -793,3 +798,22 @@ func apiTaskToRuntimeTask(task *libapi.Task) *runtime.Task {
...
+
+func runKindSpecificInstallation(kind build.TaskKind) error {
+	switch kind {
+	case build.TaskKindNode:
+		cwd, err := os.Getwd()
+		if err != nil {
+			return errors.Wrap(err, "getting working directory")
+		}
+		if err := node.CreatePackageJSON(cwd, node.NodeDependencies{
+			Dependencies: []string{"airplane@next"},
```
colinking, line 810: It's live 🎉 

```suggestion
			Dependencies: []string{"airplane"},
```

pkg/node/scaffolding/taskTSConfig.json
```diff
@@ -0,0 +1,15 @@
...
+  "compilerOptions": {
+    "target": "ESNext",
+    "lib": ["ESNext"],
+    "allowSyntheticDefaultImports": true,
+    "forceConsistentCasingInFileNames": true,
+    "strict": true,
```
colinking, line 7: @rose-huang why do we enable `strict` mode for users? I would have expected we disabled it so that folks don't need to worry about having perfect TS types.

pkg/node/node.go
```diff
@@ -0,0 +1,311 @@
...
+		if parentName != "" {
+			keyName = fmt.Sprintf("%s.%s", parentName, key)
+		}
+		if existingSubMap, isSubMap := existingVal.(map[string]interface{}); isSubMap {
+			if printTSConfigChanges(newVal.(map[string]interface{}), existingSubMap, keyName) {
+				hasChanges = true
+			}
+		} else {
+			if !ok {
+				_, _ = w.Write([]byte(fmt.Sprintf("%s:\t(<absent>) -> (%v)\n", keyName, newVal)))
```
colinking, line 300: Minor from the demo -- could we JSON serialize these values? e.g. arrays of strings will look nicer

pkg/node/scaffolding/taskTSConfig.json
rose-huang: I think we wanted most of these checks associated with strict: https://www.typescriptlang.org/tsconfig#strict

I don't feel super strongly about this though 
cc @leeweisberger if you have any opinions

pkg/node/node.go
```diff
@@ -0,0 +1,311 @@
...
+func CreateViewTSConfig() error {
+	return mergeTSConfig(defaultViewTSConfig, MergeStrategyPreferTemplate)
+}
+
+//go:embed scaffolding/taskTSConfig.json
+var defaultTaskTSConfig []byte
+
+func CreateTaskTSConfig() error {
+	return mergeTSConfig(defaultTaskTSConfig, MergeStrategyPreferExisting)
+}
```
rose-huang, line 193: what's the reasoning for why view tsconfig prefers template and why task tsconfig prefers existing?

pkg/node/node.go
```diff
@@ -0,0 +1,311 @@
...
+
+type MergeStrategy string
+
+const (
+	MergeStrategyPreferExisting MergeStrategy = "existing"
+	MergeStrategyPreferTemplate MergeStrategy = "template"
+)
+
+func mergeTSConfig(configFile []byte, strategy MergeStrategy) error {
```
rose-huang, line 202: can you write a test for this (or the merge TS configs recursively function)? 😇 or if it's too complicated and we need to get this out, make a follow up ticket to make a test?

rose-huang (approved):
just a few comments but not blocking

pkg/node/node.go
ameyakhare: I want views to override anything the user has but dont want tasks to step over what the user has defined

pkg/node/scaffolding/taskTSConfig.json
colinking: Actually, this seems reasonable. I had thought config like https://www.typescriptlang.org/tsconfig#noImplicitReturns was included in `strict`, but it doesn't look like it does.

PR #641, opened 2022-09-09
Materialize env vars for local dev, switch envVars to configVars

ameyakhare:
Resolves TSK-610

## Description
1) Switches envVars to configVars since env/config will be set in the task definition file.
2) Properly materializes env vars from env literals and envs that reference configvars.

## Test plan
Tested this locally.


cmd/airplane/tasks/dev/config/delete_config/delete_env.go
```diff
@@ -1,4 +1,4 @@
-package delete_env
```
calufornia, line 1: Rename this file to `delete_config.go`

calufornia (approved):
Thanks for the fix - we'll also need to update our docs to reflect the changed field. https://docs.airplane.dev/dev-lifecycle/dev-config-file#environment-variables

PR #643, opened 2022-09-09
Auto create missing task / view if no prompt available

leeweisberger:
## Description

Context: A user requested that when they use the github action (which uses the CLI under the hood) they want tasks to be auto created. We do this for github deploys, but not for github action deploys since they use the CLI.

We could enable this with a flag, but we decided when doing github actions to automate task creation. If we're going to do it on that path, I don't see a reason why we wouldn't also do it on the CLI path. 

This change auto creates tasks if prompting is not supported. Thoughts?


francescaleung:
Does adding `--yes` to the github action do more than what we want it to do?

leeweisberger:
> Does adding `--yes` to the github action do more than what we want it to do?

You're right that's a better idea.

PR #646, opened 2022-09-12
Refactor local dev server code

calufornia:
## Description

Better organizes the local dev server codebase to make it easier to navigate and also so we can have different handlers for the same path, i.e. `/i/tasks/execute` vs `/v0/tasks/execute`, which we accomplish by splitting the api handlers into different packages (each test file is in its own package to avoid circular dependency issues). 

## Test plan
This PR is exclusively a refactor, and verified that existing behavior is unchanged.


kellymtrinh (approved):
🤩

PR #653, opened 2022-09-14
Add airplane runtime env vars to local dev

calufornia:
## Description

Adds most of the environment variables we've documented [here](https://docs.airplane.dev/tasks/runtime-api-reference#environment-variables) to the dev runtime. We omit `AIRPLANE_REQUESTER_EMAIL`, `AIRPLANE_REQUESTER_ID`, `AIRPLANE_SESSION_ID`, `AIRPLANE_TASK_REVISION_ID`, and `AIRPLANE_TRIGGER_ID`.

Depends on https://github.com/airplanedev/lib/pull/283

## Test plan

Created a task the printed out `process.env` and verified the environment variables below were included.


cmd/airplane/tasks/dev/server.go
```diff
@@ -30,9 +30,22 @@ func runLocalDevServer(ctx context.Context, cfg taskDevConfig) error {
 		TeamID: cfg.root.Client.TeamID,
 	}
 
+	authInfo, err := cfg.root.Client.AuthInfo(ctx)
+	if err != nil {
+		return err
+	}
+
+	var envID string
+	if cfg.envSlug != "" {
```
colinking, line 39: We'll need to handle the default environment, which is denoted by an empty string for a slug.

pkg/dev/dev.go
```diff
@@ -362,9 +364,31 @@ func entrypointFromDefn(file string) (string, error) {
 
 func appendAirplaneEnvVars(env []string, config LocalRunConfig) ([]string, error) {
 	env = append(env, fmt.Sprintf("AIRPLANE_API_HOST=http://127.0.0.1:%d", config.Port))
-	env = append(env, "AIRPLANE_RUNTIME=dev")
+	env = append(env)
```
colinking, line 367: undo?

pkg/dev/dev.go
```diff
@@ -362,9 +364,31 @@ func entrypointFromDefn(file string) (string, error) {
...
+		runnerID = config.AuthInfo.User.ID
+		runnerEmail = config.AuthInfo.User.Email
+	}
+
+	var teamID string
+	if config.AuthInfo.Team != nil {
+		teamID = config.AuthInfo.Team.ID
+	}
+
+	env = append(env,
```
colinking, line 381: Can you link to the runtime env var docs here? And list out the env vars that we aren't setting here + why, so it's easier to identify if we're missing others.

colinking (commented):
(will defer to @kellymtrinh)

pkg/dev/dev.go
calufornia: Moved with the other env vars below since it's one of the documented environment variables.

kellymtrinh (approved):
Nice!!!

PR #656, opened 2022-09-14
Add vite debug logs

leeweisberger:
Add vite logs as debug logs so we can tell what's happening on the vite server.


pkg/views/dev.go
```diff
@@ -242,6 +242,14 @@ func runVite(opts ViteOpts, tmpdir string) (*exec.Cmd, string, error) {
 	if err = cmd.Start(); err != nil {
 		return nil, "", err
 	}
+	logger.Debug("Started vite with process id: %v", cmd.Process.Pid)
+
+	go func() {
```
calufornia, line 247: I think you want to start this goroutine after `wg.Wait()` - I tested this locally and the view doesn't load in the editor

calufornia: Also, you probably want to do this only if `!opts.TTY` - in the interactive case, we're already logging debug logs: https://github.com/airplanedev/cli/blob/ac6555d65be1c2a505ba9ba689c59fb50ca05a12/pkg/views/dev.go#L222-L224

leeweisberger: Yeah I misread good catch. Added some comments for posterity

leeweisberger:
I realized this actually fixes the problem.

Talked to @ericliuche and scanner has a max buffer size. Since we are never flushing the buffer, vite crashes after the buffer fills up :hand:

kellymtrinh:
> I realized this actually fixes the problem.
> 
> Talked to @ericliuche and scanner has a max buffer size. Since we are never flushing the buffer, vite crashes after the buffer fills up ✋

Wait does this mean it's not because of erroring/invalid code? (Just that that makes the buffer fill up faster??)

leeweisberger:
> > I realized this actually fixes the problem.
> > Talked to @ericliuche and scanner has a max buffer size. Since we are never flushing the buffer, vite crashes after the buffer fills up ✋
> 
> Wait does this mean it's not because of erroring/invalid code? (Just that that makes the buffer fill up faster??)

Exactly!

calufornia (approved):

PR #659, opened 2022-09-15
Search for dev config file in editor work directory

calufornia:
## Description

The CLI now attempts to automatically find a dev config file in the editor's dev root (i.e. the argument that's passed in to `airplane dev --editor [dev root]`). Previously, the CLI would attempt to look for a file called `airplane.dev.yaml` in the directory where the airplane CLI was being run - now, it looks in the dev root and traverses through parent directories until it reaches `/`. If none is found, the user is prompted to create one in the dev root. 

The user is prompted to create a dev config file only if they run `airplane dev` with the `--editor` flag - users are not prompted in the legacy usage of `airplane dev` (which generally does not require the dev config file).

## Test plan
```
➜  aplocal dev --editor ~/airplane/node_test
? Dev config file not found - would you like to create one? Yes
? Where would you like to create the dev config file? /Users/justinlu/airplane/node_test/airplane.dev.yaml
Created dev config file at /Users/justinlu/airplane/node_test/airplane.dev.yaml
Loaded dev config file at /Users/justinlu/airplane/node_test/airplane.dev.yaml
Found 3 tasks:
- Country lookup
- Find user
- Generate report

Found 2 views:
- Admin panel
- test_view

^Ctrap: received "interrupt"
```

Moving dev config file into parent directory:

```
➜ mv ~/airplane/node_test/airplane.dev.yaml ~/airplane
➜ aplocal dev --editor ~/airplane/node_test
Loaded dev config file at /Users/justinlu/airplane/airplane.dev.yaml
Found 3 tasks:
- Country lookup
- Find user
- Generate report

Found 2 views:
- Admin panel
- test_view

^Ctrap: received "interrupt"
```

Testing with no explicit dev root:

```
(⎈ |stage:airplane)➜  cli git:(justin/find-dev-config) ✗ aplocal dev --editor
Loaded dev config file at /Users/justinlu/go/src/github.com/airplanedev/cli/airplane.dev.yaml
...
```


cmd/airplane/tasks/dev/dev.go
```diff
@@ -72,7 +73,33 @@ func New(c *cli.Config) *cobra.Command {
 				} else if len(args) > 1 {
 					return errors.New("Multiple dev server roots detected, please supply only one directory to discover tasks and views")
 				} else {
+					// Use absolute path to dev root to allow the local dev server to more easily calculate relative paths.
 					cfg.fileOrDir = args[0]
+					if cfg.fileOrDir, err = filepath.Abs(cfg.fileOrDir); err != nil {
+						return err
```
colinking, line 79: wrap the error

cmd/airplane/tasks/dev/dev.go
```diff
@@ -72,7 +73,33 @@ func New(c *cli.Config) *cobra.Command {
...
+					}
+				}
+
+				if cfg.devConfigPath == "" {
+					var devDir string
+					// Calling filepath.Dir on a directory path that doesn't end in '/' returns that directory's parent -
+					// we want to return the directory itself in that case, and so we check if cfg.fileOrDir is a directory.
```
colinking, line 86: nice

cmd/airplane/tasks/dev/dev.go
```diff
@@ -72,7 +73,33 @@ func New(c *cli.Config) *cobra.Command {
...
+					}
+
+					// Recursively search for dev config file from the dev dir until root.
+					devConfigDir, ok := fsx.Find(devDir, conf.DefaultDevConfigFileName)
+					if !ok {
+						// If a dev config file is not found, set the dev config dir to the dev root and prompt for creation
+						// of the file below.
+						devConfigDir = devDir
+					}
+					cfg.devConfigPath = filepath.Join(devConfigDir, conf.DefaultDevConfigFileName)
```
colinking, line 102: Will this be an absolute path? If so, let's convert `devConfigPath` to an absolute path when explicitly set, too.

cmd/airplane/tasks/dev/dev.go
```diff
@@ -72,7 +73,33 @@ func New(c *cli.Config) *cobra.Command {
...
+					// we want to return the directory itself in that case, and so we check if cfg.fileOrDir is a directory.
+					if info, err := os.Stat(cfg.fileOrDir); err != nil {
+						return err
+					} else if info.IsDir() {
+						devDir = cfg.fileOrDir
+					} else {
+						devDir = filepath.Dir(cfg.fileOrDir)
+					}
+
+					// Recursively search for dev config file from the dev dir until root.
```
colinking, line 95: Root is confusing (user root vs. task root)

```suggestion
					// Recursively search for dev config file, starting from the dev dir.
```

cmd/airplane/tasks/dev/dev.go
```diff
@@ -144,54 +165,24 @@ func run(ctx context.Context, cfg taskDevConfig) error {
...
+		DevConfig: cfg.devConfig,
 	})
 	if err != nil {
 		return errors.Wrap(err, "starting local dev api server")
 	}
 
 	defer func() {
 		if err := apiServer.Stop(context.Background()); err != nil {
-			l.Error("failed to stop local api server: %+v", err)
+			logger.Error("failed to stop local api server: %+v", err)
```
colinking, line 181: Why this change?

pkg/conf/dev.go
```diff
@@ -161,6 +161,45 @@ func WriteDevConfig(config *DevConfig) error {
 	return nil
 }
 
+// LoadDevConfigFile attempts to load in the dev config file at devConfigPath, and prompts the user to create one if it
+// does not exist.
```
colinking, line 165: Hmm, IMO we shouldn't prompt the user to create a dev config file. Can we lazily create it? e.g. once a user tries to create a resource--via the CLI or via the UI -- then we should create it.

pkg/conf/dev.go
```diff
@@ -161,6 +161,45 @@ func WriteDevConfig(config *DevConfig) error {
...
+			}
+		} else {
+			devConfigLoaded = true
+		}
+	}
+
+	if devConfigLoaded {
+		logger.Log("Loaded dev config file at %s", devConfigPath)
+	} else {
+		logger.Log("Using empty dev config")
```
colinking, line 196: Does this log mean anything to a user? Do we need it?

pkg/conf/dev.go
```diff
@@ -161,6 +161,45 @@ func WriteDevConfig(config *DevConfig) error {
...
+			} else {
+				return nil, errors.Wrap(err, "unable to read dev config")
+			}
+		} else {
+			devConfigLoaded = true
+		}
+	}
+
+	if devConfigLoaded {
+		logger.Log("Loaded dev config file at %s", devConfigPath)
```
colinking, line 194:
```suggestion
		logger.Log("Loaded dev config from %s", devConfigPath)
```

colinking (changes requested):

pkg/conf/dev.go
calufornia: Changed to a debug log.

pkg/conf/dev.go
calufornia: Moved dev config file creation to `WriteDevConfig` - the first time that a resource/config is written to it it'll create the dev config file.

cmd/airplane/tasks/dev/config/set_resource/set_resource.go
```diff
@@ -152,11 +151,6 @@ func run(ctx context.Context, cfg config) error {
 	if err := devConfig.SetResource(cfg.slug, resource); err != nil {
 		return errors.Wrap(err, "setting resource in dev config file")
 	}
-	devConfig.RawResources = append(devConfig.RawResources, serializedResource)
-
-	if err := conf.WriteDevConfig(devConfig); err != nil {
```
calufornia, line 157: `SetResource` already persists the dev config file: https://github.com/airplanedev/cli/blob/4804dce5fb7ef11dd155253530dad58817253ce4/pkg/conf/dev.go#L71

cmd/airplane/tasks/dev/config/config.go
```diff
@@ -27,40 +26,33 @@ func New(c *cli.Config) *cobra.Command {
...
+			airplane dev config delete-configvar API_KEY
 			airplane dev config set-resource --kind postgres db
 			airplane dev config delete-resource db
 		`),
-		PersistentPreRunE: utils.WithParentPersistentPreRunE(func(cmd *cobra.Command, args []string) error {
+		// The subcommands of `airplane dev config` don't have PersistentPreRunE functions, and so they automatically
+		// inherit the parent run's PersistentPreRunE. If we wrapped this in utils.WithParentPersistentPreRunE, this
+		// would cause the dev config file to get loaded twice (once by the subcommand, and once by the parent command).
+		// This command doesn't need any of its parent's PersistentPreRunE checks (e.g. checking login), and so we omit
+		// the wrapper here.
```
colinking, line 38: good catch, this is confusing. I'd love to replace cobra with something easier/prettier at some point 😛

cmd/airplane/tasks/dev/config/config.go
```diff
@@ -27,40 +26,33 @@ func New(c *cli.Config) *cobra.Command {
...
 				wd, err := os.Getwd()
 				if err != nil {
 					return errors.Wrap(err, "error determining current working directory")
 				}
-
-				defaultPath := filepath.Join(wd, conf.DefaultDevConfigFileName)
-				if _, err := os.Stat(defaultPath); err == nil {
-					cfg.Filepath = defaultPath
-				} else {
-					path, err := conf.PromptDevConfigFileCreation(conf.DefaultDevConfigFileName)
```
colinking, line 46: This should still prompt to create the dev config, right? IIUC, looks like LoadDevConfigFile will error if it doesn't exist yet.

pkg/conf/dev.go
```diff
@@ -154,37 +152,48 @@ func WriteDevConfig(config *DevConfig) error {
...
+				return errors.Wrap(createErr, "creating dev config file")
+			}
+			f.Close()
+			logger.Log("Created dev config file at %s", config.Path)
+		} else {
+			return errors.Wrap(err, "checking if dev config file exists")
+		}
+	}
+
 	if err := os.WriteFile(config.Path, buf, 0600); err != nil {
```
colinking, line 168:
```suggestion
	if err := os.WriteFile(config.Path, buf, 0644); err != nil {
```

colinking (approved):

cmd/airplane/tasks/dev/config/config.go
calufornia: Ben mentioned https://github.com/segmentio/cli 🙂

cmd/airplane/tasks/dev/config/config.go
calufornia: It shouldn't error! We check `if !errors.Is(err, ErrMissing)` in `LoadDevConfigFile`, and only return an error in that case - otherwise, we return an empty dev config file with the default path so that it gets lazily created.

PR #662, opened 2022-09-16
Add an appropriate title to the webpage produced by `airplane views dev`

geodn:
## Description
Parsing the slug is difficult because the user may be using yaml or an inline config. In the meantime, let's set the title to be the base filename of the entrypoint, without the extension. E.g. for an entrypoint file `adminPanel/admin_panel.view.tsx`, we set the title to admin_panel.

## Test plan
```
go install ./cmd/airplane
airplane views dev
cat .airplane-view/index.html
```


zhan-airplane (changes requested):
Hmm I think it's a little brittle to do this parsing manually, can we change `IndexHtmlString()` to accept a title param instead? e.g. like in `MainTsxString()` we take an entrypoint module param

zhan-airplane (approved):

PR #665, opened 2022-09-16
Bump lib version

ericliuche:
## Description
Adds python inline config

## Test plan
ran inline python config deploy


ericliuche:
@calufornia also includes your most recent lib change, lmk if this is okay

PR #666, opened 2022-09-16
POC: Local demo DB via postlite

colinking:
## Description

POC showing how the editor could run a local version of the Demo DB using [SQLite](https://www.sqlite.org/index.html), exposed as client/server via a new PostgreSQL proxy called [Postlite](https://github.com/benbjohnson/postlite).

This currently loads the Northwind DB, which was already available in SQLite format. We will need to replace this with the real Demo DB contents.

## Test plan

Create a task like so:

```typescript
import airplane from "airplane"

export default airplane.workflow(
  {
    slug: "workflows_getting_started",
    name: "Workflows getting started",
    nodeVersion: "18",
    resources: ["demo_db"]
  },
  async () => {
    const run = await airplane.sql.query("demo_db", `SELECT * FROM Categories`)

    return run.output
  }
)

```

Execute it in the UI via the editor. You'll also see the Demo DB show as a resource.


colinking:
* main
  * **PR #666** <a href="https://app.graphite.dev/github/pr/airplanedev/cli/666" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/cli/666?utm_source=stack-comment).

PR #667, opened 2022-09-19
Allow local views to execute local tasks

calufornia:
## Description

Adds the ability to execute a local task when developing a view in the editor. We introduce a `--local` flag to the `airplane dev` command, which defaults to true and indicates that any task referenced by a view should be executed locally instead of in one of the user's team's environments. If `--local=false`, the task will be executed in the user's default environment (usually production), or if an `--env` is specified, then the designated environment.

## Test plan
Tested by creating a view with a task-backed component and verifying that the view called the local task and the component's data was populated correctly.


cmd/airplane/tasks/dev/server.go
```diff
@@ -18,28 +18,38 @@ import (
...
 func runLocalDevServer(ctx context.Context, cfg taskDevConfig) error {
+	authInfo, err := cfg.root.Client.AuthInfo(ctx)
+	if err != nil {
+		return err
+	}
+
+	var envID, envSlug string
+	if cfg.local {
+		// Use a special env ID and slug when executing tasks locally and the `--env` flag doesn't apply.
+		envID = "local"
```
kellymtrinh, line 29: NIT: move `"local"` into a const

kellymtrinh (approved):

PR #670, opened 2022-09-19
Update view scaffolding and release root level init

rose-huang:
## Description
- New scaffolding to be more relevant than hydrogen helium example
- Removes --dev flag on root level init so it will switch over from defaulting to task to asking for task, view, or template

## Test plan

new scaffolding:
Screen Shot 2022-09-19 at 7 37 29 PM

I built CLI and init-ed a new view and made sure it worked with dev and had no type errors

cc @joshma  on releasing the new root level command


joshma:
nice. can you also update the task scaffolding? 😇 just have it return the table of customers

cmd/airplane/views/initcmd/scaffolding/default.view.tsx
```diff
@@ -1,43 +1,134 @@
...
+const ExampleView = () => {
+  const customersState = useComponentState("customers");
+  const selectedCustomer = customersState.selectedRow;
+
+  return (
+    <Stack>
+      <Title>Customer overview</Title>
+      <Text>An example view that showcases customers and users.</Text>
+      <Table
+        id="customers"
```
leeweisberger, line 14: Let's do `id={customState.id}`. And then remove `"customers"` above no need to use a string

cmd/airplane/views/initcmd/scaffolding/default.view.tsx
```diff
@@ -1,43 +1,134 @@
...
+          columns={usersCols}
+          hiddenColumns={["customer_id"]}
+        />
+      )}
+    </Stack>
+  );
 };
 
-const data: Row[] = [
+// Example data: replace with your own data or with a task-backed table
```
leeweisberger, line 33: Maybe link to table docs? What do you think?

leeweisberger (approved):

cmd/airplane/views/initcmd/scaffolding/default.view.tsx
rose-huang: @leeweisberger I updated the comment and linked to task backed components in case people aren't trying to use the table component

PR #672, opened 2022-09-20
CLI QOL improvements

ameyakhare:
Resolves TSK-712

## Description
1. Allows logging in with a token provided on the cmdline
2. Allows initializing a task without entrypoint prompt if -y flag and entrypoint file specified.

## Test plan
image


cmd/airplane/tasks/initcmd/init.go
```diff
@@ -217,9 +217,14 @@ func initWithTaskDef(ctx context.Context, cfg config) error {
 		}
 
 		for {
-			entrypoint, err = promptForEntrypoint(def.GetSlug(), kind, entrypoint, cfg.inline)
-			if err != nil {
-				return err
+
+			if cfg.assumeYes && cfg.file != "" {
```
calufornia, line 221: Error if `cfg.file == ""`?

calufornia (approved):

cmd/airplane/tasks/initcmd/init.go
ameyakhare: it shouldn't break existing flows, where the -y flag without the cfg file should still work

PR #673, opened 2022-09-21
Add ability to execute local tasks against remote resources

calufornia:
## Description

Adds the ability to set a "fallback" environment to discover resources in. We do this by calling the `/v0/resources/getWithCredentials` endpoint to return a resource with sensitive data unscrubbed. We do this on-the-fly whenever the user executes a run through the editor, so that they don't need to save it to their dev config file.

This is a relatively common use case if the user doesn't want to configure their resource locally, and for getting started guides that require the demo db resource (and would be tedious for the user to setup a local postgres instance).

If the user already has a local resource with the same slug as a remote resource, even if the fallback environment is set, the local resource takes precedence, which is the basis for the "merging" behavior seen below. Note that this still requires that the remote resource be accessible from the user's laptop (or wherever they're running their code).

We utilize the existing `--env` flag for `airplane dev --editor` to control the fallback behavior - if no `--env` flag is specified, then only resources will be used.

cc @kellymtrinh 

Depends on: https://github.com/airplanedev/lib/pull/295

## Test plan
Tested the following scenarios:
- [x] No resource in dev config file, no resource in remote, with `--env` flag
- [x] No resource in dev config file, resource in remote, no `--env` flag, verified that remote resource was _not_ used
- [x] No resource in dev config file, resource in remote, with `--env` flag, verified that remote resource was used
- [x] Resource in dev config file, no resource in remote, no `--env` flag, verified that local resource was used
- [x] Resource in dev config file, resource in remote, no `--env` flag, verified that local resource was used
- [x] Resource in dev config file, resource in remote, with `--env` flag, verified that local resource was used even though env flag was specified.

```
aplocal dev --editor ~/airplane/node_test --env ''
Loaded dev config from /Users/justinlu/airplane/node_test/airplane.dev.yaml
Found 3 tasks:
- Country lookup
- Find user
- Generate report

Found 2 views:
- Admin panel
- test_view

Your environment is set to prod:
- Any tasks referenced by your views will execute remotely in that environment.
- Any resource used by a task or builtin will be pulled from prod, unless a resource with the same slug exists in your dev config file.
- To instead run all tasks locally, run `airplane dev --editor` without the `--env` flag.

```

```
aplocal dev --editor ~/airplane/node_test
Loaded dev config from /Users/justinlu/airplane/node_test/airplane.dev.yaml
Found 3 tasks:
- Country lookup
- Find user
- Generate report

Found 2 views:
- Admin panel
- test_view

The following tasks have resource attachments that are not defined in the dev config file or remotely in local. Please add them through the editor or run `airplane dev config set-resource`.
- Country lookup: [prayge]

Your environment is set to local:
- Any tasks referenced by your views will execute locally.
- Any resource used by a task or builtin will be from your dev config file.
- To instead run tasks in aremote environment, run `airplane dev --editor --env <env>`.

Visit http://127.0.0.1:4000/editor?host=http://localhost:4000 for a development UI.
```


pkg/dev/run.go
```diff
@@ -41,6 +41,7 @@ func NewLocalRun() *LocalRun {
 	return &LocalRun{
 		Status:      api.RunQueued,
 		ParamValues: map[string]interface{}{},
+		Parameters:  &libapi.Parameters{},
```
calufornia, line 44: Unrelated to this PR, but I noticed that sometimes the run details page fails to load because of parameters being null if the run is still in the Queued state.

colinking:
FYI, the run page doesn't render correctly for built-ins executed against remote resources:

CleanShot 2022-09-21 at 11 58 46

Though I figure this is a known issue (handling remote resources in the UI) for a future PR.

cmd/airplane/tasks/dev/server.go
```diff
@@ -145,21 +144,27 @@ func runLocalDevServer(ctx context.Context, cfg taskDevConfig) error {
...
 	} else {
-		logger.Log("Your environment is set to %s - any tasks referenced by your views will execute remotely in that environment. To run all tasks locally, run `airplane dev --editor` without the `--env` flag.", logger.Bold(envSlug))
+		logger.Log("Your environment is set to %s:\n- Any tasks referenced by your views will execute remotely in "+
+			"that environment.\n- Any resource used by a task or builtin will be pulled from %s, unless a resource with "+
+			"the same slug exists in your dev config file.\n- To instead run all tasks locally, run "+
+			"`airplane dev --editor` without the `--env` flag.", logger.Bold(envSlug), logger.Bold(envSlug))
 	}
 
+	logger.Log("")
+	logger.Log("Visit %s for a development UI.", logger.Bold(fmt.Sprintf("%s/editor?host=http://localhost:%d", cfg.root.Client.AppURL(), cfg.port)))
```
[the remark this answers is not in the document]
calufornia: Yep, defaulting to `http://localhost:4000` sounds reasonable, and defaulting to `localhost` if only `port` is specified also sounds reasonable. We'll also still have the `host` case in case the host isn't localhost in the future.

cmd/airplane/tasks/dev/dev.go
```diff
@@ -229,7 +229,20 @@ func run(ctx context.Context, cfg taskDevConfig) error {
 		return err
 	}
 
-	resources, err := resource.GenerateAliasToResourceMap(taskConfig.Def.GetResourceAttachments(), cfg.devConfig.Resources)
+	slugToResource := make(map[string]resource.ResourceWithEnv)
+	for slug, res := range cfg.devConfig.Resources {
+		slugToResource[slug] = resource.ResourceWithEnv{
```
kellymtrinh, line 234: Maybe we can set this in `LoadDevConfigFile` when it's first loaded?

cmd/airplane/tasks/dev/server.go
```diff
@@ -145,21 +144,27 @@ func runLocalDevServer(ctx context.Context, cfg taskDevConfig) error {
...
 	}
 
 	logger.Log("")
-	logger.Log("Visit %s/editor?host=http://localhost:%d for a development UI.", cfg.root.Client.AppURL(), cfg.port)
 	if envID == dev.LocalEnvID {
-		logger.Log("Your environment is set to %s - any tasks referenced by your views will execute locally. To run tasks in a remote environment, run `airplane dev --editor --env <env>`.", logger.Bold(dev.LocalEnvID))
+		logger.Log("Your environment is set to %s:\n- Any tasks referenced by your views will execute locally.\n- "+
+			"Any resource used by a task or builtin will be from your dev config file.\n- To instead run tasks in a"+
+			"remote environment, run `airplane dev --editor --env <env>`.", logger.Bold(dev.LocalEnvID))
 	} else {
```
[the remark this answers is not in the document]
kellymtrinh: I think this is more clear as "this is your default environment. all tasks and resources will be executed from this environment, unless it is overridden by a local task or resource"? It's more familiar to me to say something is the default and have overrides, then that local tasks/resources are required and will fallback. But I will defer to you/everyone else because we already discussed this and the general consensus seemed to be around "fallback" envs

pkg/resource/resources.go
```diff
@@ -1,33 +1,100 @@
...
+		var resource resources.Resource
+		resourceWithEnv, ok := slugToResource[slug]
 		if !ok {
-			// TODO: Augment error message with airplane subcommand to add resource
-			return nil, errors.Errorf("Cannot find resource with slug %s in dev config file", slug)
+			return nil, errors.Errorf("Cannot find resource with slug %s in dev config file or remotely", slug)
+		} else if resourceWithEnv.Remote {
+			remoteResourceWithCredentials, err := state.CliConfig.Client.GetResource(ctx, api.GetResourceRequest{
+				Slug:                 slug,
+				EnvSlug:              state.EnvSlug,
```
kellymtrinh, line 43: Should we do a check because `state` can be nil? so this doesn't nil pointer error in the future- one of the usages of `GenerateAliasToResourceMap` in `cmd/airplane/tasks/dev/dev.go` passes in no state, even though in that use case it wouldn't have remote resources yet.

kellymtrinh (approved):

PR #677, opened 2022-09-22
Rewrite package.json with dev deps

leeweisberger:
## Description
Rather than copy the root package.json into the .airplane-views dev folder, only copy development dependencies. This way additional dependencies will be used from the root node_modules

This fixes the bug where the root dependencies (used when you deploy the view) are different from the dev dependencies. Now the dev folder and root folder are not sharing dependencies.

## Test plan
I manually tested devving


pkg/views/dev.go
```diff
@@ -156,41 +162,52 @@ func createViteConfig(tmpdir string) error {
...
 		"react":           "18.0.0",
 		"react-dom":       "18.0.0",
-		"@airplane/views": "*",
+		"@airplane/views": "^1.0.0",
 	}
 	defaultDevDeps := map[string]string{
-		"@vitejs/plugin-react": "^2.0.0",
-		"vite":                 "^3.0.0",
+		"@vitejs/plugin-react": "2.1.0",
+		"vite":                 "3.1.3",
```
leeweisberger, line 192: I'm locking these to avoid breaking vite changes breaking airplane dev. We should probably put these in their own package.json file and use renovate so we stay up to date

pkg/views/dev.go
```diff
@@ -200,7 +217,7 @@ type ViteOpts struct {
 }
 
 func runVite(ctx context.Context, opts ViteOpts, tmpdir string) (*exec.Cmd, string, error) {
-	cmd := exec.Command("node_modules/.bin/vite", "dev")
+	cmd := exec.Command("npx", "vite", "dev")
```
leeweisberger, line 220: npx has been around since npm 5.2 which shipped with node 12 so should be safe to use.

The only time this wouldn't be safe is if someone installed yarn without npm which is unlikely. We can add npm as a pre-req which seems reasonable.

pkg/views/dev.go
```diff
@@ -74,20 +74,26 @@ func Dev(ctx context.Context, v viewdir.ViewDirectory, viteOpts ViteOpts) (*exec
...
-	}
-	if err := json.NewEncoder(newPackageJSONFile).Encode(packageJSON); err != nil {
-		return nil, "", errors.Wrap(err, "writing new package.json")
-	}
-	if err := newPackageJSONFile.Close(); err != nil {
-		return nil, "", errors.Wrap(err, "closing new package.json file")
+	if depsAdded {
+		// Write package.json back out.
+		l.Debug("rewriting package.json")
+		newPackageJSONFile, err := os.Create(filepath.Join(root, "package.json"))
```
zhan-airplane, line 85: Wait don't we still need to copy this back to the temp dir?

Also, if we're doing a rewrite, can we do it indirectly? i.e. write to a new file, then move it onto the old file. Otherwise if e.g. the user ctrl-Cs in the middle of this, their `package.json` is going to be corrupted

leeweisberger: I updated to strategy a bit. We're still writing a package.json folder into the temp directory and installing it. But rather than copy the root package.json in, we are only adding the dependencies that are needed for development and that aren't already included in the root package.json.

This means that most of the time the .airplane-views package will only need to install vite. Deps like airplane/views will be read from the root node_modules

zhan-airplane (approved):
Hm so just to confirm, after this `.airplane-view` should only really contain the `main.tsx` and `index.html` stub files used when dev-ving?

leeweisberger:
> Hm so just to confirm, after this `.airplane-view` should only really contain the `main.tsx` and `index.html` stub files used when dev-ving?

We are creating the same content as we were before - so that would be the main.tsx and index.html shims.
We are now only installing dependencies needed for development. This would be react (if not already installed - it probably will be), airplane/views (if not already installed - it probably will be), and vite (we install this even if the user for some reason already has this so that we know we have a version that will successfully dev the project). All other dependencies are grabbed from the parent node_modules. 

The change is that we aren't copying over their whole package.json and installing everything - only what is needed for development

PR #679, opened 2022-09-22
Execute task asynchronously

calufornia:
## Description

This makes task execution async in the `/v0/tasks/execute` handler, which is what we do in our production API. The execute handler is now responsible for initializing the run and kicking off execution in a separate goroutine, so that it can return instantly. This is necessary for executing remote tasks from the editor, which we will do by proxying the request through the local dev server.

Also removes the `/dev/runs/create` endpoint, which is no longer needed after this change.

## Test plan
Tested by executing a task locally and verifying it succeeded.


kellymtrinh (approved):

colinking:
CodeCov is down so the tests CI step failed. The tests themselves passed, only uploading to CodeCov failed. I'm force merging to bypass.

PR #680, opened 2022-09-23
Properly set output when task errors

calufornia:
## Description

We weren't properly setting the output in the case that a task errored - previously, if the command to run a task's code errored, we would always return empty outputs, but we change it to return the output even in the case that the command errors.

## Test plan
image


colinking (approved):
good catch :)

PR #682, opened 2022-09-23
Move builtins to lib

kellymtrinh:
## Description
Depends on https://github.com/airplanedev/lib/pull/301
Moves the builtin client from CLI to lib. The PR to wrap around the REST and SQL top level task execution is going to be implemented here: https://github.com/airplanedev/lib/blob/cbcb25d215bf3758057df21bc38d63c924fc9517/pkg/runtime/sql/sql.go in lib, so it makes sense to move it there and share the code.

- [x] TODO: release lib and use that version
## Test plan 
- [x] test that executing a builtin still works in the CLI


go.mod
```diff
@@ -3,10 +3,9 @@ module github.com/airplanedev/cli
 go 1.19
 
 require (
-	cloud.google.com/go/storage v1.26.0
 	github.com/AlecAivazis/survey/v2 v2.3.6
 	github.com/MakeNowJust/heredoc v1.0.0
-	github.com/airplanedev/lib v0.5.70
+	github.com/airplanedev/lib v0.5.71-0.20220923233449-8c27735a6058
```
kellymtrinh, line 8: Reminder to self to update to released version of lib after https://github.com/airplanedev/lib/pull/301 is merged

PR #686, opened 2022-09-26
Add no views found error message

zhan-airplane:
## Description
Add better error message for no views being found

## Test plan
None


yolken-airplane (approved):
Thanks!!

PR #688, opened 2022-09-27
Automatically make demo db accessible

calufornia:
## Description

Similar to the Slack resource, automatically makes the demo db resource accessible without needing to provide a fallback env. The demo db is created for a team upon team creation, and so its slug should be a canonical `demo_db`. The only case where the demo db wouldn't exist is if this initial creation failed, in which case the demo db simply isn't added to the merged resources.

Depends on https://github.com/airplanedev/lib/pull/307

## Test plan
Tested locally by executing a SQL builtin with the demo db attached and running `airplane dev --editor` without any env flags.


pkg/resource/resources.go
```diff
@@ -101,6 +106,35 @@ func MergeRemoteResources(ctx context.Context, state *state.State) (map[string]e
...
+	// Also add demo DB for convenience, which is required by the getting started guides.
+	if _, ok := mergedResources[demoDBSlug]; !ok {
+		// Unlike the Slack resource above, the demo db does not have a fixed resource id, and so we get the resource
+		// by slug here.
+		demoDB, err := state.CliConfig.Client.GetResource(ctx, api.GetResourceRequest{
+			Slug: demoDBSlug,
+		})
+		if err != nil {
+			// If demo_db resource isn't found, don't error.
+			if _, ok := err.(*libapi.ResourceMissingError); ok {
```
colinking, line 118: Use `errors.As` so that this defensively handles error wrapping.

pkg/resource/resources.go
```diff
@@ -101,6 +106,35 @@ func MergeRemoteResources(ctx context.Context, state *state.State) (map[string]e
...
+				return nil, nil
```
colinking, line 119: This should return `mergedResources`, right? Maybe extract this into a function if that's easier.

colinking (approved):
LGTM!

pkg/resource/resources.go
calufornia: Good catch, fixed!

PR #689, opened 2022-09-27
Check node version in cli

geodn:
## Description
Ensure that when `airplane views dev` is run, node is at least v14.18.0.
## Test plan
Run the following with `requiredNodeVersion` set to different values.
`go install ./cmd/airplane`
`airplane views dev`


zhan-airplane (approved):
Can you give some context for what breaks with older node versions?

geodn:
With node 12:
Screen Shot 2022-09-27 at 10 48 48 AM

With node 14.17:
Screen Shot 2022-09-27 at 10 53 01 AM

PR #691, opened 2022-09-27
Install @types/node by default on task init

ameyakhare:
## Description
Adds @types/node to dev dependencies so we don't get ts errors when initializing task.

## Test plan
Exists in tsconfig now.


colinking (approved):

leeweisberger (changes requested):
I don't think we want this. You use lib "DOM" when you run the code in a browser environment and want access to "document" and "window" etc.

leeweisberger (approved):

PR #695, opened 2022-09-27
Check if slack resource exists before merging with local resources

calufornia:
## Description
Currently, we attempt to automatically add a team's Slack resource for convenience, so they don't have to specify a `--env` fallback. If a team hasn't configured Slack, however, this leads to a confusing error message to the user about the request to `/v0/resources/get` failing. Similar to demo DB, we should only add slack to the editor's merged (local + remote) resources map if it exists remotely.

The user also receives a better error message when they call a slack builtin but haven't configured Slack: `Your team has not configured Slack. Please visit https://app.airplane.so:5000/settings/team to authorize Slack to perform actions in your workspace.`

## Test plan
Tested in conjunction with API changes to return a 404 if a team has not configured Slack.


pkg/server/apiext/apiext.go
```diff
@@ -121,6 +121,9 @@ func ExecuteTaskHandler(ctx context.Context, state *state.State, r *http.Request
 			}
 
 			if !foundResource {
+				if resourceID == resource.SlackID {
+					return dev.LocalRun{}, errors.Errorf("Your team has not configured Slack. Please visit %s/settings/team to authorize Slack to perform actions in your workspace.", state.CliConfig.Client.AppURL().String())
```
colinking, line 125: Minor: but do we capitalize errors or not? e.g. see error on line 127

Also: should we link to the docs instead? https://docs.airplane.dev/platform/slack-integration#connect-to-slack

colinking (approved):

pkg/server/apiext/apiext.go
calufornia: Decided to capitalize the errors here since this is "user-facing" - for more internal errors we should keep this lowercase. We should go through all the errors the user could see and capitalize them if they aren't already.

colinking: FWIW, our API doesn't capitalize errors until serializing them. Maybe we do the same in the CLI? e.g. always lowercase, then when printing the error, capitalize it?

PR #697, opened 2022-09-28
Add local client for legacy airplane dev

calufornia:
## Description

We made a change to the `--editor` flag to pass in a local client to the server state instead of manually overriding the host of the root client to easily control where we use a local client vs a remote client. We port that change over to the `airplane dev` command without the `--editor` flag as well.

Also sets the `EnvID` and `EnvSlug` for `airplane dev` to always be "local"; the `--editor` flag must be provided to hit remote resources and tasks.

## Test plan
Tested execution of a task with `aplocal dev` that also included task composition.


colinking (approved):
thanks for the quick fix!

PR #700, opened 2022-09-30
Show warning for tasks with no resources

kellymtrinh:
## Description
Instead of returning an error if a task has a resource attached that doesn't exist, return the metadata anyway.

Adds a new endpoint `dev/tasks/errors\?slug={slug} ` so the UI can display warnings from local dev if needed. Right now the only message we show is if an attached resource doesn't exist:

```
curl localhost:4000/dev/tasks/errors\?slug=builtins_example | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   182  100   182    0     0  22405      0 --:--:-- --:--:-- --:--:--  177k
{
  "errors": [],
  "warnings": [
    {
      "level": "warning",
      "name": "builtins_example",
      "kind": "task",
      "reason": "Attached resource: [my_rest1] not found in dev config file or remotely."
    }
  ],
  "info": null
}
```

Screen Shot 2022-09-29 at 4 55 54 PM

## Test plan
- [x] Test that I can still view the task pages for tasks that have params and without params.
- [x] Test  that the warning endpoint returns the correct values.


pkg/server/apiext/apiext.go
```diff
@@ -337,11 +370,36 @@ func GetTaskInfoHandler(ctx context.Context, state *state.State, r *http.Request
 	if !ok {
 		return libapi.UpdateTaskRequest{}, errors.Errorf("Task with slug %s not found", taskSlug)
 	}
-	req, err := taskConfig.Def.GetUpdateTaskRequest(ctx, state.LocalClient)
+	req := libapi.UpdateTaskRequest{
```
kellymtrinh, line 373: Basically `taskConfig.Def.GetUpdateTaskRequest` did validation on whether the resource exists- so if we don't want to do validation we can populate turning a task def -> task request with metadata filled out.

kellymtrinh: I'm on the fence on whether we prefer this or doing a one line change:
```
logger.Error("Encountered error while getting task info: %v", err)
return libapi.UpdateTaskRequest{}, errors.Errorf("error getting task %s", taskSlug)
```

To be an `errors.Wrap(err, "error getting task")` which give a better error message like this:
Screen Shot 2022-09-30 at 2 59 18 PM

pkg/server/apiext/apiext.go
```diff
@@ -33,6 +33,7 @@ func AttachExternalAPIRoutes(r *mux.Router, state *state.State) {
 	r.Handle("/tasks/execute", handlers.HandlerWithBody(state, ExecuteTaskHandler)).Methods("POST", "OPTIONS")
 	r.Handle("/tasks/getMetadata", handlers.Handler(state, GetTaskMetadataHandler)).Methods("GET", "OPTIONS")
 	r.Handle("/tasks/get", handlers.Handler(state, GetTaskInfoHandler)).Methods("GET", "OPTIONS")
+	r.Handle("/tasks/warnings", handlers.Handler(state, GetTaskWarningsHandler)).Methods("GET", "OPTIONS")
```
calufornia, line 36: Could we make this a `/dev/` endpoint? Since this is local-dev specific, and not part of our external API.

pkg/server/apiext/apiext.go
calufornia: I like the way you have it now - to add on, we could even have the ability to disable the "execute" button until the issues have been resolved. But I think being able to see the other task metadata is useful, even if some parts are missing.

pkg/server/server.go
```diff
@@ -184,9 +185,13 @@ func (s *Server) RegisterTasksAndViews(ctx context.Context, taskConfigs []discov
 				TaskName:      cfg.Def.GetName(),
 				ResourceSlugs: missingResources,
 			})
+			taskWarnings[cfg.Def.GetSlug()] = []string{
+				fmt.Sprintf("Attached resource: %v not found in dev config file or remotely.", missingResources),
```
calufornia, line 189: Instead of setting the warning to be a string, could we make `taskWarnings` a `map[string]interface{}` and have different warning structs (similar to what we do for registration warnings). That way, the frontend can handle structuring the exact error message (font-mono for slugs, etc.)

pkg/server/apiext/apiext.go
```diff
@@ -327,6 +342,24 @@ func GetOutputsHandler(ctx context.Context, state *state.State, r *http.Request)
 	}, nil
 }
 
+func ParamWithComponent(p libapi.Parameter) (libapi.Parameter, error) {
```
calufornia, line 345: This seems like duplicate code from lib - could we create a helper method in lib and then use that here?

kellymtrinh: The lib code converts `ParameterDefinition_0_3` object (which has "shorttext", "longtext", etc. types) to a `api.Parameter` object (type "string"). `ParameterDefinition_0_3` isn't accessible in most places since we're using the `DefinitionInterface` instead which uses `api.Parameter`. This needs to convert the `api.Parameter` -> `api.Parameter` - that doesn't feel like it belongs in lib since this conversion doesn't happen anywhere else (and isn't reusable with the lib part).

I think this might require some revisiting later since it feels specific to the CLI - in airport/the db itself we store all the parameter definitions in the format web expects ... - so maybe this should happen in the task discoverer (to convert the parameter definitions to the correct "types") so this conversion doesn't even need to happen in the CLI?  I don't have full context on why we have two representations, but can file at ticket to revisit

calufornia: I guess I'm suggesting to move this function into lib and use it [here](https://github.com/airplanedev/lib/blob/01542ef2e9906679933145d3c7f1b783806ec36c/pkg/deploy/taskdir/definitions/def_0_3.go#L1068-L1081) (which uses `api.Parameter`) and in this PR. You could call it something like `populateParameter(p api.Parameter, paramType string)` - in both places, we're just filling in / modifying a parameter struct.

Also at a higher level, why do we need to do this conversion? As you mentioned, `api.Parameter` structs already have type fields with `"string"`, etc.

pkg/server/apidev/apidev.go
```diff
@@ -31,6 +32,8 @@ func AttachDevRoutes(r *mux.Router, s *state.State) {
 	r.Handle("/list", handlers.Handler(s, ListEntrypointsHandler)).Methods("GET", "OPTIONS")
 	r.Handle("/startView/{view_slug}", handlers.Handler(s, StartViewHandler)).Methods("POST", "OPTIONS")
 	r.Handle("/logs/{run_id}", handlers.HandlerSSE(s, LogsHandler)).Methods("GET", "OPTIONS")
+	r.Handle("/tasks/errors", handlers.Handler(s, GetTaskErrorsHandler)).Methods("GET", "OPTIONS")
+
```
calufornia, line 36:
```suggestion
```

pkg/server/apidev/apidev.go
```diff
@@ -231,3 +234,31 @@ func LogsHandler(ctx context.Context, state *state.State, r *http.Request, flush
 		}
 	}
 }
+
+type GetTaskErrorResponse struct {
+	Errors   []dev_errors.AppError `json:"errors"`
+	Warnings []dev_errors.AppError `json:"warnings"`
+	Info     []dev_errors.AppError `json:"info"`
```
calufornia, line 241: This doesn't look like it's used - let's also set it in `GetTaskErrorsHandler`

pkg/server/apiext/apiext.go
kellymtrinh: When the task definition's parameters are read (out of yaml, for example), they have values like `type: shorttext` and empty component. But what they should be stored in the db and returned to web as is 
`type: string, component: componentTextarea`

I think this might be specific to CLI - in airport/db we store it in the converted form:
```
parameters| {"parameters": [{"name": "test short text", "slug": "test_short_text", "type": "string", "constraints": {"optional": false}}, {"name": "t
est sql", "slug": "test_sql", "type": "string", "component": "editor-sql", "constraints": {"optional": false}}]}
```
in the task_revisions table

PR #701, opened 2022-10-03
Cache vite processes when developing views within editor

ameyakhare:
Resolves TSK-690

## Description
Caches the last 5 views that were loaded with the editor using LRU eviction policy.

## Test plan
Loaded a view, switched away, and loaded the same view again without killing/restarting vite.


pkg/server/apidev/apidev.go
```diff
@@ -128,23 +129,28 @@ type StartViewResponse struct {
 }
 
 // StartViewHandler handles requests to the /dev/tasks/<task_slug> endpoint.
-func StartViewHandler(ctx context.Context, state *state.State, r *http.Request) (StartViewResponse, error) {
+func StartViewHandler(ctx context.Context, svs *state.State, r *http.Request) (StartViewResponse, error) {
```
calufornia, line 132: Can you rename it to `s` - TBH we should just call it `s` everywhere

pkg/server/server.go
```diff
@@ -108,18 +109,36 @@ func newServer(router *mux.Router, state *state.State) *Server {
 
 // Start starts and returns a new instance of the Airplane API server.
 func Start(opts Options) (*Server, error) {
+	onEvict := func(key interface{}, value interface{}) {
+		viteContext, ok := value.(state.ViteContext)
+		if !ok {
+			logger.Error(fmt.Sprintf("could not evict existing vite process in lru"))
```
calufornia, line 115: No need for `fmt.Sprintf` - also, the error message should probably be something like `"expected vite context from context cache"` like the other type assertion

pkg/server/server.go
```diff
@@ -108,18 +109,36 @@ func newServer(router *mux.Router, state *state.State) *Server {
...
+
+		if err := viteContext.Process.Kill(); err != nil {
+			logger.Error(fmt.Sprintf("could not shutdown existing vite process: %v", err))
+		}
+	}
+
+	viteContextCache, err := lrucache.NewWithEvict(5, onEvict)
+	if err != nil {
+		logger.Error(fmt.Sprintf("failed to create vite context cache: %v", err))
+		os.Exit(1)
```
calufornia, line 126: We can just return the error here

calufornia (approved):

PR #705, opened 2022-10-04
Copy endpoints to internal

calufornia:
## Description

To maintain parity with the rest of lib, we move `/displays/list`, `/runs/list`, and `/tasks/get` to be under the `/i/` subpath, since these endpoints are used exclusively by web. We'll remove the corresponding handlers in apiext in a future PR.

## Test plan
Tested in conjunction with web changes.


kellymtrinh (approved):
👍 these don't look used in the SDK either

PR #707, opened 2022-10-05
List remote resources alongside local resources

calufornia:
## Description

Return remote resources in the `/i/resources/list` endpoint of the local dev server if a fallback environment is specified. This will be used for listing remote resources in the editor sidebar, and is also required by the run details page for builtins executed against a remote resource.

## Test plan
Tested executing a task with a builtin and verified that run details page was populated with remote resource configuration.

### DIFF
--- a/pkg/server/apiint/apiint.go
+++ b/pkg/server/apiint/apiint.go
@@ -182,7 +182,7 @@ func ListResourcesHandler(ctx context.Context, state *state.State, r *http.Reque
 		})
 	}
 
-	if state.EnvID == env.LocalEnvID {
+	if state.EnvID != env.LocalEnvID {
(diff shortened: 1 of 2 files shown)


pkg/server/apiint/apiint.go
```diff
@@ -154,22 +155,52 @@ func GetResourceHandler(ctx context.Context, state *state.State, r *http.Request
 	return GetResourceResponse{}, errors.Errorf("Resource with slug %s is not in dev config file", slug)
 }
 
+type APIResourceWithEnv struct {
+	libapi.Resource
+	Remote bool `json:"remote"`
+}
```
kellymtrinh, line 161: Would it be useful to put the env here as well? Unless we don't have plans to list resources from multiple remote envs

kellymtrinh (approved):

PR #714, opened 2022-10-07
Return custom error for normalize resource errors

calufornia:
## Description
In https://github.com/airplanedev/lib/pull/319, validation was added to SQL + REST tasks to make sure that the resource attached to those tasks was valid. However, the `ResourceMissingError` that is returned doesn't have an `AppURL` field set, and so the error message that is bubbled up to the user is malformed. This returns a custom error in the case we get a `ResourceMissingError`.

cc @francescaleung 

## Test plan
Tested locally by calling `aplocal dev --editor` in a directory with a SQL tasks with an unknown resource.


francescaleung:
sorry, did you want to chat about this before merging?

calufornia:
> sorry, did you want to chat about this before merging?

Nope, was just flagging so you were aware

PR #715, opened 2022-10-07
Make taskConfigs and viewConfigs concurrency safe

kellymtrinh:
## Description
This is a pre-PR to support hot reloading: moves the task validation logic out to a reusable function and makes task and viewconfigs concurrency safe. We're going to start refreshing taskConfigs and viewConfigs when the user makes file changes, which used to only be discovered once at startup time.

This involves adding a write lock to the maps which represent them- there is a `sync.Map` implementation but it's not generic (uses `interface{}`) - so I decided to make a small generic `store` struct that can be used to store generic types since we end up reusing it 3x. 

The requirements for hot reloading are going to be: getting, replacing individual task/view configs, and optionally replacing the entire map (ie. if we run "discover" again on the entire working directory)

## Test plan
- [x] task execution is not affected
- [x] views are a no-op
- [x] task warnings for unattached resources still show up


pkg/server/state/state.go
```diff
@@ -30,18 +30,17 @@ type State struct {
...
-	TaskErrors map[string][]dev_errors.AppError
-
+	TaskErrors Store[[]dev_errors.AppError]
 	// Mapping from view slug to view config
-	ViewConfigs  map[string]discover.ViewConfig
+	ViewConfigs Store[discover.ViewConfig]
+
 	DevConfig    *conf.DevConfig
 	ViteContexts *lrucache.Cache
-	ViteMutex    sync.Mutex
```
kellymtrinh, line 44: This stopped being needed after `ViteContexts` was switched from map to `lrucache. Cache`, which is concurrency safe

pkg/server/state/store.go
```diff
@@ -0,0 +1,71 @@
+package state
+
+import (
+	"sync"
+
+	"github.com/pkg/errors"
+)
+
+// Generic store that is concurrency safe
+type Store[V any] struct {
```
calufornia, line 10: You can represent the store using two generics:
```
type Store[K comparable, V any] struct {
        items map[K]V
	mu    sync.Mutex
}
```

pkg/server/state/store.go
```diff
@@ -0,0 +1,71 @@
...
+	items map[string]V
+	mu    sync.Mutex
+}
+
+type KV[V any] struct {
```
calufornia, line 15: If we change the store representation to have keys be generic, then we don't need this struct.

pkg/server/state/store.go
```diff
@@ -0,0 +1,71 @@
...
+
+func NewStore[V any](items map[string]V) Store[V] {
+	if items != nil {
+		return Store[V]{items: items}
+	}
+	return Store[V]{items: map[string]V{}}
+}
+
+// Returns a list of the keys and values in Items, which users can iterate over
+func (s *Store[V]) Items() []KV[V] {
```
calufornia, line 28: Same as above, but can just directly return `items` here instead

pkg/server/state/store.go
```diff
@@ -0,0 +1,71 @@
...
+	return res, nil
+}
+
+func (s *Store[V]) Delete(key string) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	delete(s.items, key)
+}
+
+func (s *Store[V]) ReplaceItems(items map[string]V) {
```
calufornia, line 67: What's the advantage of mutating the existing store (and requiring lock/unlocking) vs just using `NewStore`?

pkg/server/state/store.go
kellymtrinh: Nice, updated this

pkg/server/state/store.go
kellymtrinh: You still need this because otherwise you'd be exposing the pointer to `items`? Which I don't want clients to have direct access to mutate at all, unless it's through the methods provided

kellymtrinh: Changed to return a copy of items, which :/ does not work with deep values that are pointers?

pkg/server/state/store.go
kellymtrinh: It seemed safer: https://stackoverflow.com/questions/60820731/is-assigning-value-to-a-struct-atomic-in-go but it's probably unlikely we'll run into this

pkg/server/state/store.go
```diff
@@ -0,0 +1,70 @@
...
+func (s *Store[K, V]) Add(key K, val V) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.items[key] = val
+}
+
+// Returns a copy of the items map for reading
+func (s *Store[K, V]) Items() map[K]V {
+	m := make(map[K]V, len(s.items))
+	for k, v := range s.items {
```
calufornia, line 31: This isn't thread-safe, right? As in, while iterating over s.items, an item could get added - I think we also want a lock/unlock at the beginning of this method.

calufornia: You can test this method in `store_test.go` - call `Items()` and in another goroutine call `Add()` - I _believe_ that go should complain about a data race.

calufornia (approved):

pkg/server/state/store.go
calufornia: Yep, returning a copy of the map is good enough - I just wanted to make sure `Store` was as "map-like" as possible, i.e. iterating of key-value pairs instead of idx, value

PR #716, opened 2022-10-10
Use api client interface in dev state and always include demo db when listing resources

calufornia:
## Description

Changes the `/i/resources/list` endpoint on the local dev server to always include the demo db resource, even if a fallback environment hasn't been set - previously, the list endpoint would only return local resources in this case. The demo db is a special resource in that the user doesn't have to supply a fallback env to use it - however, the builtin run details page of the web app utilizes `/i/resources/list`, and so we need to make sure to return any default remote resources here too.

Also stops passing through the entire CLI config to the dev state in favor of passing in a `RemoteClient` interface, which makes testing easier (the ListResources test broke after the changes above) to do since the introduction of the fallback environment there's a few places in local dev where we make "remote" calls to the actual airplane API instead of our local dev server. To accomplish this, we add some missing methods that weren't included on the `api.APIClient` interface, and also move the `ParseTokenForAnalytics()` method from the CLI config to the api client, since the only field that this method uses is the token from the API client.

## Test plan
Tested `aplocal dev --editor` without a fallback environment, ran a builtin against the demo db, and verified that the run details page was populated correctly:
image


pkg/api/client.go
```diff
@@ -109,6 +114,9 @@ type APIClient interface {
 	CreateDemoDB(ctx context.Context, name string) (string, error)
 
 	GetEnv(ctx context.Context, envSlug string) (Env, error)
+
+	// All methods below this point represent CLI-specific API operations, and not requests to api.airplane.dev.
+	ParseTokenForAnalytics() AnalyticsToken
```
colinking, line 119: Does this need to be a method on the API client? I expect we merge the `APIClient` interface with the one in `lib` eventually.

Could it be a utility function that accepts a token?

```
func ParseTokenForAnalytics(token string) AnalyticsToken
```

pkg/api/client_mock.go
```diff
@@ -193,5 +205,15 @@ func (mc *MockClient) GetEnv(ctx context.Context, envSlug string) (Env, error) {
...
+		if r.Slug == req.Slug {
+			return libapi.GetResourceResponse{Resource: r}, nil
+		}
+	}
+
+	return libapi.GetResourceResponse{}, errors.Errorf("resource with slug %s does not exist", req.Slug)
+}
+
+func (mc *MockClient) ParseTokenForAnalytics() AnalyticsToken {
+	return AnalyticsToken{}
```
colinking, line 218: panic not implemented?

pkg/resource/resources.go
```diff
@@ -45,7 +46,7 @@ func GenerateAliasToResourceMap(
 			if state.HasFallbackEnv() {
 				envSlug = state.Env.Slug
 			}
-			remoteResourceWithCredentials, err := state.CliConfig.Client.GetResource(ctx, api.GetResourceRequest{
+			remoteResourceWithCredentials, err := state.RemoteClient.GetResource(ctx, api.GetResourceRequest{
```
colinking, line 49: nice, much clearer

pkg/resource/resources.go
```diff
@@ -125,12 +125,33 @@ func mergeDefaultRemoteResource(
 	}
 }
 
+// ListRemoteResources returns any remote resources that the user can develop against. If not fallback environment is
```
colinking, line 128:
```suggestion
// ListRemoteResources returns any remote resources that the user can develop against. If no fallback environment is
```

pkg/server/apidev/apidev.go
```diff
@@ -166,12 +166,14 @@ func StartViewHandler(ctx context.Context, s *state.State, r *http.Request) (Sta
...
-	viewsClient := s.CliConfig.Client
+	viewsClient := s.RemoteClient
 	if !s.HasFallbackEnv() {
 		viewsClient = s.LocalClient
 	}
 	cmd, viteServer, err := views.Dev(ctx, vd, views.ViteOpts{
-		Client:  viewsClient,
+		// Views dev expects an api.Client implementation since it passes the client's token, API key, etc. as
+		// environment variables to the vite process.
+		Client:  viewsClient.(*api.Client),
```
colinking, line 176: Oof, we really need to unify the lib and CLI definitions for API clients at some point soon

colinking (approved):

PR #719, opened 2022-10-12
Support hot reloading in studio

kellymtrinh:
## Description
**Previous behavior:**
Needed to stop and restart the server each time your task/view code changes.

**New behavior:**
Behind a new `--watch` flag.
This PR starts listening to file changes and automatically refreshes task/view/resource configs when they're changed.

The logic:

- For updating and creating new views/tasks, run `discover` on that task/view only. We only care about files that could contain a task or view definition, not the task code or view code itself.
-  If a file is deleted or renamed, rerun discover on the directory 
- Everything is debounced by filepath. If you're still editing a file, it will wait. Once you finish editing the file (aka the debounce duration has passed),  it will kick off reloading that file. The reason to debounce by filepath

Other supporting changes:

- Adds a lock to the devConfig struct and unexports fields in favor of using setters like `SetConfig`
 

- [x]  update to latest lib when https://github.com/airplanedev/lib/pull/340 is merged and released

## Test plan

- [x] `airplane dev task` (not the studio) is unaffected, print statements look the same
- [x] Tested updating, creating, and deleting task files
  - updating or creating a new task through `airplane init` would correctly pick up and show the new one in the UI
  - introducing a task error would display the error. 
  - Fixing the task error (ie. unattached resource) would make the error go away.
- [x] updating, creating, and deleting views
- [x] updating the resource config file directly 
- [x] starting/deving a view does not trigger reloading

- [x] resource and config setting from CLI is unaffected
- [x] resource setting from the studio UI is unaffected 

Screen Shot 2022-10-13 at 3 03 03 PM


kellymtrinh:
I'm going to ask for more dogfooding for some of this first!

cmd/airplane/tasks/dev/server.go
```diff
@@ -186,10 +168,30 @@ func runLocalDevServer(ctx context.Context, cfg taskDevConfig) error {
 		logger.Log("- Any resources not declared in your dev config will be loaded from your %s environment.", logger.Bold(devEnv.Name))
 	}
 
+	// FILEWATCHER
```
calufornia, line 171: More descriptive comment?

pkg/conf/dev.go
```diff
@@ -90,14 +102,62 @@ func (d *DevConfig) RemoveResource(slug string) error {
...
-	if err := WriteDevConfig(d); err != nil {
+	if err := writeDevConfig(d); err != nil {
 		return errors.Wrap(err, "writing dev config")
 	}
 
 	return nil
 }
 
-func ReadDevConfig(path string) (*DevConfig, error) {
+// RemoveResource deletes the config from the dev config file, if it exists.
```
calufornia, line 112:
```suggestion
// RemoveConfig deletes the config from the dev config file, if it exists.
```

calufornia: Maybe for clarity, `RemoveConfigVariable`?

pkg/conf/dev.go
```diff
@@ -90,14 +102,62 @@ func (d *DevConfig) RemoveResource(slug string) error {
...
+	if _, ok := d.ConfigVars[key]; !ok {
+		return errors.Errorf("Config variable `%s` not found in dev config file", key)
+	}
+
+	delete(d.ConfigVars, key)
+	if err := writeDevConfig(d); err != nil {
+		return err
+	}
+
+	logger.Log("Deleted config variable `%s` from dev config file.", key)
```
calufornia, line 125: Could we use `%q`, I should have used that originally

pkg/conf/dev.go
```diff
@@ -90,14 +102,62 @@ func (d *DevConfig) RemoveResource(slug string) error {
...
+	if d.ConfigVars == nil {
+		d.ConfigVars = map[string]string{}
+	}
+	d.ConfigVars[key] = value
+	err := writeDevConfig(d)
+	if err != nil {
+		return err
+	}
+
+	logger.Log("Successfully wrote config variable `%s` to dev config file.", key)
```
calufornia, line 142: Same here

pkg/server/filewatcher/filewatcher.go
```diff
@@ -0,0 +1,96 @@
...
+	return false
+}
+
+func (f *AppWatcher) Watch(wd string) error {
+	logger.Log(" ")
+	logger.Log(logger.Green("Watching for changes in: %s", wd))
+	logger.Log(logger.Green("Changes to tasks and views will be applied automatically."))
+
+	wd = wd + "/..."
+	if err := notify.Watch(wd, f.changes, notify.All); err != nil {
```
calufornia, line 73:
```suggestion
	if err := notify.Watch(wd+"/...", f.changes, notify.All); err != nil {
```

pkg/server/filewatcher/filewatcher.go
```diff
@@ -0,0 +1,96 @@
...
+	return &AppWatcher{
+		changes:  make(chan notify.EventInfo, 20),
+		callback: opts.Callback,
+		isValid:  opts.IsValid,
+	}
+}
+
+func IsValidDefinitionFile(path string) bool {
+	// ignore directories
+	if discover.IgnoredDirectories[filepath.Base(path)] {
```
calufornia, line 48: So path can be both a file or directory?

pkg/server/filewatcher/filewatcher.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	// ignore files inside ignored directories
+	for dir := range discover.IgnoredDirectories {
+		fileDir := filepath.Dir(path)
+		if strings.Contains(fileDir, dir) {
+			return false
+		}
+	}
+	// only allow supported file extensions that can contain task/view definitions
+	for _, ext := range discover.DefinitionFileExtensions {
```
calufornia, line 59: We need to check that path is a file, and not a directory right?

pkg/server/filewatcher/filewatcher.go
```diff
@@ -0,0 +1,96 @@
...
+
+func IsValidDefinitionFile(path string) bool {
+	// ignore directories
+	if discover.IgnoredDirectories[filepath.Base(path)] {
+		return false
+	}
+	// ignore files inside ignored directories
+	for dir := range discover.IgnoredDirectories {
+		fileDir := filepath.Dir(path)
+		if strings.Contains(fileDir, dir) {
```
calufornia, line 54: I feel like this might lead to some false negatives, but I'm fine if the alternative (splitting the path into directories, e.g. `/a/b/c` -> `[a, b, c]`, and looking for the ignored directory somewhere inside that, is too computationally expensive)

pkg/server/server.go
```diff
@@ -218,30 +224,157 @@ func ValidateTasks(ctx context.Context, resources map[string]env.ResourceWithEnv
...
+				}
+			}
+			return []discover.TaskConfig{}, []discover.ViewConfig{}, errors.Errorf(message)
+		}
+		return []discover.TaskConfig{}, []discover.ViewConfig{}, errors.Wrap(err, "discovering task configs")
+	}
+	return taskConfigs, viewConfigs, err
+}
+
+// shouldReloadDirectory returns whether the  entire directory should be refreshed
```
calufornia, line 257:
```suggestion
// shouldReloadDirectory returns whether the entire directory should be refreshed
```

pkg/server/server.go
```diff
@@ -218,30 +224,157 @@ func ValidateTasks(ctx context.Context, resources map[string]env.ResourceWithEnv
...
+	shouldRefreshDir := shouldReloadDirectory(e)
+	if shouldRefreshDir {
+		path = wd
+	}
+	var reload func()
+
+	if path == s.state.DevConfig.Path {
+		reload = func() {
+			if err := s.state.DevConfig.LoadConfigFile(); err != nil {
+				logger.Log(err.Error())
```
calufornia, line 280: `logger.Error` here and elsewhere

calufornia (approved):
This is great - I also recall a discussion around showing "last deployed at" or something similar in the studio UI - will that be a follow up to this PR?

cmd/airplane/tasks/dev/server.go
```diff
@@ -186,10 +168,30 @@ func runLocalDevServer(ctx context.Context, cfg taskDevConfig) error {
...
+	} else {
+		logger.Log("")
+		logger.Log("Changes require restarting the studio to take effect.")
+	}
+
 	logger.Log("")
 	studioURL := fmt.Sprintf("%s/studio?host=http://localhost:%d", appURL, cfg.port)
 	logger.Log("Started studio session at %s (^C to quit)", logger.Blue(studioURL))
 	logger.Log("Press ENTER to open the studio in the browser.")
+	logger.Log("")
```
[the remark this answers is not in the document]
kellymtrinh: It used to be the last line but now it's going to start logging a bunch of "reloaded x" - I thought it looked nicer to have all the startup stuff finish, do a newline and then print everything else.

pkg/server/filewatcher/filewatcher.go
kellymtrinh: A path can be either! Since "file changes" could be CRUD directory or CRUD a file. This is quitting early if we know the path is the ignored directory itself. After that, check if it's inside an ignored directory. Then it only allows supported file types.

kellymtrinh:
> This is great - I also recall a discussion around showing "last deployed at" or something similar in the studio UI - will that be a follow up to this PR?

Yeah I'll do this in a follow up since it will require a web change too!

PR #722, opened 2022-10-13
Add support for JSTs in local dev

calufornia:
## Description

Makes the following changes:

- Uses the `/v0/templates/evaluate` endpoint to interpolate kind options - we'll remove the logic in lib that interpolates expressions after landing this. 
- Interpolates environment variables for code tasks, which were previously not interpolated.
- Adds `resources` and `env` to the lookup map.
- Sets an output error even if the command hasn't started yet, to better inform the user.

Depends on https://github.com/airplanedev/lib/pull/344

cc @francescaleung 

## Test plan
Tested with a SQL task that had JSTs in the query, and also tested with a shell task that used JSTs in environment variables.

image


pkg/server/apiext/apiext.go
```diff
@@ -180,22 +181,31 @@ func ExecuteTaskHandler(ctx context.Context, state *state.State, r *http.Request
...
 		state.Runs.Add(req.Slug, runID, run)
 
-		// use a new context while executing
-		// so the handler context doesn't cancel task execution
+		// Use a new context while executing so the handler context doesn't cancel task execution
 		go func() {
 			outputs, err := state.Executor.Execute(context.Background(), runConfig)
 			completedAt := time.Now()
-			run, err = state.Runs.Update(runID, func(run *dev.LocalRun) error {
+			var updateErr error
```
kellymtrinh, line 188: This is a bit confusing to me for which err and run variable is being used -  there's the run that's the func argument in update and also the run that's assigned outside of that. 
Could we handle the 
`outputs, err := state.Executor.Execute(context.Background(), runConfig)` to say 
`if err != nil { 
status = failed
outputs = whatever
}`

and then update that inside `state.Runs.Update(runID, func(run *dev.LocalRun) error { // ... }`

calufornia: Oh yeah good point, I'm not sure why it was being handled in Update to begin with

pkg/api/client.go
```diff
@@ -543,6 +550,7 @@ func (c Client) do(ctx context.Context, method, path string, payload, reply inte
 
 	if payload != nil {
 		buf, err := json.Marshal(payload)
+		fmt.Println(string(buf))
```
kellymtrinh, line 553: Delete?

PR #725, opened 2022-10-14
Generate resource ID if it doesn't exist in the config file

kellymtrinh:
## Description

This hides the concept of "resource IDs" in the dev config yaml, because users shouldn't need to know about or interact with IDs.

- Stop writing `id` field into a resource file:
```
configVars:
    API_KEY: abc-my-apikey
    hello: world3
resources:
    - auth: null
      baseURL: https://jsonplaceholder.typicode.com
      headers: {}
      kind: rest
      name: Rest1
      secretHeaders: []
      slug: my_rest
```

- The ID is generated when reading from the dev config file, for use in the dev server
```
{
  "resources": [
    {
      "id": "res-test",
      "slug": "test",
... etc.
```

## Test plan
- [x] Creating, updating, and deleting a resource via studio UI still works. The dev config yaml doesn't have the ID field
- [x] using a resource in a task still works. The resources endpoint returns the generated ID correctly. 
- [x] Setting a resource from the command line works, and I'm able to use it


pkg/conf/dev.go
```diff
@@ -211,11 +219,14 @@ func readDevConfig(path string) (*DevConfig, error) {
 	return cfg, nil
 }
 
+func GenerateLocalResourceID(slug string) string {
```
calufornia, line 222: This seems like it more belongs in the `resources` package - can we put it there?

calufornia (approved):

PR #726, opened 2022-10-14
Support developing multiple views in parallel

calufornia:
## Description
Previously, if two views shared the same root (i.e. used the same package.json), the second view that was booted up would overwrite the contents of the `.airplane-view/` directory (namely, `main.tsx` and `index.html`). This would cause the first view to display the contents of the second view. To support developing multiple views at the same time, we instead create the vite assets in a slug-specific subdirectory of `.airplane-view/`, while keeping common files (e.g. `node_modules`, `vite.config.ts`, etc.) still in the top level of `.airplane-view/`. The new directory structure looks something like:
```
.airplane-view/
|_ admin_panel
    |_ index.html
    |_ main.tsx
|_ admin_panel_2
    |_ index.html
    |_ main.tsx
|_ package.json
|_ vite.config.ts
|_ yarn.lock
```

Also renames some of the variables in views dev to be more clear - `.airplane-view/` is not a temporary directory.

## Test plan
Tested loading multiple views into the Studio and switching between them. Also tested multiple `airplane views dev` commands running simultaneously.


pkg/views/dev.go
```diff
@@ -40,25 +40,35 @@ func Dev(ctx context.Context, v viewdir.ViewDirectory, viteOpts ViteOpts) (*exec
...
+	airplaneViewDir := filepath.Join(root, ".airplane-view")
+	if _, err := os.Stat(airplaneViewDir); os.IsNotExist(err) {
+		if err := os.Mkdir(airplaneViewDir, 0755); err != nil {
+			return nil, "", errors.Wrap(err, "creating .airplane-view dir")
 		}
-		l.Debug("created temporary dir %s", tmpdir)
+		l.Debug("created .airplane-view dir %s", airplaneViewDir)
 	} else {
-		l.Debug("temporary dir %s exists", tmpdir)
+		l.Debug(".airplane-view dir %s exists", airplaneViewDir)
```
leeweisberger, line 50: If there is already a legacy (before this change) `.airplane.view` dir, is it still going to contain a hanging index/main?

Same question for if you delete one of your views - is the nested slug folder for that view still going to exist on subsequent airplane devs?

Can you check this out and then clean up hanging files? Feel free to push back if this seems difficult/not super important

leeweisberger (changes requested):
1 ask

pkg/views/dev.go
calufornia: Good point, my proposal is now to keep view-specific directories for the duration of the vite command - when `airplane views dev` or `airplane dev --studio` finishes, the view-specific directories are removed. Also added some logic to remove the legacy `main.tsx` and `index.html` files, if the user had one.

pkg/views/dev.go
```diff
@@ -222,12 +258,16 @@ type ViteOpts struct {
...
 }
 
-func runVite(ctx context.Context, opts ViteOpts, tmpdir string) (*exec.Cmd, string, error) {
-	cmd := exec.Command("node_modules/.bin/vite", "dev")
+func runVite(ctx context.Context, opts ViteOpts, airplaneViewDir string, viewSlug string) (*exec.Cmd, string, error) {
+	// By default, Vite attempts to locate a config file `vite.config.ts` inside the project root, which it determines
+	// based on the location of the index.html file. Because this is part of the view-specific subdirectory, Vite will
+	// not find the config file there, and we instead need to tell it to use the config file one level higher, at
+	// .airplane-view/vite.config.ts.
+	cmd := exec.Command("node_modules/.bin/vite", "dev", "--config", "vite.config.ts", viewSlug)
```
leeweisberger, line 266: Can you remove the todo on the next line? I don't think it applies anymore?

leeweisberger (approved):

PR #733, opened 2022-10-17
Return remote tasks in /dev/list when fallback env is specified

ameyakhare:
TSK-841

## Description
Returns remote tasks in list response so that these can be rendered under the entity list in the studio.

## Test plan
Called the endpoint and observed it worked.


calufornia (approved):
You'll want to fix up / add to this test: https://github.com/airplanedev/cli/blob/e96b8a539339ab15acb01bc745b57a7fc25ddcf6/pkg/server/apidev/apidev_test/apidev_test.go#L41

See the ListResources test for an example of mocking out remote entities: https://github.com/airplanedev/cli/blob/e96b8a539339ab15acb01bc745b57a7fc25ddcf6/pkg/server/apiint/apiint_test/apiint_test.go#L63-L72

PR #734, opened 2022-10-17
Add "last modified at" for when tasks are registered at

kellymtrinh:
## Description
Keeps track of when the task was last registered and returns this metadata in the task endpoint. This supports showing "last modified at x" now that we support hot reloading.

Instead of storing just app errors, I created a new struct `AppMetadata` to hold any additional info we want about an a task/view.

Depends on https://github.com/airplanedev/lib/pull/352

## Test plan
- [x] Verified that loading the editor still works and I can execute tasks. With the updated web changes, the last modified at x date is shown.


pkg/server/apiint/apiint.go
```diff
@@ -528,42 +528,50 @@ func ListRunsHandler(ctx context.Context, state *state.State, r *http.Request) (
...
+		return libapi.Task{}, errors.Errorf("Task with slug %q not found", taskSlug)
 	}
 
-	// We do not use taskConfig.Def.GetUpdateTaskRequest() directly since the task definition may be invalid.
-	// We want to best-effort support invalid task definitions (e.g. unknown resources) so that we can render
+	metadata, ok := state.Metadata.Get(taskSlug)
+	if !ok {
+		return libapi.Task{}, errors.Errorf("Task with slug %q not found", taskSlug)
+	}
+	// For our purposes, the libapi.Task and libapi.UpdateTaskRequest structs contain the same critical data.
```
kellymtrinh, line 545: 😓 Looking at it more we were only using `UpdateTaskRequest` because it conveniently had `taskConfig.Def.GetUpdateTaskRequest` - which as discussed before we'd want to remove the validation logic from.

calufornia: Nice, it did feel off that we were using UpdateTaskRequest before (especially when we don’t do validation anymore).

pkg/server/state/state.go
```diff
@@ -55,6 +55,11 @@ type State struct {
 	VersionCache version.Cache
 }
 
+type AppMetadata struct {
+	RegisteredAt time.Time
```
calufornia, line 59: Actually, what about “RefreshedAt” instead of “RegisteredAt” (same for the web app PR) - encompasses that all tasks/views/workflows are refreshed, not just new ones.

calufornia, line 59: Actually, what about “RefreshedAt” instead of “RegisteredAt” (same for the web app PR) - encompasses that all tasks/views/workflows are refreshed, not just new ones.

calufornia: AppMetadata sounds a bit generic to me (and metadata historically has been for name/slug/etc.) - maybe something like AppCondition?

pkg/server/state/state.go
```diff
@@ -38,10 +38,10 @@ type State struct {
 	Runs *runsStore
 	// Mapping from task slug to task config
 	TaskConfigs Store[string, discover.TaskConfig]
-	// Warnings to display on the task page
-	TaskErrors Store[string, []dev_errors.AppError]
 	// Mapping from view slug to view config
 	ViewConfigs Store[string, discover.ViewConfig]
+	// Metadata holds extra metadata such as errors to display, time registered, etc.
```
calufornia, line 43: Update the comment to mention it only apply to tasks and workflows, not views (and rename to AppConditions)

calufornia (approved):

pkg/server/state/state.go
kellymtrinh: RefreshedAt here makes sense- in web I like `modified at` more because users don't have a concept of apps refreshing or registering (except maybe when we print out "registered 5 apps" in the CLI at startup)?

PR #737, opened 2022-10-19
Add support for referencing resources by name

calufornia:
## Description

The templates currently use the demo db name (`[Demo DB]`) - we should support this when developing from the Studio. Depends on https://github.com/airplanedev/lib/pull/355

## Test plan
Tested a template with a name of the resource and verified that we were able to resolve it.


kellymtrinh (approved):

zhan-airplane:
+1 to this just ran into this :D

PR #741, opened 2022-10-20
Update studio startup text and don't log task logs and outputs

calufornia:
## Description

Updates the startup text for the local dev studio, and also stops logging logs and outputs when running a task from the studio (these logs will still be output for headless dev).

## Test plan
image

image


cmd/airplane/tasks/dev/server.go
```diff
@@ -106,31 +108,48 @@ func runLocalDevServer(ctx context.Context, cfg taskDevConfig) error {
...
 	if len(viewConfigs) == 1 {
 		viewNoun = "view"
 	}
-	logger.Log("Found %d %s:", len(viewConfigs), viewNoun)
-	for _, view := range viewConfigs {
-		logger.Log("- %s", view.Def.Name)
-	}
+	logger.Log(
+		"registered %s %s, %s %s, and %s %s.",
+		logger.Yellow(strconv.Itoa(numTasks)),
```
kellymtrinh, line 145: Should we make all of them (tasks, workflows, and views) logger.Green? It looked strange to me to have different colors when they're all being registered. Also red (and possibly purple) seems like it should be reserved for errors.

kellymtrinh (approved):
LGTM but I think we should change some of the print colors!

PR #742, opened 2022-10-20
Add env slug to dev url

ameyakhare:
## Description

We need this to be able to open remote tasks in the correct environment.

## Test plan
Opens the studio in the correct environment context.


cmd/airplane/tasks/dev/server.go
```diff
@@ -192,7 +192,7 @@ func runLocalDevServer(ctx context.Context, cfg taskDevConfig) error {
 	}
 
 	logger.Log("")
-	studioURL := fmt.Sprintf("%s/studio?host=http://localhost:%d", appURL, cfg.port)
+	studioURL := fmt.Sprintf("%s/studio?host=http://localhost:%d&__env=%s", appURL, cfg.port, cfg.envSlug)
```
calufornia, line 195: I would use `remoteEnv.Slug` here instead - `cfg.envSlug` might be unset

calufornia (approved):

PR #747, opened 2022-10-24
Set task slug in PrepareRunOptions

calufornia:
## Description
This is now required by `PrepareRun` to create task subdirectories.

## Test plan
Tested in conjunction with https://github.com/airplanedev/lib/pull/361


calufornia:
Will update lib after https://github.com/airplanedev/lib/pull/361 is landed

leeweisberger (approved):

PR #748, opened 2022-10-25
Add feature flagging. Cache cli version

leeweisberger:
## Description
This pr contains two changes
1. Add feature flagging capability. Cache feature flags in the airplane config file for an hour. Run the bundle deploy command if the corresponding feature flag is on
2. Cache CLI version in the airplane config file rather than checking for new versions of the cli on each execution for one day.

## Test plan
I manually tested both of these


cmd/airplane/root/deploy/deployer_test.go
```diff
@@ -233,3 +232,52 @@ func TestDeploy(t *testing.T) {
 		})
 	}
 }
+
+func TestParseRemote(t *testing.T) {
```
leeweisberger, line 236: Moved from tasks/deploy

cmd/airplane/tasks/deploy/deploy.go
```diff
@@ -99,14 +103,23 @@ type taskDeployedProps struct {
 }
 
 func run(ctx context.Context, cfg config) error {
+	l := logger.NewStdErrLogger(logger.StdErrLoggerOpts{WithLoader: true})
+	defer l.StopLoader()
+
+	if cfg.root.Flagger.Bool(ctx, l, flagsiface.DeployBundles) {
```
leeweisberger, line 109: If the flag is on, run bundle deploy

pkg/flags/flagsiface/flags.go
```diff
@@ -0,0 +1,34 @@
+package flagsiface
```
leeweisberger, line 1: copied this interface from airport

ameyakhare: should be in lib then, no?

ameyakhare (approved):
seems fine

pkg/flags/flagsiface/flags.go
leeweisberger: It should be :/

PR #749, opened 2022-10-25
Remove getting started flag from views

rose-huang:
## Description
Should use the `airplane init template=views_getting_started` command instead

## Test plan
```
➜  cli git:(main) ✗ bin/airplane views init --getting-started

Error: Unknown flag: --getting-started

```


cmd/airplane/views/initcmd/init.go
```diff
@@ -64,16 +57,7 @@ func GetConfig(client *api.Client) config {
 }
 
 func Run(ctx context.Context, cfg config) error {
-	if cfg.gettingStarted {
-		_, err := createDemoDB(ctx, cfg)
```
zhan-airplane, line 68: linter says `createDemoDB` is dead code after this, let's remove it.

zhan-airplane (approved):
Make sure docs gets updated too, if this flag appears anywhere there!

PR #750, opened 2022-10-25
Change filewatcher library

kellymtrinh:
## Description
- Moves from `github.com/rjeczalik/notify v0.9.3-0.20210809113154-3472d85e95cd`, which watches for filesystem notifications to `github.com/radovskyb/watcher v1.0.7`, **which periodically polls and doesn't use file system notifications**.
- Fixed some lint errors it was complaining about, unrelated to this change
- Made a new `Event` struct since each file watching lib has a slightly different set. If we need to change it out in the future, we should be able to only touch `filewatcher.go` 

What this addresses:
A user ran into this error:
```
Fcntl ... too many open files
```
Internally we end up bumping our `ulimit`, but users shouldn't need to use cli.

CC: @leeweisberger 
## Test plan
- [x] Checked that node_modules, pycache, the hidden airplane folders are ignored and don't trigger infinite loops
- [x] Updating tasks and view def files triggers changes, both in new folders and in the working directory work
- [x]  New files made through `airplane views init` and `airplane tasks init` are picked up and registered.


pkg/server/filewatcher/filewatcher.go
```diff
@@ -17,79 +17,128 @@ type FileWatcher interface {
...
+// Operations
+const (
+	NoOp Operation = iota
+	Create
+	Write
+	Remove
+	Move
+)
+
+// Event describes the change that occured.
```
kellymtrinh, line 32: Made some operations and an `Event` struct so we can use this instead of an individual library's Event struct. They all seem to have similar but slightly different ones.

pkg/server/filewatcher/filewatcher_test.go
```diff
@@ -12,34 +12,6 @@ func TestFilter(tt *testing.T) {
 		path   string
 		expect bool
 	}{
-		{
```
calufornia, line 15: How come we removed these test cases?

pkg/server/filewatcher/filewatcher.go
```diff
@@ -69,27 +95,50 @@ func (f *AppWatcher) Watch(wd string) error {
...
 	logger.Log(logger.Green("Changes to tasks, workflows, and views will be applied automatically."))
 
-	if err := notify.Watch(wd+"/...", f.changes, notify.All); err != nil {
+	// Ignore hidden files and known directories
+	f.watcher.IgnoreHiddenFiles(true)
+	directoriesToIgnore := make([]string, 0, len(discover.IgnoredDirectories))
+	for k := range discover.IgnoredDirectories {
+		directoriesToIgnore = append(directoriesToIgnore, k)
+	}
+	if err := f.watcher.Ignore(directoriesToIgnore...); err != nil {
```
calufornia, line 104: Ignoring `node_modules` doesn't seem to be working for me locally - when I do `aplocal dev ~/airplane/node_test`, and remove/add `~/airplane/node_test/node_modules`, changes are still detected (by adding a log below `case e := <-f.watcher.Event:`

calufornia: I also think this doesn't work for subdirectories: https://github.com/radovskyb/watcher/issues/82. So if I have a folder `~/airplane/node_test/subdir/node_modules`, changes to those node modules wouldn't get picked up. I think what we would need to do is recurse through all directories and add all `node_modules` paths we find — however, that still wouldn't handle the case where a `node_modules` doesn't exist and gets created during the studio session, but maybe that's ok.

pkg/server/filewatcher/filewatcher_test.go
kellymtrinh: It was supposed to be covered by the new watching lib - since we can ignore them now

kellymtrinh: Bringing it back for the below comment since this lib doesn't support ignored dirs that are nested

pkg/server/filewatcher/filewatcher.go
kellymtrinh: Updated- with the latest commit I think this should ignore any node_modules/ignored dirs at the working directory level. And we'll continue to filter in `IsValidDefinitionFile`, which hasn't changed- since the previous lib also didn't allow ignoring.

pkg/server/filewatcher/filewatcher.go
```diff
@@ -69,27 +112,50 @@ func (f *AppWatcher) Watch(wd string) error {
 	logger.Log(logger.Green("Watching for changes in: %s", wd))
 	logger.Log(logger.Green("Changes to tasks, workflows, and views will be applied automatically."))
 
-	if err := notify.Watch(wd+"/...", f.changes, notify.All); err != nil {
+	// Ignore hidden files and known directories
+	f.watcher.IgnoreHiddenFiles(true)
+	directoriesToIgnore := make([]string, 0, len(discover.IgnoredDirectories))
+	for k := range discover.IgnoredDirectories {
```
calufornia, line 118: use `dir` instead of `k`

pkg/server/filewatcher/filewatcher.go
```diff
@@ -69,27 +112,50 @@ func (f *AppWatcher) Watch(wd string) error {
...
+		directoriesToIgnore = append(directoriesToIgnore, wd+"/"+k)
```
calufornia, line 119: use `filepath.Join`

calufornia (approved):

PR #754, opened 2022-10-26
Retry install on missing symlinks

leeweisberger:
## Description
This is to specifically unblock a customer who is running into issues developing views on a windows vm.

npm install / yarn fails for them with an error that they can't install symlinks. I was able to confirm that we could run npm install with the `--no-bin-links` flag and it worked on his machine.

In this pr I'm auto-retrying install commands with `--no-bin-links` if the installation fails due to the error he was getting

https://github.com/MicrosoftDocs/WSL/issues/26#issuecomment-206991754

## Test plan

I'm not able to repro this so I couldnt' really test. I did test that given an error with a substring the install commands run again.


pkg/node/node.go
```diff
@@ -120,38 +124,45 @@ func getPackagesToAdd(packagesToCheck, existingDeps []string) []string {
...
 		cmd = exec.Command("npm", installArgs...)
 		l.Debug("Adding packages using npm")
 	}
+	if opts.NoBinLinks {
+		l.Debug("Installing with --no-bin-links")
+	}
 
 	cmd.Dir = packageJSONDirPath
-	err := cmd.Run()
+	output, err := cmd.CombinedOutput()
```
leeweisberger, line 153: I improved the error message. Before we were outputing just "exist status 1". Now we're outputing the node/yarn logs if there is an error

pkg/node/node.go
```diff
@@ -120,38 +124,45 @@ func getPackagesToAdd(packagesToCheck, existingDeps []string) []string {
...
+			return addPackages(l, packageJSONDirPath, packageNames, dev, opts)
 		}
-		return err
-	}
-	for _, pkg := range packageNames {
-		l.Step(fmt.Sprintf("Installed %s", pkg))
+
+		return errors.New(errString)
 	}
+	l.Step(fmt.Sprintf("Installed %s", strings.Join(packageNames, ", ")))
```
leeweisberger, line 165: Log installed deps on one line instead of N, Also unrelated

pkg/utils/dependencies.go
```diff
@@ -37,18 +52,32 @@ func InstallDependencies(dir string, useYarn bool) error {
...
+	if err := cmd.Wait(); err != nil {
+		errString := strings.Join(allLogs, "\n")
+		if !opts.NoBinLinks && strings.Contains(errString, SymlinkErrString) {
+			// Try installation again with NoBinLinks to get passed the symlink error.
+			opts.NoBinLinks = true
+			return InstallDependencies(dir, opts)
+		}
+		return errors.New(errString)
+	}
+	return nil
```
leeweisberger, line 80: Similar change to above, we're now returning the full logs on error instead of a generic "status 1" error

zhan-airplane (approved):

PR #756, opened 2022-10-27
When initializing a task, write node version to airplane.config.yaml

leeweisberger:
## Description
When creating a new task, create a new airplane.config.yaml and write the node version there
When there is already an airplane.config.yaml with no node version, update that airplane.config.yaml with the correct node version
When there is already an airplane.config.yaml or package.json with a conflicting node version, do nothing and warn

## Test plan
Manually tested all of the above cases


pkg/node/node.go
```diff
@@ -17,53 +17,126 @@ import (
...
+type PackageJSONOptions = struct {
+	Dependencies NodeDependencies
+}
+
 // CreatePackageJSON ensures there is a package.json in path with the provided dependencies installed.
 // If package.json exists in cwd, use it.
 // If package.json exists in parent directory, ask user if they want to use that or create a new one.
 // If package.json doesn't exist, create a new one.
-func CreatePackageJSON(directory string, dependencies NodeDependencies) error {
+func CreatePackageJSON(directory string, packageJSONOptions PackageJSONOptions) error {
```
leeweisberger, line 40: This function is just a refactor. No functionality change.

cmd/airplane/tasks/initcmd/init.go
```diff
@@ -836,21 +836,37 @@ func apiTaskToRuntimeTask(task *libapi.Task) *runtime.Task {
...
+			},
+		}); err != nil {
+			return err
+		}
+
+		_, nodeVersion, err := def.GetBuildType()
+		if err != nil {
+			return err
+		}
+		if err := node.CreateAirplaneConfig(cwd, isNew, deployconfig.AirplaneConfig{
```
ameyakhare, line 860: what does isNew do here?

ameyakhare, line 860: shouldnt we not be creating this if it already exists?

pkg/node/node.go
```diff
@@ -263,27 +336,27 @@ func mergeTSConfig(configFile []byte, strategy MergeStrategy) error {
...
 		}
-		logger.Step("Added tsconfig.json...")
+		logger.Step("Created tsconfig.json")
 	}
 
 	return nil
 }
 
-func mergeTSConfigsRecursively(dest, src map[string]interface{}) {
+func mergeStructsRecursively(dest, src map[string]interface{}) {
```
ameyakhare, line 351: nit: these arent really structs so if you want to generalize, make it mergeMapsRecusively

cmd/airplane/tasks/initcmd/init.go
ameyakhare: nvm, it looks like CreateAirplaneConfig checks if there is already another one, a little confusing though imo based on the function name (maybe CreateOrMergeAirplaneConfig? idk)

ameyakhare (changes requested):
some open questions

cmd/airplane/tasks/initcmd/init.go
leeweisberger: I updated the code to remove it. Don't think it's necessary

leeweisberger: Yeah updated

ameyakhare (approved):
nice

PR #758, opened 2022-10-29
Fix panic in latest check

leeweisberger:
If no userconfig is passed in (which is the case in local dev using the studio) ignore and don't update the userconfig.


pkg/version/latest/latest.go
```diff
@@ -31,21 +31,22 @@ func CheckLatest(ctx context.Context, userConfig *conf.UserConfig) bool {
...
+	latest, err := getLatest(ctx)
+	if err != nil {
+		analytics.ReportError(err)
+		logger.Debug("An error ocurred checking for the latest version: %s", err)
+		return true
+	} else if latest == "" {
+		// Pass silently if we can't identify the latest version.
+		return true
+	}
+	if userConfig != nil {
```
leeweisberger, line 49: This fixes the panic

kellymtrinh (approved):

PR #762, opened 2022-11-01
Add generic steps for a builtin task

francescaleung:
## Description

Some generic suggested steps after initializing a builtin.

## Test plan

Tested GraphQL inits + deploys from definition.


yolken-airplane:
@francescaleung I got this in https://github.com/airplanedev/airport/pull/3508 and https://github.com/airplanedev/cli/pull/761 (we need to go to an even newer version to pick up a workflows fix).

francescaleung:
@yolken-airplane that was only half of this PR, but thanks

leeweisberger (approved):

PR #776, opened 2022-11-03
Add local dev fix for prompts without reviewers

ericliuche:
## Description
Applies default settings to prompts without users

## Test plan
Ran locally


pkg/server/apiext/apiext.go
```diff
@@ -457,6 +457,21 @@ func GetPromptHandler(ctx context.Context, state *state.State, r *http.Request)
 
 	for _, p := range run.Prompts {
 		if p.ID == promptID {
+			reviewersWithDefaults := p.Reviewers
```
colinking: You'll need to fix `ListPrompts` too

colinking: Also we should fix `web` so it works with older CLIs.

ericliuche:
ah one sec, gotta fix /prompts/list

pkg/server/apiext/apiext.go
colinking: Do you need a separate variable? Can you just set `p.Reviewers` directly?

colinking (approved):
nice

PR #779, opened 2022-11-04
Add display validation to dev

colinking:
## Description

Matches the updated validation in https://github.com/airplanedev/airport/pull/3530

Resolves TSK-800

## Test plan

Run a task locally that produced a large markdown doc. Ensure an error is returned.


colinking:
* main
  * **PR #779** <a href="https://app.graphite.dev/github/pr/airplanedev/cli/779" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/cli/779?utm_source=stack-comment).

pkg/server/apiext/apiext.go
```diff
@@ -370,9 +370,24 @@ func CreateDisplayHandler(ctx context.Context, state *state.State, r *http.Reque
...
+
+		maxMarkdownLength := 100000
+		if len(display.Content) > maxMarkdownLength {
+			return CreateDisplayResponse{}, errors.Errorf("content too long: expected at most %d characters, got %d", maxMarkdownLength, len(display.Content))
+		}
 	case "table":
 		display.Rows = req.Display.Rows
 		display.Columns = req.Display.Columns
+
+		maxRows := 10000
```
ameyakhare, line 382: I feel like these numbers should be from the displays package instead of copied here

ameyakhare: or like we should write some sort of validation in the displays package and then call that validation here

ameyakhare (approved):

pkg/server/apiext/apiext.go
colinking: Agreed, but that would require us to extract displays from `airport`. Not going to do that now.

PR #781, opened 2022-11-07
Set runtime for missing task creation

ericliuche:
## Description
Correctly set runtime field when creating a missing task. Before, this was causing for workflows to show up as tasks

## Test plan
ran locally


colinking (approved):
thanks!

PR #784, opened 2022-11-08
Don't attempt to open browser in sandbox

calufornia:
## Description

Adds the ability to bypass the "open browser" behavior when launching the studio, which will not be used by the remote sandbox, i.e. now we just start the local dev server and wait for a sigterm. If we don't do this, then an error occurs inside the sandbox container:
```
Started studio session at https://app.airplane.dev/studio?host=http://localhost:4000&__env=asdf (^C to quit)
Press ENTER to open the studio in the browser.

Something went wrong. Try running the command with the --debug flag for more details.
```

We gate this behind the `--sandbox` flag and hide it, since it should only be used by the remote studio.

Also removes the port from the local dev server state, which is unused.

## Test plan

Tested with and without `--sandbox` flag and verified behavior was as expected.


ameyakhare (approved):

joshma:
is this just for the ENTER behavior? if so maybe it makes sense to just not print it out if there's no TTY detected? similar to how we detect interactivity

calufornia:
> is this just for the ENTER behavior? if so maybe it makes sense to just not print it out if there's no TTY detected? similar to how we detect interactivity

We’ll likely use it for other sandbox-specific behavior as well, but just this to start.

joshma:
👍

calufornia:
@joshma One thing I just remembered we'll also need the flag for - the sandbox needs the dev server to listen on `0.0.0.0` (as opposed to localhost) so that port 4000 is accessible outside of the container (unless there's a better way of doing that).  We don't want to use `0.0.0.0` for local dev because I notice it prompts the user to accept incoming connections, etc.

^Will include this in a separate PR

PR #786, opened 2022-11-08
Fix goreleaser by pulling in lib v0.5.112

kellymtrinh:
## Description
Undos the snowflake dep upgrade, which wasn't compiling on windows:
https://github.com/airplanedev/lib/pull/400

## Test plan
- [x] Compiles


ameyakhare (approved):
send it

PR #790, opened 2022-11-10
Add ability to download deployed code in airplane init

calufornia:
## Description

Adds a hidden flag, `--download`, to airplane init, which will be used by the sandbox to download code from existing entities. Currently, it download all entities, but in the future we could also expose this flag to users who want to "eject" code that they've deployed back to their machine.

## Test plan
Tested locally and verified all entities were downloaded into specified workspace directory.


cmd/airplane/root/initcmd/download.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	urls := resp.SignedURLs
+
+	if err := os.Mkdir(cfg.workspace, 0755); err != nil {
+		if !os.IsExist(err) {
+			return errors.Wrap(err, "creating workspace directory")
+		}
+	}
+
+	for _, url := range urls {
```
ameyakhare, line 28: oof this is serial, there might be a lot

ameyakhare (approved):

cmd/airplane/root/initcmd/download.go
calufornia: Changed to download in parallel

PR #791, opened 2022-11-10
Fix issue with builtins race condition

kellymtrinh:
## Description

Depends on https://github.com/airplanedev/lib/pull/403
Imports fixes for a builtins race condition.

## Test plan
See https://github.com/airplanedev/lib/pull/403


cmd/airplane/tasks/deploy/deploy.go
```diff
@@ -197,7 +197,7 @@ func HandleMissingTask(cfg config, l logger.LoggerWithLoader, createdTasks *map[
 		}
 
 		l.Log("Creating task...")
-		utr, err := def.GetUpdateTaskRequest(ctx, cfg.client)
+		utr, err := def.GetUpdateTaskRequest(ctx, cfg.client, false)
```
kellymtrinh, line 200: I checked that this should be false from @leeweisberger

leeweisberger: lgtm

calufornia (approved):

PR #792, opened 2022-11-15
Add ability to list studio file tree from local dev server

calufornia:
## Description

Adds a new endpoint, `/dev/files/list`, that lists out all files in the current "development root" by recursing through it and all of its subdirectories. For each file node, we include an `Entities` field that will track entities (tasks, workflows, views) that are defined by that file. This is different than the `/dev/list` endpoint we had previously, which would list entities per entrypoint, which was particularly confusing for non-inline tasks because the task/view/etc. is actually defined by the task definition file, and not the entrypoint.

We'll eventually remove `/dev/list` in favor of `/dev/files/list`, and we'll want to make corresponding UI changes.

Depends on https://github.com/airplanedev/lib/pull/413

## Test plan
`curl http://localhost:4000/dev/files/list` returned a correct tree structure.


calufornia:
> nice, are you fixing the inline code defn file missing thing in lib?

Yep, here: https://github.com/airplanedev/lib/pull/413

PR #794, opened 2022-11-16
Handle local view routing for editor

zhan-airplane:
## Description
What's needed for local view routing in editor:
* If possible, tries to hit the remote API for task metadata or views, falling back if it isn't found
* Handles the `/v0/hosts/web` endpoint

## Test plan
Ran locally

---
cc @leeweisberger @stepzhou @geodn


pkg/server/apiext/apiext.go
```diff
@@ -279,25 +281,60 @@ func GetRunHandler(ctx context.Context, state *state.State, r *http.Request) (de
...
+// If a view is not local, it tries the fallback environment, so that local views
+// can route correctly to the correct URL.
 func GetViewHandler(ctx context.Context, state *state.State, r *http.Request) (libapi.View, error) {
 	slug := r.URL.Query().Get("slug")
 	if slug == "" {
 		return libapi.View{}, errors.New("slug cannot be empty")
 	}
 
+	_, ok := state.ViewConfigs.Get(slug)
+	// Not local, we try using the fallback env first, but we default to returning a dummy view if it's not found.
```
rose-huang, line 320: so if I'm understanding...
it's trying to check if there's a local view with that slug and if so, it uses that one
if not, it tries to find a remote view and if it exists, it uses the remote view

if both local view and remote view don't exist - what does this "default" dummy view look like?

rose-huang: Is this the same logic (approximately) that executing tasks in views uses in studio?

rose-huang (changes requested):
just a few questions

pkg/server/apiext/apiext.go
zhan-airplane: Yeah I think so? I'm not changing the behavior from what it was previously in this case - it just returns some view with ID and slug set and nothing else (I think something in the discoverer might rely on it returning this, so I'm trying to avoid changing the behavior here)

rose-huang (approved):

PR #795, opened 2022-11-16
Hot reload entities when entrypoints change

calufornia:
## Description
For SQL tasks, we want to hot reload the SQL task when the `.sql` files changes, not only when the definition file changes.  To do this, we re-discover any entities whose entrypoints have been modified (we should do this in general anyway).

## Test plan
Tested hot reloading by modifying a `.sql` file with a name that differs from its task definition file — previously, hot reloading only worked if both the `.sql` file and definition file were named the same.


pkg/server/server.go
```diff
@@ -288,10 +288,25 @@ func (s *Server) ReloadApps(ctx context.Context, path string, wd string, e filew
 		}
 	} else {
 		reload = func() {
-			taskConfigs, viewConfigs, err := s.DiscoverTasksAndViews(ctx, path)
+			pathsToDiscover := []string{path}
```
kellymtrinh, line 291: Will this still work if you should reload the entire directory? (ie. a file is renamed or deleted)

calufornia: Yep, this still works since we're strictly adding to the list of paths that we discover (just tested deleting a task yaml)

PR #800, opened 2022-11-18
Stop setting default timeout at deploy time

kellymtrinh:
## Description
The API already sets a default timeout if it isn't defined.
This pulls in lib changes to stop doing that:
https://github.com/airplanedev/lib/pull/422/files

The change here is to NOT show a diff if the user doesn't have the timeout defined in their yaml. 
Previously, we were relying on setting the default at deploy time to be 3600 to calculate as no diff.

- [x] TODO: use published version of lib 
## Test plan
- [x] No change: deploying a task works. I deployed and ran the task. Same with a workflow.
- [x] Deploying a task with no timeout specified doesn't show the diff about timeout being different
- [x] Deploying a task where I change the timeout to something like 3000 (instead of the default 3600) does show a diff 
- [x] workflows are inline, which doesn't show diffs


kellymtrinh:
This felt a little heavy handed to compare the timeout diff, but it has to check that the NEW task yaml that the user is trying to deploy is zero and that the old def that was already deployed has the default. This catches both cases of the old deployed task having the default 3600 timeout AND: 1. the user not having timeout defined at all in the yaml 2. the user having timeout defined as 3600

Things that didn't work:
- Normalizing the timeout has to happen before generating the diff lines. If it generates the diff lines and you try to ignore just "-timeout: 3600" and not print it out,  it still prints the name diff:
```
  --- a/new_task.task.yaml
  +++ b/new_task.task.yaml
```
- Originally I tried the approach of  just working with the old deployed task def- and seeing if it was the default of 3600 setting it to 0. This shows a diff if the user DOES have the timeout defined as 3600 in their yaml (as opposed to not having a timeout defined at all).

leeweisberger (approved):
Makes sense

nucleartide:
Hi Kelly!!! 😂 GitHub recommended me this PR haha

kellymtrinh:
> Hi Kelly!!! 😂 GitHub recommended me this PR haha

🙈

PR #801, opened 2022-11-22
Install and download builtins once on startup

ameyakhare:
Resolves TSK-1104

## Description
Fixes builtin binary corruption by storing one builtin executor on the state and downloading builtins on server startup.

## Test plan
Builtins and regular tasks still work locally.


calufornia (approved):
Remember to bump lib for landing this

PR #808, opened 2022-11-28
Create tsconfig in same directory as package.json

rose-huang:
## Description
We previously created the tsconfig in the current directory where you run airplane views/tasks init. However, it seems more expected to create it in the same directory as the package.json that we're using and ask questions if there are conflicts 

cc @ameyakhare @leeweisberger 

## Test plan
- Creating a new view/task where you use a package.json in a parent directory creates / updates the tsconfig in the parent directory and asks questions 
- Creating a new view/task where you create/use a new package.json in cwd creates/updates the tsconfig  in cwd as well

```
➜  subdir4 ~/airplanedev/cli/bin/airplane tasks init
? What should this task be called? node task 6
? What kind of task should this be? Node
? Where is the script for this task? node_task_6.ts
- Created node_task_6.ts
? Where should the definition file be created? node_task_6.task.yaml
- Created node_task_6.task.yaml
? Found existing package.json in /Users/rosehuang/test/test1128_2. Use this to manage dependencies? Yes
- Created /Users/rosehuang/test/test1128_2/tsconfig.json

➜  subdir4 ~/airplanedev/cli/bin/airplane views init
? What should this view be called? view 4
- Created directory view_4
- Created view definition at view_4/view_4.view.yaml
- Created view entrypoint at view_4/view_4.view.tsx
? Found existing package.json in /Users/rosehuang/test/test1128_2. Use this to manage dependencies? Yes

compilerOptions.lib:  (["ESNext"]) -> (["DOM","DOM.Iterable","ESNext"])
compilerOptions.jsx:  (null) -> ("react-jsx")

? Would you like to override /Users/rosehuang/test/test1128_2/tsconfig.json with these changes? Yes
- Updated /Users/rosehuang/test/test1128_2/tsconfig.json
- Created .gitignore

➜  subdir4 ~/airplanedev/cli/bin/airplane tasks init
? What should this task be called? node task 7
? What kind of task should this be? Node
? Where is the script for this task? node_task_7.ts
- Created node_task_7.ts
? Where should the definition file be created? node_task_7.task.yaml
- Created node_task_7.task.yaml
? Found existing package.json in /Users/rosehuang/test/test1128_2. Use this to manage dependencies? No, create package.json in my working directory
- Created package.json
- Installing dependencies...
- Installed airplane
- Installed @types/node
- Created tsconfig.json

➜  subdir4 ~/airplanedev/cli/bin/airplane views init
? What should this view be called? view 5
- Created directory view_5
- Created view definition at view_5/view_5.view.yaml
- Created view entrypoint at view_5/view_5.view.tsx
- Installing dependencies...
- Installed @airplane/views, react, react-dom
- Installed @types/react, @types/react-dom, typescript

compilerOptions.jsx:  (null) -> ("react-jsx")
compilerOptions.lib:  (["ESNext"]) -> (["DOM","DOM.Iterable","ESNext"])

? Would you like to override tsconfig.json with these changes? No
```


pkg/node/node.go
```diff
@@ -37,7 +37,7 @@ type PackageJSONOptions = struct {
 // If package.json exists in cwd, use it.
 // If package.json exists in parent directory, ask user if they want to use that or create a new one.
 // If package.json doesn't exist, create a new one.
-func CreatePackageJSON(directory string, packageJSONOptions PackageJSONOptions) error {
```
colinking, line 40: nit: update function comment to explain the `string` returned

pkg/node/node.go
```diff
@@ -291,8 +291,20 @@ const (
...
+	cwd, err := os.Getwd()
+	if err != nil {
+		return errors.Wrap(err, "getting working directory")
+	}
+	var configPath string
+	// Leave out the directory path if tsconfig is being created in cwd for cleaner logs
+	if filepath.Dir(configDir) == cwd {
+		configPath = "tsconfig.json"
+	} else {
+		configPath = filepath.Join(configDir, "tsconfig.json")
```
colinking, line 304: Can you do a relative path instead?

Something like:

```
Would you like to override ../tsconfig.json with these changes?
```

colinking (approved):
This is great. Thank you!

pkg/node/node.go
colinking: Looking at the logs in your PR desc, the `package.json` paths are not relative either. Could you change that too? To keep them consistent. 🙏🏻

rose-huang: @colinking I was worried that relative paths would be a bit unclear if the package.json that is found is many directories above (such as `../../../../../package.json`). We could show the relative path as a default but switch to the absolute path if it's more than 2 or 3 directories above?

colinking: Good point. ~2 directories seems like a reasonable limit.

PR #810, opened 2022-11-28
Remove hard dependency on builtins being supported for local dev

ameyakhare:
## Description
We don't want to prevent windows users from running a task locally if builtins aren't supported on the particular machine.

## Test plan
Local task execution still works.


pkg/dev/dev.go
```diff
@@ -47,15 +47,17 @@ type LocalExecutor struct {
...
 }
 
-func NewLocalExecutor(workingDir string) (Executor, error) {
-	builtinsClient, err := builtins.NewLocalClient(workingDir, goruntime.GOOS, goruntime.GOARCH, logger.NewStdErrLogger(logger.StdErrLoggerOpts{}))
+func NewLocalExecutor(workingDir string) Executor {
+	l := logger.NewStdErrLogger(logger.StdErrLoggerOpts{})
+	builtinsClient, err := builtins.NewLocalClient(workingDir, goruntime.GOOS, goruntime.GOARCH, l)
 	if err != nil {
-		return nil, err
+		l.Warning("Local builtin execution is not supported on this machine. Builtin executions will error.")
```
kellymtrinh, line 54: Can you include the bit about reaching out to support? (It's in the err message returned by `NewLocalClient`)

kellymtrinh (approved):

PR #816, opened 2022-11-29
Implement dev server cancel endpoint

ameyakhare:
## Description
Implements the dev server cancel endpoint, which allows user to cancel a local run.

## Test plan
Cancelled a run that was sleeping locally.


calufornia:
Can you add `CancelledAt` and `CancelledBy` to the `dev.LocalRun` struct (similar to [here](https://github.com/airplanedev/cli/blob/dbc9629a9f8b526565868af328f4cf50d6fe0ade/pkg/api/types.go#L195-L196))? IIRC we may need those for the web app to properly display all cancellation info

calufornia (approved):

PR #819, opened 2022-11-30
Cleanup execute task handler

kellymtrinh:
## Description
Cleans up the execute task API. This is a no-op, but should hopefully make it less confusing what depends on the LocalRun struct.
- execute/task should just return runID without any other fields (this PR)

Related, but not changed: 
- internal [runs/get ](https://github.com/airplanedev/airport/blob/34fcd182a556a30a5d8869d6f75e52ce68a2c1ad/pkg/api/apiint/runs.go#L61-L86)should return runID (it was always doing this)
- external [runs/get ](https://github.com/airplanedev/airport/blob/34fcd182a556a30a5d8869d6f75e52ce68a2c1ad/pkg/api/apiext/runs.go#L155-L184)should return id?? (it is doing this now as of https://github.com/airplanedev/cli/pull/818 - in airport it’s using toRunResponse which returns [GetRunResponse](https://github.com/airplanedev/airport/blob/34fcd182a556a30a5d8869d6f75e52ce68a2c1ad/pkg/api/apiext/runs.go#L81))

## Test plan
- [x] Test that executing a task works in studio. Double checked that web only uses the `runID` field from the tasks/execute response.
I double checked that the [JS SDK](https://github.com/airplanedev/js-sdk/blob/e24344d2a59c86c7634d20adbfe21fc49934f2b8/packages/airplane/internal/api/client.ts#L42-L59) only and [python SDK ](https://github.com/airplanedev/python-sdk/blob/7ad260307bae75e1bb3d93dd07ad70022c2a876e/airplane/api/client.py#L97-L106)only expect `runID` from the response.
- [x] Tested that executing a python slack builtin still works.
- [x] Tested JS builtins still work.


pkg/server/apiext/apiext.go
```diff
@@ -236,20 +240,20 @@ func ExecuteTaskHandler(ctx context.Context, state *state.State, r *http.Request
 		})
 		if err != nil {
 			if _, ok := err.(*libapi.TaskMissingError); ok {
-				return dev.LocalRun{}, errors.Errorf("task with slug %s is not registered locally or remotely", req.Slug)
+				return ExecuteTaskResponse{RunID: runID}, errors.Errorf("task with slug %s is not registered locally or remotely", req.Slug)
```
calufornia, line 243: Can we verify that returning the runID here (which we didn't do before) doesn't affect anything (e.g. with child runs)?

pkg/server/apiext/apiext.go
```diff
@@ -67,12 +67,16 @@ func getRunIDFromToken(r *http.Request) (string, error) {
 	return "", nil
 }
 
+type ExecuteTaskResponse struct {
```
calufornia, line 70: Use identical `RunTaskResponse` from api/types.go instead

calufornia (approved):

pkg/server/apiext/apiext.go
kellymtrinh: Yep, verified child runs works.`runID` was always returned, it was https://github.com/airplanedev/cli/pull/818/files `id` that got added as an additional thing!

calufornia: Sorry, I meant that we used to return an empty `dev.LocalRun{}` struct, but now we're returning the runID — was that intentional?

PR #823, opened 2022-12-01
Misc. fixes to runbook converter tool

yolken-airplane:
## Description
This change includes miscellaneous fixes to the converter tool added in https://github.com/airplanedev/cli/pull/815 based on @colinking's feedback. I'll update the latter change with notes on each fix that was included here.

cc: @kellymtrinh 

## Test plan
Testing to be done by running locally.


pkg/rb2wf/utils.go
```diff
@@ -59,18 +61,27 @@ func jsObj(obj interface{}) (string, error) {
...
 	return result, nil
 }
 
 func jsStr(obj interface{}) (string, error) {
 	switch v := obj.(type) {
 	case templateValue:
-		return fmt.Sprintf("`%s`", v.toTemplate()), nil
+		templateContents := v.toTemplate()
+		if strings.Contains(templateContents, "${") {
+			// Use backtick quotes
```
colinking, line 79: FWIW, prettier has logic for standardizing quotes: https://prettier.io/docs/en/rationale.html#strings

You shouldn't need any additional logic for picking quotes.

colinking (approved):

pkg/rb2wf/utils.go
colinking: Ah I see, the prettier rule does only apply to single/double quotes. There's a separate eslint rule for enforcing minimal template string quote usage: https://eslint.org/docs/latest/rules/quotes

PR #828, opened 2022-12-02
Add form to prompt support to runbook convertor tool

kellymtrinh:
## Description
- Adds support for runbook forms -> prompts. https://linear.app/airplane/issue/TSK-1203

I'm going to do the env variables in a separate PR and another pass on error messaging.

Doesn't support form template options yet:
Screenshot 2022-11-30 at 4 48 23 PM
This was introduced in https://github.com/airplanedev/airport/pull/2044/files and the lib Parameter and airport Parameter field have diverged and the client used everywhere can't unmarshal the session parameter option constraints. The long term fix is to consolidate this. I didn't want to take a short term fix, since it seems like a newer feature for runbook blocks?.

## Test plan
- Created a runbook with a form with different parameter types. Confirmed that the tool converted them. Deployed the workflow, then checked that the forms are the same.
Warning is shown for ignored fields
Screenshot 2022-12-01 at 5 35 31 PM

Side by side (runbook left, workflow right), with some different parameters. The prompt skips the config and upload types.
Screenshot 2022-12-01 at 5 52 20 PM


colinking (commented):
(Leaving some comments from Friday, haven't finished reviewing yet)

pkg/rb2wf/converter.go
```diff
@@ -203,12 +207,72 @@ func (r *RunbookConverter) getRunbookInfo(
...
+	return o, nil
+}
+
 func (r *RunbookConverter) blockToString(
 	ctx context.Context,
 	block api.SessionBlock,
 	resources map[string]libapi.Resource,
 ) (string, error) {
-	// TODO: Add support for form blocks.
+
```
colinking, line 275: delete newline

pkg/rb2wf/converter.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/airplanedev/cli/pkg/api"
 	"github.com/airplanedev/cli/pkg/logger"
 	libapi "github.com/airplanedev/lib/pkg/api"
+	"github.com/iancoleman/orderedmap"
```
colinking, line 17: We created our own package for this awhile ago -- use that instead: https://github.com/airplanedev/ojson

pkg/rb2wf/templates/blocks/form.ts.tmpl
```diff
@@ -0,0 +1,14 @@
...
+
+if ({{.StartCondition}}) {
+  let {{.BlockSlug}} = await airplane.prompt(
+    {{jsObj .Params}},
+  );
+} else {
+  console.log("Skipping over '{{.BlockSlug}}' because startCondition is false");
+}
+{{else}}
+let {{.BlockSlug}} = await airplane.prompt(
```
colinking, line 11:
```suggestion
const {{.BlockSlug}} = await airplane.prompt(
```

pkg/rb2wf/templates/blocks/form.ts.tmpl
```diff
@@ -0,0 +1,14 @@
+{{if .StartCondition}}
+
+if ({{.StartCondition}}) {
+  let {{.BlockSlug}} = await airplane.prompt(
```
colinking, line 4:
```suggestion
  const {{.BlockSlug}} = await airplane.prompt(
```

colinking (approved):
Nice job!

PR #830, opened 2022-12-03
Update runbook converter tool to replace template globals with env vars

kellymtrinh:
## Description

Addresses: https://github.com/airplanedev/cli/pull/815#discussion_r1035994134


pkg/rb2wf/converter.go
```diff
@@ -168,6 +168,10 @@ func (r *RunbookConverter) Convert(ctx context.Context, runbookSlug string) erro
 		return err
 	}
 
+	if strings.Contains(mainTsStr, manualFixPlaceholder) {
+		// TODO: include link to conversion tool docs with "see docs for more details."
+		logger.Warning("Some runbook features could not be converted directly to a task. Search for instances of '%s' and update the code accordingly.", manualFixPlaceholder)
```
colinking, line 173: Copy tweak. Though could we just print an error message when it happens, so that we can print the template?

```suggestion
		logger.Warning("Some templates used features that are not supported in tasks. Search for instances of '%s' and update the code accordingly.", manualFixPlaceholder)
```

pkg/rb2wf/converter_test.go
```diff
@@ -307,7 +323,7 @@ func TestConverter(t *testing.T) {
 									Slug:        "regex_param",
 									Name:        "regex example",
 									Type:        libapi.TypeString,
-									Constraints: libapi.Constraints{Regex: "^airplane$"},
+									Constraints: libapi.Constraints{Regex: "$airplane."},
```
colinking, line 326: probably unintended change to the regex

pkg/rb2wf/fixtures/end_to_end/test_entrypoint.ts
```diff
@@ -35,43 +35,6 @@ export default airplane.workflow(
...
-      url: "unset",
-      name: "unset",
-    };
-
-    const env = {
-      id: "unset",
-      slug: "unset",
-      name: "unset",
-      is_default: false,
-    };
```
colinking, line 74: woohoo!

colinking (approved):
🔥 🔥 🔥

PR #835, opened 2022-12-06
Update runbook converter tool to create tasks

kellymtrinh:
## Description
Update the runbook converter tool to initialize a task instead of a workflow. Includes the runtime field that's commented out by default.

## Test plan
- [x] Ran `airplane tasks init --from-workflow slug` and verified that the task was valid. I deployed the generated task and ran it.
Screenshot 2022-12-06 at 11 00 33 AM


colinking (approved):
Nice.

PR #836, opened 2022-12-06
Update runbook to task converter to set output and make from-runbook flag visible

kellymtrinh:
## Description
- Adds example of output to the end of the generated task, with a link to docs.
- Makes the `--from-runbook` flag not hidden anymore.

## Test plan
- [x] Link to doc for templates:
Screenshot 2022-12-06 at 11 46 13 AM
- [x] task output placeholder works


pkg/rb2wf/fixtures/end_to_end/test_entrypoint.ts
```diff
@@ -148,5 +148,8 @@ export default airplane.task(
         regex: "^airplane$",
       },
     });
+
+    // Set your task's output: https://docs.airplane.dev/tasks/output#set-output
+    airplane.setOutput("Task completed successfully.");
```
colinking, line 153:
```suggestion
    // Optionally return output that will be rendered in the run UI: https://docs.airplane.dev/tasks/output
    // return {}
```

colinking (approved):

PR #838, opened 2022-12-06
Add airplane init --from-runbook

kellymtrinh:
## Description
We already added `airplane tasks init --from-runbook`.
This PR adds support to do `airplane init --from-runbook`.

## Test plan
- [x] `airplane init` is unaffected
- [x] `airplane init --from-runbook` uses the runbook converter tool
```
localcli init --from-runbook test_runbook
? Where is the script for this task? test_runbook.airplane.ts
- Fetching details of runbook test_runbook
- Getting resources associated with the runbook
- Converting block note to code
- Converting block sql to code
- Converting block form1 to code
```
- [x] `airplane tasks init` is unaffected


colinking (approved):
Bless

PR #840, opened 2022-12-07
Use different field to extract api error

leeweisberger:
## Description
I noticed that I wasn't getting the body of an API error logged to me. The body is nested in `error`, not `message`.

Do you know what caused this change @colinking? Where in the code we set this? Is it this?: https://github.com/airplanedev/airport/blob/main/pkg/errorlib/errorlib.go#L128

## Test plan
I can now see API error bodies logged in the CLI


pkg/api/client.go
```diff
@@ -721,11 +721,15 @@ func (c Client) do(ctx context.Context, method, path string, payload, reply inte
 		if contentType == "application/json" {
 			var errResponse struct {
 				Message string `json:"message"`
```
colinking, line 723: Thanks for catching this! For whatever reason, the `Message` field has a different JSON tag (`error`) and I didn't think to check. You can just remove `Message` here. Can you also rename `errt.Message` to `errt.Error`?

colinking: https://airplane.sourcegraph.com/github.com/airplanedev/cli@2f1ae1c8cd1b3ea753203ff1b36563b18a39cc2e/-/blob/pkg/api/client.go?L40

leeweisberger: Yeah the name is diff because it conflicts with the `Error` function. I changed both of them to `ErrMessage` so theyre consistent.

leeweisberger: Shoot my commit did not push and merged accidentally

PR #842, opened 2022-12-07
Support uploads in Studio

calufornia:
## Description

Adds support for uploads in studio by implementing `/i/uploads/create`, `/i/uploads/get`, and `/v0/uploads/create` in the local dev server, all of which proxy to our API to handle and store upload metadata.

## Test plan
Uploads work in both tasks and views: 
image
image


pkg/server/apiext/apiext.go
```diff
@@ -70,6 +73,7 @@ func getRunIDFromToken(r *http.Request) (string, error) {
 
 // ExecuteTaskHandler handles requests to the /v0/tasks/execute endpoint
 func ExecuteTaskHandler(ctx context.Context, state *state.State, r *http.Request, req ExecuteTaskRequest) (api.RunTaskResponse, error) {
+	fmt.Printf("PARAM VALUES %+v\n", req.ParamValues)
```
geodn, line 76: Should this be removed?

pkg/params/params.go
```diff
@@ -172,3 +174,41 @@ func APIValueToInput(param libapi.Parameter, value interface{}) (string, error)
...
+	for _, param := range parameters {
+		v, ok := values[param.Slug]
+		if !ok {
+			continue
+		}
+
+		// Copy over the value
+		out[param.Slug] = v
+
+		if param.Type == libapi.TypeUpload {
```
geodn, line 194: One other thing is that we standardize dates as well in airport - see pkg/parameters/standardize/standardize.go. Should we have that logic in here as well just to prevent any date-related issues?

geodn (changes requested):

pkg/params/params_test.go
```diff
@@ -0,0 +1,55 @@
...
+			"upl1": upload,
+		},
+	}
+
+	fileParam := libapi.Parameter{
+		Name: "File param",
+		Slug: "file",
+		Type: libapi.TypeUpload,
+	}
+	stringParam := libapi.Parameter{
```
geodn, line 33: Could you add a dateParam as well?

geodn (approved):

PR #849, opened 2022-12-09
Downgrade go-git dependency

calufornia:
## Description

go-git introduced a transitive dependency on cgo, which Goreleaser does not currently support compiling with (at least our current setup). This is causing releases to fail: https://github.com/airplanedev/cli/actions/runs/3659296995/jobs/6185307857. Downgrade and anchor the dependency for now.

## Test plan
n/a


leeweisberger:
Is there a github issue tracking this that you can subscribe to?

calufornia:
> Is there a github issue tracking this that you can subscribe to?

Yep, they are fixing it here: https://github.com/go-git/go-git/pull/625

calufornia:
I think the fix will come in soon so probably don't need the renovate ignore

leeweisberger (approved):
to unblock the release. Let's roll forward once their fix is in

leeweisberger:
We're rolling back forward w/fix https://github.com/airplanedev/cli/pull/858

PR #855, opened 2022-12-14
Fix scheme determination

francescaleung:
## Description

Don't just check the host prefix, check the whole hostname.

## Test plan

Tested locally with no `--host` & with local `--host` args.


pkg/api/client.go
```diff
@@ -784,8 +784,22 @@ var httpHostPrefixes = []string{
 }
 
 func (c Client) scheme() string {
-	for _, prefix := range httpHostPrefixes {
-		if strings.HasPrefix(c.Host, prefix) {
+	if c.Host == Host {
+		return "https://"
+	}
+
+	// If the host didn't come with a scheme, force a "//" in front of it.
```
calufornia, line 791: I'm still not sure I understand why we need this

calufornia (approved):

pkg/api/client.go
francescaleung: [see here](https://go.dev/play/p/CqUTxR2cLec)

PR #862, opened 2022-12-16
Fix cloning templates on windows

leeweisberger:
## Description
We shouldn't be using filepath operations to build a URL. It's not a file so on Windows, we were using the wrong slashes.

## Test plan
Have not tested, assuming this works :)


rose-huang (approved):
what was happening on windows when cloning a template before? :0

leeweisberger:
> what was happening on windows when cloning a template before? :0

No template error every time

PR #863, opened 2022-12-16
Add getTaskReviewers support to studio

geodn:
## Description
Currently we just populate the task parameters, which is needed for task-backed forms
## Test plan
Installing cli locally and checking that task-backed forms work, both locally and with fallback env


pkg/server/apiext/apiext.go
```diff
@@ -328,6 +329,36 @@ func GetTaskMetadataHandler(ctx context.Context, state *state.State, r *http.Req
...
+			},
+		}, nil
+	}
+	if !state.UseFallbackEnv {
+		return api.GetTaskReviewersResponse{}, errors.Errorf("task with slug %s is not registered locally", taskSlug)
+	}
+
+	resp, err := state.RemoteClient.GetTaskReviewers(ctx, taskSlug)
+	if err != nil {
+		if _, ok := err.(*libapi.TaskMissingError); ok {
```
calufornia, line 353: nit: Use `errors.Is` or `errors.As` — I think we do a type assertion in `ExecuteTaskHandler` as well, that should probably use `errors.As` too...

calufornia (approved):
nice!

PR #866, opened 2022-12-20
Create airplane.yaml file for python tasks

leeweisberger:
## Description
Create an airplane.yaml file when initializing python tasks. Add python version to the airplane.yaml file.

There are a few other fixes in here, commented inline:
1. Support airplane init supplying a folder
2. Initialize package.json relative to the entrypoint rather than the cwd cc @rose-huang 

Depends on https://github.com/airplanedev/lib/pull/484

FIXES TSK-1263

## Test plan
I manually tested a whole bunch of cases


cmd/airplane/tasks/initcmd/init.go
```diff
@@ -257,10 +256,8 @@ func initWithTaskDef(ctx context.Context, cfg config) error {
 			entrypoint = cfg.file
 		}
 
-		entrypointBase := path.Base(entrypoint)
-		if strings.HasSuffix(entrypointBase, ".view.tsx") || strings.HasSuffix(entrypointBase, ".view.jsx") {
-			logger.Log("Task %s was deployed from the view file %s.", def.GetSlug(), entrypointBase)
-			logger.Log("You should run `airplane views init` if you'd like to initialize this task.")
+		if filepath.Ext(entrypoint) == "tsx" || filepath.Ext(entrypoint) == "jsx" {
```
leeweisberger, line 259: This code was a bit verbose and the log was kind of confusing.

cmd/airplane/tasks/initcmd/init.go
```diff
@@ -302,6 +299,13 @@ func initWithTaskDef(ctx context.Context, cfg config) error {
 		if err := def.SetEntrypoint(entrypoint); err != nil {
 			return err
 		}
+		absEntrypoint, err := filepath.Abs(entrypoint)
+		if err != nil {
+			return errors.Wrap(err, "determining absolute entrypoint")
+		}
+		if err := def.SetAbsoluteEntrypoint(absEntrypoint); err != nil {
+			return err
+		}
```
leeweisberger, line 308: We were never setting this. I wanted to use it further down in the code but it was unset.

cmd/airplane/tasks/initcmd/init.go
```diff
@@ -660,33 +664,37 @@ func patch(slug, file string) (ok bool, err error) {
...
-		defaultEntrypoint = slug
-		if kind == build.TaskKindNode && len(exts) > 1 {
-			// Special case JavaScript tasks and make their extensions '.ts'
-			defaultEntrypoint += ".ts"
-		} else {
-			defaultEntrypoint += exts[0]
-		}
-
-		if cwdIsHome, err := cwdIsHome(); err != nil {
+	entrypoint := defaultEntrypoint
```
leeweisberger, line 667: Part of this is refactor. The other part is supporting an entrypoint that is a folder. This fixes https://linear.app/airplane/issue/TSK-1263/cli-init-has-incorrect-default-path-when-given-directory. If the entrypoint is a folder `foo`, we create a task at `foo/foo.airplane.tsx`

cmd/airplane/tasks/initcmd/init.go
```diff
@@ -660,33 +664,37 @@ func patch(slug, file string) (ok bool, err error) {
...
+	if entrypoint == "" {
+		var err error
+		entrypoint, err = getEntrypointFile(slug, kind, cfg)
```
leeweisberger, line 670: I just refactored this out into a function

cmd/airplane/tasks/initcmd/init.go
```diff
@@ -719,6 +732,29 @@ func promptForEntrypoint(slug string, kind build.TaskKind, defaultEntrypoint str
 	return entrypoint, nil
 }
 
+func getEntrypointFile(slug string, kind build.TaskKind, cfg config) (string, error) {
+	exts := runtime.SuggestExts(kind)
```
leeweisberger, line 736: This is just refactored out with no changes

cmd/airplane/tasks/initcmd/init.go
```diff
@@ -942,14 +978,13 @@ func apiTaskToRuntimeTask(task *libapi.Task) *runtime.Task {
...
-	}
-
 	switch kind {
 	case build.TaskKindNode:
-		packageJSONDir, err := node.CreatePackageJSON(cwd, node.PackageJSONOptions{
+		entrypoint, err := def.GetAbsoluteEntrypoint()
+		if err != nil {
+			return err
+		}
+		packageJSONDir, err := node.CreatePackageJSON(filepath.Dir(entrypoint), node.PackageJSONOptions{
```
leeweisberger, line 987: This was broken imo. If you did `airplane init` and put your entrypoint in some random folder somewhere else, we would init a package.json in your cwd which may not have been a parent of the js/ts entrypoint. We should be creating the package.json relative to the entrypoint, not your current working directory cc @rose-huang

cmd/airplane/tasks/initcmd/init.go
```diff
@@ -985,6 +1018,36 @@ func runKindSpecificInstallation(ctx context.Context, cfg config, kind build.Tas
 			return err
 		}
 		return nil
+	case build.TaskKindPython:
+		if cfg.root.Flagger.Bool(ctx, logger.NewStdErrLogger(logger.StdErrLoggerOpts{}), flagsiface.AirplaneConfg) {
```
leeweisberger, line 1022: Generate an airplane.yaml file for python tasks seeded with the version of the python task or a default version

cmd/airplane/tasks/initcmd/init.go
rose-huang: Oh I think someone copied this from the views logic but the view init doesn't let you specify creating a view anywhere other than in your cwd. Good catch!

cmd/airplane/tasks/initcmd/init.go
```diff
@@ -959,19 +994,16 @@ func runKindSpecificInstallation(ctx context.Context, cfg config, kind build.Tas
...
-			if err := createOrUpdateAirplaneConfig(root, deployconfig.AirplaneConfig{
+			_, nodeVersion, buildBase, err := def.GetBuildType()
+			if err != nil {
+				return err
+			}
+			if nodeVersion == "" {
+				nodeVersion = build.DefaultNodeVersion
+			}
+
+			if err := createOrUpdateAirplaneConfig(packageJSONDir, deployconfig.AirplaneConfig{
```
leeweisberger, line 1006: This was incorrect. We should be creating the airplane config at the packageJSONDir!

ericliuche (approved):

PR #867, opened 2022-12-21
Rephrase fallback env description for airplane dev

zhan-airplane:
## Description
A user had some confusion about remote resources, and I think this help text might be contributing

## Test plan
None


cmd/airplane/tasks/dev/dev.go
```diff
@@ -135,7 +135,7 @@ func New(c *cli.Config) *cobra.Command {
 
 	cmd.AddCommand(config.New(c))
 
-	cmd.Flags().StringVar(&cfg.envSlug, "env", "", "The slug of the environment to query. Defaults to your team's default environment.")
+	cmd.Flags().StringVar(&cfg.envSlug, "env", "", "The slug of the fallback environment to query for remote resources. If not set, does not fall back to a remote environment")
```
calufornia, line 138:
```suggestion
	cmd.Flags().StringVar(&cfg.envSlug, "env", "", "The slug of the fallback environment to query for remote resources and configs. If not set, does not fall back to a remote environment")
```

calufornia: As a side note, we should really consolidate whether we use periods or not at the end of help text—I'm leaning towards no

calufornia (approved):
Thanks!

PR #868, opened 2022-12-21
Return remote run id when executing fallback task

calufornia:
## Description

We should return the remote run id when we're executing a fallback task, not the dev run id that was generated earlier.

## Test plan
Unit tests and tested with task-backed components and executing tasks with other tasks


pkg/server/apiext/apiext_test.go
```diff
@@ -100,6 +100,7 @@ func TestExecute(t *testing.T) {
 	var resp dev.LocalRun
 	err := json.Unmarshal([]byte(body.Raw()), &resp)
 	require.NoError(err)
+	// TODO: Don't check prefix
```
geodn, line 103: Why is this a TODO? Can we just remove the check?

geodn (approved):

pkg/server/apiext/apiext_test.go
calufornia: This is the only way to verify that it's a dev run currently (so we still want it in the test)—this TODO is just saying to fix the test once we have a better way

rose-huang: related to this ticket right - https://linear.app/airplane/issue/TSK-1339/dont-look-at-id-prefixes-in-studio

calufornia: Yep that's it

PR #869, opened 2022-12-21
Add goreleaser test step to GHA

kellymtrinh:
## Description
This adds a GHA step to check that goreleaser can build the projects for different architectures before merging to main.

We've run into the issue where introducing some go dependencies (usually ones with CGO) will pass all go checks but fail when trying to release the CLI. Since we don't release on every merge, this can be hard to undo when the problem is discovered. Example: https://github.com/airplanedev/cli/actions/runs/3421729247/jobs/5698258770

## Test plan
- Added unused dep `pg_query_go` which contains cgo to trigger a failure: https://github.com/airplanedev/cli/pull/869/commits/a1763dc867a7de77d7a9b8a03af29161b019b22a
- Verified there's no accidental publish step being used by goreleaser. (It's also using snapshot version)


leeweisberger (approved):
This looks good to me, but we shouldn't require this for merging as it takes way too long.

kellymtrinh:
> This looks good to me, but we shouldn't require this for merging as it takes way too long.

I'll double check with someone who's an admin for this repo

kellymtrinh:
Double checked and this shouldn't be a required check for merging.

PR #870, opened 2022-12-22
Dynamically assign studio port

calufornia:
## Description

If a `--port` option is not provided, the studio will attempt to bind to port 4000. This is problematic if another process is already bound to this port, and the user will most likely run into this if they attempt to spin up multiple studio sessions. Instead, we choose a random open port.

## Test plan
Tested by spinning up two instances of the studio, and verified that the local dev server listened on two different ports. Also verified that the `--port` flag still works.


cmd/airplane/tasks/dev/server.go
```diff
@@ -224,3 +233,16 @@ func runLocalDevServer(ctx context.Context, cfg taskDevConfig) error {
 
 	return nil
 }
+
+// findOpenPart finds an open port on the host machine, starting at the given port.
+func findOpenPort(basePort, numAttempts int) (int, error) {
+	for port := basePort; port <= basePort+numAttempts; port++ {
+		// Attempt to dial the port. If we establish a connection, it's in use.
+		conn, _ := net.DialTimeout("tcp", fmt.Sprintf(":%d", port), time.Second)
```
calufornia, line 241: We can also do this with `net.Listen`, but this causes an [inconvenient popup to occur on MacOS](https://apple.stackexchange.com/questions/393715/do-you-want-the-application-main-to-accept-incoming-network-connections-pop).

colinking: On unix systems, you should be able to do `:0` to automatically connect to a free port. Should be much more efficient.

colinking (changes requested):

cmd/airplane/tasks/dev/server.go
```diff
@@ -196,7 +205,7 @@ func runLocalDevServer(ctx context.Context, cfg taskDevConfig) error {
 	}
 
 	logger.Log("")
-	studioURL := fmt.Sprintf("%s/studio?__airplane_host=http://localhost:%d&__env=%s", appURL, cfg.port, remoteEnv.Slug)
+	studioURL := fmt.Sprintf("%s/studio?__airplane_host=http://localhost:%d&__env=%s", appURL, port, remoteEnv.Slug)
```
colinking, line 208: Unrelated, but could this have been more concise? Did we expand this just because of the SSRF issue? Could we have done `?_port=4000` (which assumes `localhost`) and remote studio could then optionally set `_host` instead? (or does remote studio even need to set a host?)

colinking (changes requested):

cmd/airplane/tasks/dev/server.go
colinking: Requesting changes just b/c I'm worried this could be slow! If that doesn't work, one idea would be to pick a less common port than `4000` to start from.

calufornia: I thought it would be nice to do what Vite did, where it incrementally tries to find a port, but I can also just have the system find one! Thought it might be nice for people to "associate" Studio with 400x ports, but those might be more popular than Vite's (5173)

cmd/airplane/tasks/dev/server.go
calufornia: We changed from `host` -> `__airplane_host` since it conflicted I think it conflicted with Views (cc @zhan-airplane for the exact issue), not related to SSRF. I think we could do `__airplane_port=4000`

colinking: Huh, what does it conflict with? I know on the task page we have arbitrary URL params from the user's parameter slugs, but that's why we have added the underscores in the past. `__port` would presumably be enough.

calufornia: From what I understand, this is to avoid conflict with a user-supplied query parameter called `host` (e.g. https://docs.airplane.dev/views/routing#reading-parameters). But yeah `_port` or `__port` should be enough to not conflict too.

zhan-airplane: ^ yup it was to avoid user parameter slug conflicts. For `__airplane_host`, I erred on the side of more verbosity to avoid potential conflicts (e.g. maybe users might also use underscores for some-reason-or-another)

zhan-airplane: I'm okay with renaming it to `__host` if you think that being concise is okay (and I guess on hindsight there's an argument with being consistent with how `__env` is named)

calufornia:
@colinking I ended up splitting up `server.Start` in two, since I didn't like how we would find a port, close the listener, only to the listen on that port again (which is also a potential race condition), so now the port allocation is implicitly part of `server.Start`. So first we call `Start` to start the dev server, and then we separately call `RegisterState` to populate its state. This means we can also trim most of the fields from `server.Options`, which was nearly identical to `State.

cmd/airplane/tasks/dev/dev.go
```diff
@@ -141,7 +142,7 @@ func New(c *cli.Config) *cobra.Command {
 	cmd.AddCommand(config.New(c))
 
 	cmd.Flags().StringVar(&cfg.envSlug, "env", "", "The slug of the fallback environment to query for remote resources and configs. If not set, does not fall back to a remote environment")
-	cmd.Flags().IntVar(&cfg.port, "port", server.DefaultPort, "The port to start the local airplane api server on - defaults to 4000.")
+	cmd.Flags().IntVar(&cfg.port, "port", 0, "The port to start the local airplane api server on - defaults to a random open port.")
```
colinking, line 145: Can you confirm that port 0 also grabs an open port on Windows?

colinking (approved):
Nice cleanup!

cmd/airplane/tasks/dev/dev.go
calufornia: Works like a charm (ignoring the logger colors not working 🙂) : 
image

PR #871, opened 2022-12-22
Fix dev config file discovery

calufornia:
## Description

If you pass in a relative development directory, the logic to auto-discover the dev config file in that directory or any of its ancestors does not work, causing the `airplane.dev.yaml` file to not get loaded. This is because `fsx.Find` expects an absolute path, since it uses [`filepath.Dir` to get the parent directory](https://github.com/airplanedev/lib/blob/c1f024b837755a990e2bbc3980731563e3e22c56/pkg/utils/fsx/fsx.go#L65) (which just returns everything except the last path element in the string). We fix this by changing the directory that we pass in to be an absolute path.

## Test plan
```
> pwd
/Users/justinlu/go/src/github.com/airplanedev/cli
> aplocal dev scripts
19:57:16 Loaded dev config from /Users/justinlu/go/src/github.com/airplanedev/cli/airplane.dev.yaml
...
```


cmd/airplane/tasks/dev/dev.go
```diff
@@ -91,12 +91,13 @@ func New(c *cli.Config) *cobra.Command {
 						devDir = filepath.Dir(cfg.fileOrDir)
 					}
 
+					absDevDir := filepath.Join(wd, devDir)
 					// Recursively search for dev config file, starting from the dev dir.
-					devConfigDir, ok := fsx.Find(devDir, conf.DefaultDevConfigFileName)
+					devConfigDir, ok := fsx.Find(absDevDir, conf.DefaultDevConfigFileName)
```
leeweisberger, line 96: Does fsx.Find say this in a comment? If not can you add please

leeweisberger (approved):

PR #875, opened 2022-12-22
Create views using inline config

leeweisberger:
## Description
Create views using inline config, behind a feature flag.

Inline views are not created in a sub-folder. I think we want to drop this behavior.

Inline views require an `airplane` dependency for doing the inline configuraiton.

## Test plan
Manually tested. It would be pretty valuable to set up unit tests for this. We'd have to find a way to mock out the filesystem.


cmd/airplane/views/initcmd/init.go
```diff
@@ -46,17 +51,24 @@ func New(c *cli.Config) *cobra.Command {
 			return Run(cmd.Root().Context(), cfg)
 		},
 	}
-	cmd.Flags().StringVar(&cfg.envSlug, "env", "", "The slug of the environment to query. Defaults to your team's default environment.")
```
leeweisberger, line 49: We weren't using this

cmd/airplane/views/initcmd/init.go
```diff
@@ -46,17 +51,24 @@ func New(c *cli.Config) *cobra.Command {
...
-func GetConfig(client *api.Client) config {
-	return config{client: client}
+func GetConfig(c *cli.Config) config {
+	return config{client: c.Client, root: c}
 }
 
 func Run(ctx context.Context, cfg config) error {
+	inlineSetByUser := cfg.cmd.Flags().Changed("inline")
+	if !inlineSetByUser {
+		if cfg.root.Flagger.Bool(ctx, logger.NewStdErrLogger(logger.StdErrLoggerOpts{}), flagsiface.DefaultInlineConfig) {
```
leeweisberger, line 68: If the user doesn't set this value, the default is set via flag

cmd/airplane/views/initcmd/scaffolding/default.view.tsx
```diff
@@ -45,13 +45,6 @@ const customersData = [
 ];
 
 const usersData = [
-  {
```
leeweisberger, line 48: Unrelated... I just think there's too much data here and it always annoys me when i create a view :( I can revert if you like it

cmd/airplane/views/initcmd/scaffolding/default_inline.airplane.tsx
```diff
@@ -0,0 +1,113 @@
...
+    label: "Role",
+    accessor: "role",
+  },
+  {
+    label: "Email",
+    accessor: "email",
+  },
+];
+
+export default airplane.view(
```
leeweisberger, line 107: Exactly the same as the other... except for this

cmd/airplane/views/initcmd/init.go
```diff
@@ -135,8 +155,10 @@ func createViewScaffolding(ctx context.Context, cfg *config) error {
 func generateEntrypointPath(cfg config, inViewDir bool) string {
 	if inViewDir {
 		return fmt.Sprintf("%s.view.tsx", cfg.slug)
-	} else {
+	} else if !cfg.inline {
 		return fmt.Sprintf("%s/%s.view.tsx", cfg.viewDir, cfg.slug)
+	} else {
+		return fmt.Sprintf("%s.airplane.tsx", cfg.slug)
```
leeweisberger, line 161: I think we want to use this new format moving forward. Both work (.view.tsx and .airplane.tsx). According to @ameyakhare we made a decision to use .airplane.tsx and I tend to like that. We can also discuss in a thread if we want

rose-huang (approved):

PR #876, opened 2022-12-23
Automatically create requirements.txt when creating python tasks

leeweisberger:
## Description
Automatically create a requirements.txt file when creating python tasks. This mirrors how we do things for JavaScript tasks.

This is necessary for python inline tasks which require the `airplane` dependency. 

Two decisions I made
1. Always generate the file, even if it is empty (not an inline task).
2. Don't run pip install. For JavaScript tasks, we run npm/yarn install to install dependencies. From my knowledge of python, most people install dependencies using virtual envs, so it's weird to run install for the users. Does this sounds good @ericliuche?

## Test plan
I manually tested this.
image

image

image


pkg/python/python.go
```diff
@@ -0,0 +1,161 @@
...
+func formatFilepath(basepath, targpath string, maxParentDirs int) (string, error) {
+	relpath, err := filepath.Rel(basepath, targpath)
+	if err != nil {
+		return "", errors.Wrap(err, "getting relative path")
+	}
+	if strings.Count(relpath, "..") > maxParentDirs {
+		return filepath.Abs(targpath)
+	} else {
+		return relpath, nil
+	}
```
ericliuche, line 160: nit: unnecessary else statement

pkg/python/python.go
```diff
@@ -0,0 +1,161 @@
...
+			}
+		}
+	} else {
+		selectedRequirementsTxtDir = directory
+		if err := createRequirementsTxtFile(selectedRequirementsTxtDir); err != nil {
+			return "", err
+		}
+	}
+
+	return selectedRequirementsTxtDir, addAllPackages(selectedRequirementsTxtDir, options.Dependencies)
```
ericliuche, line 78: it feels kind of off that we are possibly modifying the user's requirements.txt without them knowing. i feel like it should be behind some sort of confirm prompt? perhaps also since we aren't directly pip installing `airplanesdk` we should print out a snippet explaining that they'll need to install it and perhaps an example of the commmand

i see this is also how we do it for node right now

ericliuche (commented):
yeah, agreed that we shouldn't just pip install for the user

pkg/python/python.go
leeweisberger: So our options are:
1. Create/mutate the requirements.txt and add a snippet explaining what happened and how to install it (the solution in this PR plus some extra code to explain what happened and how to install).
2. Don't create/mutate the requirements.txt. Instead add a snippet informing the user that they have to install the airplane sdk.
3. Create the requirements.txt  with the airplane sdk with snippet explaining how to install (option 1). If there is already an existing requirements.txt, add a snippet informing the user they need to install the airplane sdk (option 2).

I'm pro option 1/3. For JS, I think 1 (which we do) is the correct solution. I'm less certain about python, partially b/c I'm not a python user and partially because we aren't actually installing.

Thoughts @ericliuche? I'm leaning 1. But I could be convinced to do 3.

leeweisberger: We discussed this offline. We think we should autogenerate best practices which in this case we believe to be autogenerating / mutating the requirements.txt (1). If our best practices changes (maybe we support a different system to manage python deps), we can update this to use our new best practices. Users are able to ignore our best practices and do their own thing.

pkg/python/python.go
```diff
@@ -0,0 +1,168 @@
...
+
+type PythonDependency struct {
+	Name    string
+	Version string
+}
+type RequirementsTxtOptions = struct {
+	Dependencies []PythonDependency
+}
+
+// CreateRequirementsTxt ensures there is a package.json in path with the provided dependencies installed.
```
ericliuche, line 28: nit: fix package.json

ericliuche (approved):

PR #877, opened 2022-12-23
Add sleep APIs

kellymtrinh:
## Description
Adds sleep APIs so it works in studio.
- [x] use published version of lib

## Test plan
Screenshot 2022-12-23 at 3 29 36 PM


pkg/server/apiext/sleep.go
```diff
@@ -0,0 +1,97 @@
...
+	if runID == "" {
+		return ListSleepsResponse{}, errors.New("runID is required")
+	}
+
+	run, ok := state.Runs.Get(runID)
+	if !ok {
+		return ListSleepsResponse{}, errors.New("run not found")
+	}
+
+	return ListSleepsResponse{Sleeps: run.Sleeps}, errors.New("sleep not found")
```
colinking, line 96:
```suggestion
	return ListSleepsResponse{Sleeps: run.Sleeps}, nil
```

pkg/server/apiext/sleep.go
```diff
@@ -0,0 +1,97 @@
...
+		return CreateSleepResponse{}, err
+	}
+	if runID == "" {
+		return CreateSleepResponse{}, errors.Errorf("expected runID from airplane token: %s", err.Error())
+	}
+
+	sleep := libapi.Sleep{
+		ID:        utils.GenerateID("slp"),
+		RunID:     runID,
+		CreatedAt: time.Now().UTC(),
```
colinking, line 38: Needs an `Until` and `DurationMs` fields right?

colinking (approved):

PR #881, opened 2023-01-03
Add option to start dev server on ngrok tunnel

ameyakhare:
Resolves DTS-177

## Description
We want to be able to allow users to start the dev server on an ngrok http tunnel. Users will send traffic to an ngrok host which will forward traffic to their local dev server. This adds support for browsers like Safari, which will not allow http traffic to be sent to the local dev server from an https domain (airplane.dev, where the studio UI is rendered).

## Test plan
Opened an ngrok tunnel locally and the web UI works.


cmd/airplane/tasks/dev/server.go
```diff
@@ -50,16 +54,47 @@ func runLocalDevServer(ctx context.Context, cfg taskDevConfig) error {
...
+	urlHost := ""
+	var ln net.Listener
+	if cfg.tunnel {
+		// Obtain user-specific ngrok auth token.
+		tokenResp, err := cfg.root.Client.GetTunnelToken(ctx)
+		if err != nil {
+			return errors.Wrap(err, "unable to acquire tunnel token")
+		}
+
+		tun, err := ngrok.Listen(ctx,
```
calufornia, line 67:
```suggestion
		ln, err = ngrok.Listen(ctx,
```

cmd/airplane/tasks/dev/server.go
```diff
@@ -50,16 +54,47 @@ func runLocalDevServer(ctx context.Context, cfg taskDevConfig) error {
...
+		if err != nil {
+			return errors.Wrap(err, "failed to start ngrok tunnel")
+		}
+		urlHost = tun.URL()
+		parsedURL, err := url.Parse(urlHost)
+		if err != nil {
+			return errors.Wrap(err, "parsing ngrok tunnel url")
+		}
+		devServerHost = parsedURL.Host
+		ln = tun
```
calufornia, line 80:
```suggestion
```

cmd/airplane/tasks/dev/server.go
```diff
@@ -50,16 +54,47 @@ func runLocalDevServer(ctx context.Context, cfg taskDevConfig) error {
...
-		Expose: cfg.sandbox,
+		Port:     cfg.port,
+		Expose:   cfg.sandbox,
+		Listener: ln,
 	})
 	if err != nil {
 		return errors.Wrap(err, "starting local dev server")
 	}
 
+	if !cfg.tunnel {
```
calufornia, line 92: Can just move this to the `else` case above, i.e. 

```
if cfg.tunnel {
...
} else {
  devServerHost = fmt.Sprintf("127.0.0.1:%d", port)
  urlHost = fmt.Sprintf("http://localhost:%d", port)
}
```

cmd/airplane/tasks/dev/server.go
```diff
@@ -50,16 +54,47 @@ func runLocalDevServer(ctx context.Context, cfg taskDevConfig) error {
...
+
+		tun, err := ngrok.Listen(ctx,
+			config.HTTPEndpoint(),
+			ngrok.WithAuthtoken(tokenResp.Token),
+		)
+		if err != nil {
+			return errors.Wrap(err, "failed to start ngrok tunnel")
+		}
+		urlHost = tun.URL()
+		parsedURL, err := url.Parse(urlHost)
```
calufornia, line 75: For ngrok URLs, what's the difference between the url host and the dev server host?

calufornia (approved):

cmd/airplane/tasks/dev/server.go
ameyakhare: we need the port returned from the server start call, so we can't do this

cmd/airplane/tasks/dev/server.go
ameyakhare: now that the domain is well known (not generated by ngrok), we can do this, since we dont need access to the tun object to get the constructed URL

cmd/airplane/tasks/dev/server.go
ameyakhare: the localClientDevServerHost is the host that the cli uses to access the dev server endpoints. the studioUIHost is what is exposed in the URL that we show to the user to open the studio UI

PR #885, opened 2023-01-04
Default to inline configured tasks

leeweisberger:
## Description
Behind a feature flag, default to inline configured tasks.

I also removed the option to init a code-only task (with comment at top). We still support this through our deploy pipeline, but there is no option to create one now.

## Test plan
Manually tested this works with fflag enabled


cmd/airplane/tasks/initcmd/init.go
```diff
@@ -108,36 +107,36 @@ func New(c *cli.Config) *cobra.Command {
 	cmd.Flags().StringVar(&cfg.from, "from", "", "Slug of an existing task to initialize.")
 	cmd.Flags().StringVar(&cfg.fromRunbook, "from-runbook", "", "Slug of an existing runbook to convert to a task.")
 
-	cmd.Flags().BoolVar(&cfg.codeOnly, "code-only", false, "True to skip creating a task definition file; only generates an entrypoint file.")
```
ericliuche, line 111: do you think it's worth keeping this here, but saying it's no longer supported and raising an error? no strong opinion here

cmd/airplane/tasks/initcmd/init.go
```diff
@@ -494,111 +485,6 @@ func writeDefnFile(def *definitions.Definition_0_3, cfg config) (*writeDefnFileR
 	}, nil
 }
 
-func initCodeOnly(ctx context.Context, cfg config) error {
```
ericliuche, line 497: 🫡

ericliuche (approved):

cmd/airplane/tasks/initcmd/init.go
leeweisberger: We could hide it... but I kind of want to take a more active step towards deprecating this... I haven't tested but I would be surprised if code tasks work with all the new work we've put in, especially in the studio. We don't document this anywhere, it probably doesn't work full stack, I don't see any reason why someone would use this

PR #886, opened 2023-01-05
Rebundle Vite dependencies when root dependencies change

calufornia:
## Description

Vite has some caching (and invalidation) logic to [rebundle when dependencies have changed](https://vitejs.dev/guide/dep-pre-bundling.html). However, it only tracks dependencies that are at the same level as itself, i.e. it will automatically rebundle changed dependencies in `.airplane-view` but not in the root directory. This is problematic because the majority of dependencies are in the latter (e.g. `@airplane/views`). We force vite to rebundle dependencies by tracking the hash of the root lockfile.

Furthermore, the user is no longer required to restart the Studio to pick up on dependency changes—they can simply refresh the page, and we'll automatically restart the vite process for them (while still maintaining the vite process caching that we had previously so view loading is still quick). 

cc @geodn 

## Test plan
Tested upgrading `@airplane/views` from `1.11.1` -> `1.13.1` and verified that the new package was being used.


stepzhou (approved):
🚢 

Should probably verify that this also works on Windows

stepzhou:
Now that I think about it more, I think it's slightly more correct to hash the lockfiles like Vite does since the node_modules/lockfiles can sometimes change without the package.json changing

calufornia:
> Now that I think about it more, I think it's slightly more correct to hash the lockfiles like Vite does since the node_modules/lockfiles can sometimes change without the package.json changing

Updated to hash the lockfile instead

PR #889, opened 2023-01-05
Support env flag in runbook converter tool

kellymtrinh:
## Description
Updates the runbook converter tool to respect the `--env ` flag.

## Test plan
I created a runbook that only existed in the stage environment, and now this command works:
```
localcli tasks init --from-runbook test_runbook_stage_only --env stage
? Where is the script for this task? test_runbook_stage_only.airplane.ts
- Fetching details of runbook test_runbook_stage_only in env stage
- Getting resources associated with the runbook
- Converting block note to code
- Creating test_runbook_stage_only.airplane.ts
- Formatting code via prettier

✅ To complete your task:
- Write your task logic in test_runbook_stage_only.airplane.ts

🛫 To deploy your task to Airplane:
  airplane deploy test_runbook_stage_only.airplane.ts
```

Without this change, it would 404 because it was defaulting to the `prod` environment, where it doesn't exist:
```
localcli tasks init --from-runbook test_runbook_stage_only --env stage
? Where is the script for this task? test_runbook_stage_only.airplane.ts
- Fetching details of runbook test_runbook_stage_only

Error: Api: 404 - No runbook found

exit status 1
```


pkg/api/client.go
```diff
@@ -375,8 +375,8 @@ func (c Client) GetOutputs(ctx context.Context, runID string) (res GetOutputsRes
 }
 
 // GetRunbook returns the details of a runbook by slug.
-func (c Client) GetRunbook(ctx context.Context, runbookSlug string) (res GetRunbookResponse, err error) {
-	q := url.Values{"runbookSlug": []string{runbookSlug}}
+func (c Client) GetRunbook(ctx context.Context, runbookSlug string, envslug string) (res GetRunbookResponse, err error) {
+	q := url.Values{"runbookSlug": []string{runbookSlug}, "envSlug": []string{envslug}}
```
yolken-airplane, line 379: If `envSlug` is an empty string, this should still work, right? Just wanted to check!

yolken-airplane (approved):

pkg/api/client.go
kellymtrinh: Yep, still passes in nothing and ends up being the default env!

yolken-airplane: Cool

PR #892, opened 2023-01-09
Adopt core API client from lib

colinking:
## Description

Adopts the shared core API client logic from `lib`. By doing so, the CLI client will now send `Idempotency-Key` headers and perform more gradual retries.

Depends on: https://github.com/airplanedev/lib/pull/501

Resolves TSK-936

## Test plan

Tested the CLI locally by performing a deploy. This tests a good cross section of API requests.


pkg/api/client.go
```diff
@@ -78,6 +50,50 @@ type Client struct {
...
+	}
+	if opts.Source != "" {
+		headers["X-Airplane-Client-Source"] = opts.Source
+	}
+
+	return Client{
+		ClientOpts: opts,
+		http: libhttp.NewClient(libhttp.ClientOpts{
+			Headers: headers,
+			RequestLogHook: func(req *http.Request, attempt int) {
```
colinking, line 66: I added LogHook's to `libhttp.Client` so that we could retain the existing debug logging in CLI.

yolken-airplane (approved):

pkg/api/client.go
```diff
@@ -78,6 +50,54 @@ type Client struct {
...
+	}
+
+	return Client{
+		ClientOpts: opts,
+		http: libhttp.NewClient(libhttp.ClientOpts{
+			Headers:   headers,
+			UserAgent: "airplane/cli/" + version.Get(),
+			// Temporarily bump the default timeout to 30s.
+			// TODO: revert to the default 10s after optimizing the long-tail of slow API endpoints.
+			Timeout: 30 * time.Second,
```
colinking, line 69: FYI: I'm going to merge this change. The API client has baked for long enough that I'm more confident this change will not cause issues. The primary issues we've seen are related to a few long-tail API endpoints that are slow, so I've bumped the default timeout here as a pre-caution. I'll come back and revert this later.

PR #893, opened 2023-01-09
Add update files endpoint

calufornia:
## Description

Adds an endpoint `/dev/files/update` to the local dev server that will be used by the Studio to save files. For now, we just send over the contents of the file in the request, but in the future we could use SSE, WS, etc. to make this more efficient. Also adds a directory traversal check to `/dev/files/get`.

## Test plan
Tested in conjunction with web changes


pkg/server/apidev/apidev.go
```diff
@@ -288,6 +293,30 @@ func GetFileHandler(ctx context.Context, state *state.State, r *http.Request) (G
...
+	// Ensure the path is within the dev server root.
+	if !strings.HasPrefix(req.Path, s.Dir) {
+		return struct{}{}, errors.New("path is outside dev root")
+	}
+
+	if strings.Contains(req.Path, "..") {
+		return struct{}{}, errors.New("path may not contain directory traversal elements (`..`)")
+	}
+
+	if err := os.WriteFile(req.Path, []byte(req.Content), 0644); err != nil {
```
ameyakhare, line 313: we don't validate that the path exists + is in the actual airplane directory the dev server is started from. I would add that validation here

ameyakhare (changes requested):
one small thing

pkg/server/apidev/apidev.go
ameyakhare: or if its new, that its just in the dev server folder. worried about some bug overwriting a system file lol

calufornia: that's the validation on lines 305-311, right?

ameyakhare (approved):

PR #894, opened 2023-01-09
Install dependencies on init

francescaleung:
## Description

Runs discovery over bundle directories, queries for build instructions per bundle, applies build instructions (somewhat). Discoverer instantiation is pulled from the deploy code & could probably get deduped.

Resolves PDS-1370. Depends on https://github.com/airplanedev/lib/pull/503.

## Test plan

Tested locally, but minimally.


cmd/airplane/root/initcmd/download.go
```diff
@@ -35,7 +44,122 @@ func initializeCodeWorkspace(ctx context.Context, cfg config) error {
...
+			if envVar.Value != nil {
+				cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", key, *envVar.Value))
+			}
+		}
+		output, err := cmd.CombinedOutput()
+		if err != nil {
+			fmt.Println(string(output))
+			return errors.Wrapf(err, "error running build")
+		}
+		if err := os.Remove(path.Join(bundle.RootPath, "airplane_build.sh")); err != nil {
```
calufornia, line 158: nit: Probably should defer after `os.WriteFile` instead of here

cmd/airplane/root/initcmd/download.go
```diff
@@ -35,7 +44,122 @@ func initializeCodeWorkspace(ctx context.Context, cfg config) error {
...
+			BuildContext: bundle.BuildContext,
+		})
+		if err != nil {
+			if _, ok := errors.Cause(err).(build.ErrUnsupportedBuilder); ok {
+				continue
+			}
+			return err
+		}
+
+		var b strings.Builder
```
calufornia, line 102: Could you extract this into a separate function? Something like `generateBuildScript`?

cmd/airplane/root/initcmd/download.go
```diff
@@ -35,7 +44,122 @@ func initializeCodeWorkspace(ctx context.Context, cfg config) error {
...
+		b.WriteString("#!/bin/sh\n")
+		b.WriteString("set -xeo pipefail\n")
+
+		for _, inst := range instructions.InstallInstructions {
+			if inst.SrcPath != "" && inst.DstPath != "" && inst.DstPath != "." {
+				// Weed out things that don't exist
```
calufornia, line 108: Can you expand on this comment? i.e. mention why it's ok to ignore non-existent sources

calufornia (approved):

cmd/airplane/root/initcmd/download.go
francescaleung: extracted to pkg/dev

PR #896, opened 2023-01-09
For inline tasks, add option to print to std out

leeweisberger:
## Description
See https://airplanedev.slack.com/archives/C038E9488F5/p1673301480047359 for context.

The user flow I'm trying to fix here is: I already have `my_task.airplane.ts` on my machine. I go to the UI, and change the description of `my_task.airplane.ts`. To sync that back to my machine, I run `airplane init --from my_task my_task.airplane.ts`. I want to get the configuration, but I don't necessarily want to overwrite my entire file.

This flow is not super important and subject to change, so I'm trying to move passed this relatively quickly. In the future, here's what I think this should look like:
- I go to my studio. There is a wysiwyg editor that lets me change the description of my task.
- Changing this description automatically changes my configuration file.

Until we get there ^ I don't think we should invest in the semi-complex parsing required to do this correctly!

## Test plan
Manually tested

image


calufornia (commented):
I'm a little confused by the test plan: why is the user still prompted for the task's script? How did `bb.ts` turn into `bb.airplane.ts`?

leeweisberger:
> I'm a little confused by the test plan: why is the user still prompted for the task's script? How did `bb.ts` turn into `bb.airplane.ts`?

bb.ts -> bb.airplane.ts. That's logic we already had. We automatically add an airplane extension b/c it has to have that to be a task.

Prompting for the script. This we also already had. You are cloning from a task by slug. You still need to tell us what file to clone it into. It defaults to the slug name, so in this example `inline_node.airplane.ts`. I just picked a different name.

cmd/airplane/tasks/initcmd/init.go
```diff
@@ -359,6 +356,52 @@ func initWithTaskDef(ctx context.Context, cfg config) error {
...
+			overwriteOption = fmt.Sprintf("Overwrite %s with configuration from %s.", entrypoint, cfg.from)
+		}
+		shouldPrintToStdOutOption := "Print to stdout instead of writing to a file."
+		if cfg.from != "" {
+			shouldPrintToStdOutOption = fmt.Sprintf("Print %s to stdout instead of writing to a file.", cfg.from)
+		}
+		options := []string{
+			overwriteOption,
+			shouldPrintToStdOutOption,
+			"Choose a new file.",
```
calufornia, line 372: More clear: `Write the configuration to a new file`/`Write the configuration for %s to a new file`

calufornia (approved):
LGTM

PR #897, opened 2023-01-11
Add skip sleep endpoint

kellymtrinh:
## Description
Add endpoint to skip sleep locally.

Related to: https://github.com/airplanedev/js-sdk/pull/255

Order of operations:
- release CLI with skip sleep
- release web with skip button and SDK

## Test plan
- [x] Tested with UI to skip sleep. The run finishes immediately after if there's nothing else, or proceeds with the rest of the logic.


pkg/server/apiint/apiint_test.go
```diff
@@ -199,6 +199,91 @@ func TestSubmitPrompts(t *testing.T) {
 	require.Nil(listPrompts.Prompts[1].SubmittedBy)
 }
 
+func TestSkipSleeps(t *testing.T) {
```
colinking, line 202: nice!

colinking (approved):

PR #898, opened 2023-01-11
Don't use worktree status

leeweisberger:
## Description
Don't use worktree.status as it is super slow. Even in our small internal tools repo it takes >2 seconds. Instead shell out to git status.

https://github.com/go-git/go-git/issues/181

## Test plan
Manually tested this with both dirty and non-dirty repositories


cmd/airplane/root/deploy/git.go
```diff
@@ -66,11 +70,19 @@ func GetGitMetadata(repo *git.Repository) (api.GitMetadata, error) {
...
+	// https://github.com/go-git/go-git/issues/181
+	cmd := exec.Command("git", "status", "--porcelain")
+	cmd.Dir = w.Filesystem.Root()
+	var stdout bytes.Buffer
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stdout
+	err = cmd.Run()
 	if err != nil {
-		return meta, err
+		analytics.ReportMessage(fmt.Sprintf("failed to get git status: %v", string(stdout.Bytes())))
```
ameyakhare, line 82:
```suggestion
		analytics.ReportMessage(fmt.Sprintf("failed to get git status: %v", string(stdout.String())))
```

ameyakhare (approved):
need to fix tests, but otherwise looks fine

cmd/airplane/root/deploy/git.go
```diff
@@ -66,11 +70,19 @@ func GetGitMetadata(repo *git.Repository) (api.GitMetadata, error) {
...
 
-	status, err := w.Status()
+	// Worktree.Status() is very slow so fall back to the command line instead.
+	// https://github.com/go-git/go-git/issues/181
+	cmd := exec.Command("git", "status", "--porcelain")
+	cmd.Dir = w.Filesystem.Root()
+	var stdout bytes.Buffer
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stdout
+	err = cmd.Run()
```
ameyakhare, line 80: no need for temp var, can just do `if cmd.Run() != nil`

PR #902, opened 2023-01-13
Only create tsconfig for typescript tasks

leeweisberger:
## Description

FIXES PDS-1401

Don't create tsconfig for js tasks/views

## Test plan
Manually tested both ts and js tasks and views


cmd/airplane/views/initcmd/init.go
```diff
@@ -141,9 +142,11 @@ func createViewScaffolding(ctx context.Context, cfg *config) error {
 	if err != nil {
 		return err
 	}
-	// Create/update tsconfig in the same directory as the package.json file
-	if err := node.CreateViewTSConfig(packageJSONDir); err != nil {
-		return err
+	if filepath.Ext(entrypoint) == ".tsx" {
```
rose-huang, line 145: currently this is still always creating tsconfig on views init right? because path always ends in tsx? are we just adding this check in case we change the path in the future?

rose-huang (approved):

cmd/airplane/views/initcmd/init.go
leeweisberger: Yeah that was the idea.

PR #904, opened 2023-01-17
Incrementally find open studio port

calufornia:
## Description

Similar to Vite, we attempt to incrementally find a port for the local dev server starting at 4000, instead of getting a random open port. This is because users tend to refresh the studio page when restarting the studio, which is not possible if the port changes every time, and forces users to use the `--port` flag.

## Test plan

Tested `airplane dev` in two different tabs—the first dev server was bound to 4000 and the second one was bound to 4001. Also verified that the `--port` flag still worked.


pkg/server/server.go
```diff
@@ -105,6 +106,14 @@ func newServer(router *mux.Router, state *state.State, port int, expose bool) (*
 		Handler: router,
 	}
 
+	var err error
+	if port == 0 {
+		port, err = findOpenPort(defaultPort, 100)
```
ericliuche, line 111: nit: declare `var err error` inside if block

pkg/server/server.go
```diff
@@ -118,6 +127,19 @@ func newServer(router *mux.Router, state *state.State, port int, expose bool) (*
 	}, nil
 }
 
+// findOpenPart finds an open port on the host machine, starting at the given port.
+func findOpenPort(basePort, numAttempts int) (int, error) {
+	for port := basePort; port <= basePort+numAttempts; port++ {
+		// Attempt to dial the port. If we establish a connection, it's in use.
+		conn, _ := net.DialTimeout("tcp", fmt.Sprintf(":%d", port), time.Second)
```
ericliuche, line 134: add a comment explaining that we can't use `Listen` since it opens a popup on every try

ericliuche (approved):

PR #905, opened 2023-01-18
Add envVars to dev config file

calufornia:
## Description

Adds an `envVars` field to the dev config file that is used to mimic the behavior of the (deprecated) airplane.env file. That is, the user can utilize this field to set env vars that will be used purely for local development, without needing to convert an "env var from value" into an "env var from config" and setting a local config. The dev config file's environment variables will override any env vars set on the task definition itself.

For now, we maintain backwards compatibility with dotenv files, and so the order or precedence when it comes to reading  in environment variables is `Dev config file` > `airplane.env` > `Task definition`.

Also cleans up some dev config file method names/tests/comments.

## Test plan
Tested loading in environment variables using the dev config file.


cmd/airplane/tasks/dev/config/delete_envvar/delete_envvar.go
```diff
@@ -0,0 +1,38 @@
...
+	key    string
+}
+
+func New(c *cli.DevCLI) *cobra.Command {
+	var cfg = config{devCLI: c}
+	cmd := &cobra.Command{
+		Use:   "delete-envvar",
+		Short: "Deletes an environment variable from the dev config file",
+		Example: heredoc.Doc(`
+			airplane dev config delete-envvar <key>
```
colinking, line 22: Should we do `airplane dev config envvars [set|delete] ...`?

pkg/conf/conf_test.go
```diff
@@ -85,88 +85,138 @@ func TestUserConfig(t *testing.T) {
 	})
 }
 
-func TestDevConfig(t *testing.T) {
-	t.Run("read missing", func(t *testing.T) {
-		var assert = require.New(t)
-		var dir = tempdir(t)
-		var path = filepath.Join(dir, "dev.yaml")
+func TestDevConfigMissing(t *testing.T) {
+	var assert = require.New(t)
```
colinking, line 89: Minor, but we should call this `require`. `assert` is a different package from `testify` that does not fail on errors and we should never use it because it leads to confusing errors.

pkg/conf/dev.go
```diff
@@ -189,21 +190,59 @@ func (d *DevConfig) RemoveConfigVar(key string) error {
...
+	if err := writeDevConfig(d); err != nil {
+		return err
+	}
+
+	logger.Log("Successfully wrote environment variable %q to dev config file.", key)
+	return nil
+}
+
+// RemoveEnvVar deletes an env var from the dev config file, if it exists.
+func (d *DevConfig) RemoveEnvVar(key string) error {
```
colinking, line 208: nit - match the command name

```suggestion
func (d *DevConfig) DeleteEnvVar(key string) error {
```

colinking (approved):

calufornia:
> > The dev config file's environment variables will override any env vars set on the task definition itself.
> 
> My intuition is that task-specific env vars are more granular than dev-config env vars, so I'd expect the former to take precedence. wdyt?

I've been thinking about this throughout the day and flip-flopping on it. I agree with your intuition—OTOH, I'm thinking about the scenario where they have a task with an env var from value, and they want to override it. If the dev config file env vars don't override it, then they'd need to either:
1. Change the task env var but revert it before deploying/committing, _or_ 
2. Convert the env var from value to an env var from config variable and then add configs in local/prod

Both of these are a bit suboptimal, and I think overriding a task env var will be a common use case for envVars (and more importantly, more common than the reverse), so I'm inclined to keep it like this for now...

cmd/airplane/tasks/dev/config/delete_envvar/delete_envvar.go
calufornia: I thought about this early on, but the number of nested subcommands seemed a bit too much (4!)

pkg/conf/conf_test.go
calufornia: Fixed all instances in the file!

PR #910, opened 2023-01-20
Add endpoint to zip and return dev root

calufornia:
## Description

Adds an endpoint, `/dev/files/downloadBundle`, to the local dev server that zips the user's development workspace and returns it in the response. The archiving is done in-memory, but since we'll only be doing this for templates (and filtering out `node_modules/`, etc.), this should be ok.

## Test plan
Tested in conjunction with web changes.


pkg/server/apidev/apidev.go
```diff
@@ -478,3 +482,12 @@ func GetTaskErrorsHandler(ctx context.Context, state *state.State, r *http.Reque
 	}
 	return GetTaskErrorResponse{Info: info, Errors: errors, Warnings: warnings}, nil
 }
+
+func DownloadBundleHandler(ctx context.Context, state *state.State, r *http.Request) ([]byte, string, error) {
+	buf := new(bytes.Buffer)
+	if err := utils.Zip(buf, state.Dir, ignoredDirs); err != nil {
```
francescaleung, line 488: Is there still an `.airplaneignore` file? Do we want this to be a static list of directories?

francescaleung (approved):

pkg/server/apidev/apidev.go
calufornia: That's a good idea—I guess this should just be the same logic that we use for ignoring files/directories when packaging we archive bundles

PR #911, opened 2023-01-20
Fix error checking

calufornia:
## Description

We were using `errors.As` wrong—the second argument should be a pointer to an empty error struct, _not_ and nil pointer. However, we don't use the resulting error anyway, and so we can perform a type assertion.

## Test plan
Verified that a view using a non-existent task slug properly shows `task with slug %s is not registered locally or remotely` instead of causing the CLI to panic.


leeweisberger (approved):

pkg/server/apiext/apiext.go
```diff
@@ -262,8 +262,7 @@ func ExecuteTaskHandler(ctx context.Context, state *state.State, r *http.Request
 			EnvSlug:     state.RemoteEnv.Slug,
 		})
 		if err != nil {
-			var taskMissingError *libapi.TaskMissingError
-			if errors.As(err, taskMissingError) {
+			if _, ok := err.(*libapi.TaskMissingError); ok {
```
colinking, line 265: We should use `errors.As` (or `errors.Is`) everywhere, otherwise this code will stop working if we start wrapping the underlying error. https://pkg.go.dev/errors

You just need a pointer to an Error type, so if the `Error()` method is declared on `*libapi.TaskMissingError`, then you would use a double pointer here.

```
var taskMissingError *libapi.TaskMissingError
if errors.As(err, &taskMissingError) {
```

kellymtrinh: :/ I feel like this bug is really hard to catch (the issue is it's 
```
var taskMissingError *libapi.TaskMissingError
    if errors.As(err, taskMissingError) {
```

but it should be

```
var taskMissingError libapi.TaskMissingError
    if errors.As(err, &taskMissingError) {
```

colinking: @kellymtrinh those are equivalent! But I agree it's easy to mis-use `errors.Is/As` :(

Though looking at the code, @calufornia you shouldn't need a double pointer here? From my comment above ("so if the Error() method is declared on *libapi.TaskMissingError"), I was just too lazy to check, but it's not:

```
func (err TaskMissingError) Error() string {
```

So you want to pass a `*libapi.TaskMissingError` in to `errors.As`.

FWIW, there's no reason for us to be returning a `&api.TaskMissingError{...}` instead of `api.TaskMissingError`.

calufornia: > FWIW, there's no reason for us to be returning a &api.TaskMissingError{...} instead of api.TaskMissingError.

In client.go, right? I'd prefer to return the struct directly, not a reference to it. The double pointer logic here was just to support the fact that it is returning a pointer—I thought that maybe there was a reason we were returning a pointer in the client? But if there isn't, then we should remove the pointer.

colinking: Pretty sure there's no reason. I imagine that comment is just trying to document the behavior.

PR #914, opened 2023-01-23
Add optional auth middleware to dev server

ameyakhare:
## Description
We need to add auth middleware to the dev server for when the server is opened on a public ngrok tunnel. The middleware is optionally installed and checks for the presence of the correct token in the X-Airplane-Dev-Token header. I couldn't repurpose X-Airplane-Token since we use that to pass optional fake unverified JWTs to tasks.

## Test plan
Started studio with the --tunnel option and can't access the dev server publicly.


cmd/airplane/tasks/dev/server.go
```diff
@@ -64,9 +65,13 @@ func runLocalDevServer(ctx context.Context, cfg taskDevConfig) error {
 			return errors.Wrap(err, "unable to acquire tunnel token")
 		}
 
+		randString := utils.RandomString(20, utils.CharsetLowercaseNumeric)
```
calufornia, line 68: Use `crypto/rand` instead, which is cryptographically secure:
```
b := make([]byte, 40)
_, err := rand.Read(b)
...
```

pkg/server/server.go
```diff
@@ -83,8 +87,20 @@ func NewRouter(state *state.State) *mux.Router {
 			"x-airplane-client-version",
 			"x-airplane-client-source",
 			"idempotency-key",
+			"x-airplane-dev-token",
 		}),
 	))
+	if opts.Token != nil {
+		r.Use(func(next http.Handler) http.Handler {
+			return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				if r.Header.Get("X-Airplane-Dev-Token") != *opts.Token {
```
calufornia, line 96: 🥇

calufornia (approved):

cmd/airplane/tasks/dev/server.go
ameyakhare: this can generate random ascii characters that are not allowed in URLs so im going to leave it as alphanumeric. I will add another charset which includes uppercase characters as well to increase entropy

PR #917, opened 2023-01-23
Add endpoint for dependency reinstallation + install dependencies on sandbox startup

francescaleung:
## Description

This patch moves dependency installation to dev sandbox startup (instead of at the template cloning stage, which was the original plan). Partially this is because dependency installation can take a while (~10s for selenium), & it didn't make sense for it to be on the critical path for studio startup. This only happens in sandbox mode.

The `/dev/info` endpoint now also includes some sandbox-specific information, so the client can have some basic idea of the state of dependency management. Reinstallation of dependencies happens in an asynchronous mode because it's potentially a long operation; the `/dev/dependencies/reinstall` endpoint only triggers a reinstallation to happen.

Resolves DTS-112, DTS-165.

## Test plan

Tested locally.


cmd/airplane/tasks/dev/server.go
```diff
@@ -235,6 +244,9 @@ func runLocalDevServer(ctx context.Context, cfg taskDevConfig) error {
 				logger.Log("Something went wrong. Try running the command with the --debug flag for more details.")
 			}
 		}()
+	} else {
+		go func() {
```
calufornia, line 248: What is this for?

pkg/dev/dependencies.go
```diff
@@ -1,19 +1,65 @@
...
+	"github.com/airplanedev/cli/pkg/api"
+	"github.com/airplanedev/cli/pkg/logger"
 	"github.com/airplanedev/lib/pkg/build"
 	"github.com/airplanedev/lib/pkg/deploy/bundlediscover"
+	"github.com/airplanedev/lib/pkg/deploy/discover"
 	"github.com/flynn/go-shlex"
 	"github.com/pkg/errors"
 )
 
+func BundleDiscoverer(client api.APIClient, l logger.Logger, envSlug string) *bundlediscover.Discoverer {
```
calufornia, line 21: Since this is used by both dev and deploy, I would move it to a different package. Maybe `discover` or something along those lines

pkg/dev/dependencies.go
```diff
@@ -66,13 +112,17 @@ func InstallBundleDependencies(bundle bundlediscover.Bundle) error {
...
+	}
+	defer os.RemoveAll(tmpdir)
+
+	if err := os.WriteFile(path.Join(tmpdir, "airplane_build.sh"), []byte(b.String()), 0777); err != nil {
 		return err
 	}
-	defer os.Remove(path.Join(bundle.RootPath, "airplane_build.sh"))
 
-	cmd := exec.Command("./airplane_build.sh")
+	cmd := exec.Command(path.Join(tmpdir, "/airplane_build.sh"))
```
calufornia, line 125: nit: Just set `cmd.Dir` to be `tmpdir`?

calufornia (approved):

pkg/dev/dependencies.go
francescaleung: The build needs to run in the workspace directory; `npm install` won't help in `tmpdir`. This is just moving it out of the workspace directory so you don't get `airplane_build.sh` in the explorer.

cmd/airplane/tasks/dev/server.go
francescaleung: oops

pkg/server/apidev/apidev.go
```diff
@@ -479,6 +502,28 @@ func GetTaskErrorsHandler(ctx context.Context, state *state.State, r *http.Reque
...
+
+type ReinstallDependenciesStatus string
+
+const (
+	ReinstallDependenciesStatusDone       ReinstallDependenciesStatus = "done"
+	ReinstallDependenciesStatusInProgress ReinstallDependenciesStatus = "in progress"
+)
+
+func ReinstallDependenciesHandler(ctx context.Context, s *state.State, r *http.Request) (ReinstallDependenciesResponse, error) {
+	inProgress := s.SandboxState.RebuildWithTimeout(ctx, s.BundleDiscoverer, s.Dir)
```
calufornia, line 517: I'm not sure I understand why we have a timeout now—don't we want to just kick off a dependency reinstallation, return right away, then poll `/dev/info`?

francescaleung: Sometimes dependency installation is quick. So quick, in fact, that you may poll the server (reply: not rebuilding), kick off a reinstallation, & have that reinstallation finish before you poll the server again (reply: not rebuilding). On the client-side, that means you indicate that the server's rebuilding when you kick off the reinstallation, but you never change the status back to not rebuilding, because the results of the server poll never change.

PR #918, opened 2023-01-23
Convert error type assertions to errors.As calls

calufornia:
## Description

Using `errors.As` is preferable to type asserting errors in case the underlying error is wrapped. Converts all type assertions in the repo to use `errors.As` instead.

## Test plan
Verified that errors are appropriately unwrapped.


colinking (approved):
bless

PR #919, opened 2023-01-24
Add postcss config to .airplane-view if a view has a tailwind config

geodn:
## Description
Adds a postcss config to the .airplane-view dir if tailwind is detected in the root dir. This postcss config references the aforementioned tailwind config.

This is needed as part 1 of 2 of supporting tailwind in views, where part 1 is dev and part 2 is deploy.

## Test plan
`go install ./cmd/airplane`
`airplane dev` with a view containing tailwind.config.js
verify that tailwind styles are applied


pkg/views/dev.go
```diff
@@ -120,6 +120,23 @@ func Dev(ctx context.Context, v viewdir.ViewDirectoryInterface, viteOpts ViteOpt
 		return nil, "", nil, errors.Wrap(err, "closing new package.json file")
 	}
 
+	// Add postcss config if tailwind config is detected.
+	tailwindConfig := filepath.Join(root, "tailwind.config.js")
+	if _, err := os.Stat(tailwindConfig); err == nil {
+		postcssConfigStr := `module.exports = {
```
calufornia, line 126: Can you put this in an embedded file `postcss.config.js` instead?

calufornia (changes requested):

pkg/views/dev.go
geodn: With lib v0.5.144, this string is exported by libbuild - updated to use that

calufornia (approved):

PR #922, opened 2023-01-26
Append dev token header for dev server self-requests

calufornia:
## Description

There are some instances where the dev server needs to make a request to itself (i.e. anytime `LocalClient`) is used. An example is the [`/i/runs/get` endpoint](https://github.com/airplanedev/cli/blob/be0e03b13bbd8b8b7c98274b7fc77fab6208bfc9/pkg/server/apiint/apiint.go#L508), which calls `UpdateTaskRequest` which eventually issues a request to another dev server endpoint. We need to ensure that the tunnel token is included in the header for any local client requests, and so we some logic to the API client to do so.

## Test plan
Verified that the token is properly included in headers for intra-dev server requests


ameyakhare (approved):
can you confirm that this wouldnt be necessary for sandbox routing since the localclient will use localhost (without auth), even though the client is interacting with the server via the fly router?

calufornia:
> can you confirm that this wouldnt be necessary for sandbox routing since the localclient will use localhost (without auth), even though the client is interacting with the server via the fly router?

correct, this is not necessary for sandbox because the dev server does not auth sandbox requests—the sandbox router does.

PR #928, opened 2023-01-26
Return remote run and outputs for apiint endpoints

rose-huang:
## Description
For the debug panel, we need the `/i/runs/get?runID=` and `/i/runs/getOutputs?runID=` endpoints to work. 

note: `/i/runs/get?runID=` still returns slightly diff info in studio than in our prod api so we are going to handle that in the web code

## Test plan
with changes in web:

https://user-images.githubusercontent.com/34196250/214923024-13049f75-b052-44c1-b9cb-e3de52a58680.mov


pkg/api/types.go
```diff
@@ -264,21 +264,22 @@ func (r RunStatus) IsTerminal() bool {
...
-	CancelledAt *time.Time `json:"cancelledAt"`
-	CancelledBy *string    `json:"cancelledBy"`
-	EnvSlug     string     `json:"envSlug"`
+	RunID       string             `json:"runID"`
+	TaskID      string             `json:"taskID"`
+	TaskName    string             `json:"taskName"`
+	TeamID      string             `json:"teamID"`
+	Status      RunStatus          `json:"status"`
+	ParamValues Values             `json:"paramValues"`
+	Parameters  *libapi.Parameters `json:"parameters"`
```
rose-huang, line 273: i just added the parameters field bc the web code wants it to show the parameter pills

pkg/print/types.go
```diff
@@ -49,21 +49,22 @@ func printTasks(tasks []libapi.Task) []printTask {
...
-	CancelledAt *time.Time    `json:"cancelledAt" yaml:"cancelledAt"`
-	CancelledBy *string       `json:"cancelledBy" yaml:"cancelledBy"`
-	EnvSlug     string        `json:"envSlug" yaml:"envSlug"`
+	RunID       string             `json:"id" yaml:"id"`
+	TaskID      string             `json:"taskID" yaml:"taskID"`
+	TaskName    string             `json:"taskName" yaml:"taskName"`
+	TeamID      string             `json:"teamID" yaml:"teamID"`
+	Status      api.RunStatus      `json:"status" yaml:"status"`
+	ParamValues api.Values         `json:"paramValues" yaml:"paramValues"`
+	Parameters  *libapi.Parameters `json:"parameters" yaml:"parameters"`
```
rose-huang, line 58: same also just adding parameters

pkg/server/apiint/apiint.go
```diff
@@ -552,6 +586,31 @@ func ListRunsHandler(ctx context.Context, state *state.State, r *http.Request) (
 	}, nil
 }
 
+type GetOutputsResponse struct {
+	Output api.Outputs `json:"output"`
+}
+
+func GetOutputsHandler(ctx context.Context, state *state.State, r *http.Request) (GetOutputsResponse, error) {
```
calufornia, line 593: We have identical code in `apiext.go`—can you extract this code into a separate file (maybe into `/server/outputs.go`) and just use that function here?

pkg/server/apiint/apiint.go
```diff
@@ -498,14 +499,47 @@ type GetRunResponse struct {
 
 func GetRunHandler(ctx context.Context, state *state.State, r *http.Request) (GetRunResponse, error) {
```
calufornia, line 500: Can we modify the `/i/runs/get` endpoint to also return a `taskRevision` field (in addition to the `run` and `task` fields it currently returns)? I understand that this results in duplicate information in the response for now, but we can eventually remove the `task` field from the response so that it's just `run` and `taskRevision` and fully matches our prod API. We should be able to populate the TaskRevision struct from the Task struct, except for the task revision's ID and createdAt, but I don't think we're looking at those fields anyway so they can be generated on-the-fly. I think this will make the web changes a lot cleaner.

calufornia (changes requested):

pkg/server/apiint/apiint.go
rose-huang: AD I left this as just run and task after adding task to the prod endpoint in https://github.com/airplanedev/airport/pull/3974/files

pkg/server/apiint/apiint.go
```diff
@@ -498,14 +500,47 @@ type GetRunResponse struct {
...
+			SucceededAt: remoteRun.SucceededAt,
+			FailedAt:    remoteRun.FailedAt,
+			ParamValues: remoteRun.ParamValues,
+			Parameters:  remoteRun.Parameters,
+			TaskID:      remoteRun.TaskID,
+			TaskName:    remoteRun.TaskName,
+			Remote:      true,
+		}
+		response := GetRunResponse{Run: run}
+		task, err := state.RemoteClient.GetTaskByID(ctx, remoteRun.TaskID)
```
calufornia, line 531: `GetRun` should return the task now, right? Can we just modify `GetRunResponse` in `client.go`?

calufornia (approved):

pkg/server/apiint/apiint.go
rose-huang: @calufornia GetRun is hitting the external api endpoint which doesn't return the task rn 😅

calufornia: Ah right...maybe should also modify the external endpoint at some point. But this looks fine to me now

PR #932, opened 2023-01-30
Clean up non-bundle based deploys

leeweisberger:
## Description
Clean up a few feature flags:
- Bundle based deploys. This was the largest one. There is no need for the non-bundle route anymore
- Airplane config creation
- Inline views

## Test plan
I tested airplane deploy, airplane tasks deploy, airplane views deploy


ameyakhare (approved):
:prayge:

PR #934, opened 2023-01-31
Support adding env vars alongside dev config file creation

calufornia:
## Description
Supports adding an env var to a dev confile file that has not been created yet—previously, we'd get an error around inserting an entry into a nil map.

## Test plan
Tested env var creation when a dev config file did not exist yet.


pkg/conf/dev.go
```diff
@@ -39,14 +39,34 @@ type DevConfig struct {
 // NewDevConfig returns a default dev config.
 func NewDevConfig(path string) *DevConfig {
 	return &DevConfig{
-		RawConfigVars: map[string]string{},
 		RawResources:  []map[string]interface{}{},
-		ConfigVars:    map[string]env.ConfigWithEnv{},
 		Resources:     map[string]env.ResourceWithEnv{},
+		RawConfigVars: map[string]string{},
+		ConfigVars:    map[string]env.ConfigWithEnv{},
+		EnvVars:       map[string]string{},
```
calufornia, line 46: This is the fix

colinking (approved):

PR #942, opened 2023-02-02
Return non-500 errors in Studio

calufornia:
## Description

Goes through most of the studio error paths and returns `errorsx` errors (e.g. `ErrBadRequest`, etc.). Adds logic to convert these errors into various HTTP response codes, similar to what we do in our prod API. Also makes some errors in Studio more clear.

TODO: Add sentry reporting

Depends on https://github.com/airplanedev/lib/pull/537

## Test plan
Verified that non-500 errors were returned in Studio.


cmd/airplane/tasks/dev/dev.go
```diff
@@ -162,9 +161,6 @@ func New(c *cli.Config) *cobra.Command {
 		logger.Debug("marking --server-host as hidden: %v", err)
 	}
 
-	if err := cmd.Flags().MarkDeprecated("editor", "launching the Studio is now the default behavior."); err != nil {
```
calufornia, line 165: Studio has been the default behavior for a while, so we can remove this. If users want an escape hatch, they can still do `--studio=false`

pkg/server/handlers/handlers.go
```diff
@@ -115,14 +113,54 @@ func HandlerZip(state *state.State,
...
+	if _, err := w.Write(out); err != nil {
+		// TODO: Report to sentry
 		return
 	}
 }
+
+// GetErrorStatus gets an errors HTTP status, code and message.
+func GetErrorStatus(ctx context.Context, err error) (errStatusCode libhttp.ErrStatusCode, retryable bool) {
+	switch {
+	case errors.As(err, &errStatusCode):
```
calufornia, line 146: @colinking Much cleaner with `errors.As`!

pkg/server/handlers/handlers.go
```diff
@@ -26,9 +24,9 @@ func Wrap(f func(ctx context.Context, w http.ResponseWriter, r *http.Request) er
 	}
 }
 
-// HandlerWithBody is an API handler that reads a JSON-encoded request body, calls a provided handler,
+// WithBody is an API handler that reads a JSON-encoded request body, calls a provided handler,
 // and then writes the JSON encoded response. It is used for handling an API request with a body.
-func HandlerWithBody[Req any, Resp any](state *state.State,
+func WithBody[Req any, Resp any](state *state.State,
```
calufornia, line 29: I renamed a bunch of these since functions in Golang ideally do not start with their package name 🙂

pkg/resources/resources.go
```diff
@@ -36,7 +39,12 @@ func GenerateAliasToResourceMap(
 		var resource resources.Resource
 		resourceWithEnv, ok := LookupResource(slugToResource, ref)
 		if !ok {
-			return nil, errors.Errorf("Cannot find resource with ref %s in dev config file or remotely", ref)
+			msg := fmt.Sprintf("cannot find resource %s, is it defined in your dev config file", ref)
```
colinking, line 42:
```suggestion
			msg := fmt.Sprintf("cannot find resource %q. Is defined in your dev config file", ref)
```

pkg/server/apidev/apidev.go
```diff
@@ -364,12 +369,12 @@ func StartViewHandler(ctx context.Context, s *state.State, r *http.Request) (Sta
...
 	viewSlug, ok := vars["view_slug"]
 	if !ok {
-		return StartViewResponse{}, errors.Errorf("View slug was not supplied, request path must be of the form /dev/startView/<view_slug>")
+		return StartViewResponse{}, libhttp.NewErrBadRequest("view slug was not supplied")
 	}
 
 	viewConfig, ok := s.ViewConfigs.Get(viewSlug)
 	if !ok {
-		return StartViewResponse{}, errors.Errorf("View with slug %s not found", viewSlug)
+		return StartViewResponse{}, libhttp.NewErrBadRequest("View with slug %s not found", viewSlug)
```
colinking, line 377: Prefer %q -- it's better for empty strings 

```suggestion
		return StartViewResponse{}, libhttp.NewErrBadRequest("view with slug %q not found", viewSlug)
```

pkg/server/apidev/apidev.go
```diff
@@ -472,12 +477,12 @@ func LogsHandler(ctx context.Context, state *state.State, r *http.Request, flush
...
 	runID, ok := vars["run_id"]
 	if !ok {
-		return errors.Errorf("Run id was not supplied, request path must be of the form /dev/logs/<run_id>")
+		return libhttp.NewErrBadRequest("run id was not supplied")
 	}
 
 	run, ok := state.Runs.Get(runID)
 	if !ok {
-		return errors.Errorf("Run with id %s not found", runID)
+		return libhttp.NewErrNotFound("Run with id %s not found", runID)
```
colinking, line 485: nit: lowercase error msgs for wrapping

```suggestion
		return libhttp.NewErrNotFound("run with id %q not found", runID)
```

pkg/server/handlers/handlers.go
colinking: It would be nice to match the naming conventions in API, but that's something we could do another time. Would also be nice to share them (and we could add hooks if we need to add additional logic in airport)

colinking (approved):
Nice! Just make sure the casing is consistent and we're using %q

PR #943, opened 2023-02-03
Make tsconfig diffs prettier

yolken-airplane:
## Description
This change updates the message we show to users when we need to change their tsconfig. The current version is hard to understand because it shows the diff by key. The updated one uses more of a git-style diff (generated by a third-party library).

Resolves DTS-3

## Test plan
Testing completed successfully via running locally:

Screenshot 2023-02-02 at 7 44 53 PM


calufornia (approved):
nice!

PR #946, opened 2023-02-06
Add studio status and properly set server host in remote studio

calufornia:
## Description

Returns the proper server host (e.g. `sandbox-router.fly.dev`) in the `/dev/info` endpoint, and also adds the notion of "Status" to the dev server. For now, there are two statuses: `Discovering`, while entities are being discovered, and `Ready`, after entities have been discovered.

## Test plan
Expected server host and status returned in `/dev/info`.


pkg/server/status/status.go
```diff
@@ -0,0 +1,8 @@
+package status
+
+type ServerStatus string
+
+const (
+	ServerDiscovering ServerStatus = "discovering"
```
ameyakhare, line 6: are there any more states than this?

ameyakhare (approved):

pkg/server/status/status.go
calufornia: We can probably add some more states (with finer granularity), but just adding these to start.

PR #950, opened 2023-02-08
Report internal server errors to sentry

calufornia:
## Description
Reports Studio 500s to sentry, similar to our prod API.

## Test plan
Triggered a 500 manually and verified that the sentry reporting codepath was hit.


pkg/server/handlers/handlers.go
```diff
@@ -111,19 +112,21 @@ func Zip(state *state.State,
 	})
 }
 
-// WriteHTTPError writes an error to response and optionally logs it
-func WriteHTTPError(w http.ResponseWriter, r *http.Request, err error) {
+// WriteHTTPError writes an error to response and optionally reports it
+func WriteHTTPError(w http.ResponseWriter, r *http.Request, err error, report func(err error)) {
```
colinking, line 116: Why did you make this configurable? AFAICT, `WriteHTTPError` is not used anywhere else.

pkg/server/handlers/handlers.go
```diff
@@ -111,19 +112,21 @@ func Zip(state *state.State,
...
 	ctx := r.Context()
 
 	errStatusCode, retryable := GetErrorStatus(ctx, err)
 
-	// TODO: Log 500 errors to sentry
+	if errStatusCode.StatusCode == http.StatusInternalServerError {
```
colinking, line 121: Report anything >= 500. In case we add logic in the future to return other 5xxs.

colinking (approved):

pkg/server/handlers/handlers.go
calufornia: I imagine a world where this function will be used by our API as well 😢

PR #953, opened 2023-02-08
Configure port proxy auth in sandbox mode

calufornia:
## Description

Similar to https://github.com/airplanedev/cli/pull/949, configure port proxy auth when the dev server is started up in sandbox mode. We fetch a sandbox token from our API and append it to view URLs. We'll handle verifying this token in our API.

Also splits out the server startup code into various files.

## Test plan
Verified that sandbox token was appended to view URLs


ameyakhare (approved):
test plan?

PR #957, opened 2023-02-10
Add remote env slug to LocalRun

francescaleung:
## Description

Supports fallback env switcher. Runs can now exist in multiple environments.

Resolves DTS-281, DTS-290, DTS-289.

## Test plan

Tested locally.


pkg/server/apiext/apiext.go
```diff
@@ -278,6 +293,8 @@ func ExecuteTaskHandler(ctx context.Context, state *state.State, r *http.Request
 		run.Remote = true
 		run.ID = resp.RunID
 		run.RunID = resp.RunID
+		run.EnvSlug = pointers.ToString(envSlug)
+		run.FallbackEnvSlug = pointers.ToString(envSlug)
```
calufornia, line 297: We don't need to set the fallback env slug on remote runs, right?

calufornia (approved):
Could you add a test to `apiext_test.go` that tests executing a task with a parent run and verifying that the env slug is set properly?

pkg/server/apiext/apiext.go
francescaleung: probably not, but it seems like a nice thing to include anyway. could be nice for debugging.

PR #958, opened 2023-02-10
Add ListEnvs endpoint

francescaleung:
## Description

Add ListEnvs endpoint to support fallback env switcher. Depends on https://github.com/airplanedev/cli/pull/957.

## Test plan

Tested locally.


pkg/server/apiint/apiint.go
```diff
@@ -834,3 +836,21 @@ func GetUploadHandler(
...
+}
+
+func ListEnvsHandler(ctx context.Context, state *state.State, r *http.Request) (ListEnvsResponse, error) {
+	resp, err := state.RemoteClient.ListEnvs(ctx)
+	if err != nil {
+		return ListEnvsResponse{}, errors.Wrap(err, "error getting envs")
+	}
+
+	for _, env := range resp.Envs {
+		state.EnvCache.Add(env.Slug, env)
```
calufornia, line 851: Do we want to remove an env from the env cache if it is no longer returned in the remote response?

pkg/server/apiint/apiint_test.go
```diff
@@ -512,6 +512,7 @@ func TestRemoteConfigs(t *testing.T) {
 		context.Background(),
 		t,
 		server.NewRouter(&state.State{
+			EnvCache: state.NewStore[string, libapi.Env](nil),
```
calufornia, line 515: Can you add a dedicated test for `/i/envs/list`?

calufornia (approved):

PR #959, opened 2023-02-10
Improve error message when ngrok tunnel is already bound

ameyakhare:
Resolves DTS-241

## Description
Adds a better description if the user already has multiple tunnel-based studio sessions running.

## Test plan
Opened multiple tunnel sessions and the second one errored with the correct message.


cmd/airplane/tasks/dev/tunnel.go
```diff
@@ -29,6 +30,12 @@ func configureTunnel(ctx context.Context, client *api.Client, authInfo api.AuthI
 		ngrok.WithAuthtoken(tokenResp.Token),
 		ngrok.WithRegion("us"),
 	); err != nil {
+		if strings.Contains(err.Error(), "is already bound to another tunnel session") {
+			// This is a heuristic since we don't have direct exposure to the ngrok error object (where
+			// we could check the error code). Further, the error code is wrong according to their
```
calufornia, line 35: Where in the docs do they mention this? Can you link it in the comment?

calufornia (approved):

PR #961, opened 2023-02-10
Pull fallback env slug from header

francescaleung:
## Description

Uses fallback env slug from header in server requests (configs, resources, entrypoints, task executions). Does not address views.

Resolves DTS-278. Depends on https://github.com/airplanedev/cli/pull/957, https://github.com/airplanedev/cli/pull/958/.

## Test plan

Tested locally.


pkg/configs/configs.go
```diff
@@ -70,9 +74,9 @@ func MergeRemoteConfigs(ctx context.Context, state *state.State) (map[string]env
 	}
 }
 
-func ListRemoteConfigs(ctx context.Context, state *state.State) ([]api.Config, error) {
+func ListRemoteConfigs(ctx context.Context, state *state.State, envSlug *string) ([]api.Config, error) {
```
calufornia, line 77: Just have this take in a `string`, no need for pointer—we shouldn't be calling this method with a nil env slug

pkg/server/apidev/apidev.go
```diff
@@ -82,7 +83,7 @@ func GetVersionHandler(ctx context.Context, s *state.State, r *http.Request) (ve
 type StudioInfo struct {
 	Workspace            string              `json:"workspace"`
 	DefaultEnv           libapi.Env          `json:"defaultEnv"`
-	FallbackEnv          *libapi.Env         `json:"fallbackEnv"`
+	FallbackEnv          *libapi.Env         `json:"fallbackEnv"` // This is specifically the fallback env that the server starts with
```
calufornia, line 86: Rename this to `InitialFallbackEnv`?

cmd/airplane/tasks/dev/server.go
```diff
@@ -142,7 +142,7 @@ func runLocalDevServer(ctx context.Context, cfg taskDevConfig) error {
 	apiServer.RegisterState(&state.State{
 		LocalClient:      &localClient,
 		RemoteClient:     cfg.root.Client,
-		RemoteEnv:        remoteEnv,
+		RemoteEnvSlug:    cfg.envSlug,
 		UseFallbackEnv:   cfg.useFallbackEnv,
```
calufornia, line 146: Do we need `UseFallbackEnv` anymore? If `RemoteEnvSlug` is non-nil, then that means we have a fallback environment right?

pkg/resources/resources.go
```diff
@@ -129,9 +130,9 @@ func mergeDefaultRemoteResource(
...
 // set, we still return a set of default remote resources for convenience.
-func ListRemoteResources(ctx context.Context, state *state.State) ([]libapi.Resource, error) {
-	if state.UseFallbackEnv {
-		resp, err := state.RemoteClient.ListResources(ctx, state.RemoteEnv.Slug)
+func ListRemoteResources(ctx context.Context, state *state.State, envSlug *string) ([]libapi.Resource, error) {
+	if envSlug != nil {
+		resp, err := state.RemoteClient.ListResources(ctx, *envSlug)
 		if err != nil {
 			return nil, err
 		}
```
calufornia, line 138: Do we also need to include `envSlug` in the `GetResource` request below?

pkg/resources/resources.go
```diff
@@ -50,7 +51,7 @@ func GenerateAliasToResourceMap(
 		if resourceWithEnv.Remote {
 			remoteResourceWithCredentials, err := state.RemoteClient.GetResource(ctx, api.GetResourceRequest{
 				ID:                   resourceWithEnv.Resource.GetID(),
-				EnvSlug:              state.RemoteEnv.Slug,
+				EnvSlug:              *envSlug,
```
calufornia, line 54: So IIRC `state.RemoteEnv` before was always populated to allow fetching default resource credentials. It seems like now, `*envSlug` has the possibility of being nil, in which case this would panic right?

calufornia (commented):
Dropping some initial comments

francescaleung:
note to self: this needs some transition so we don't break fallback environments before we're ready to release this

pkg/server/apidev/apidev.go
francescaleung: I'm leaving the JSON name alone, I don't want to mess with that.

pkg/resources/resources.go
francescaleung: no, these are specifically "default" resources that come from your default environment (just the demo db). I'll leave a comment though.

francescaleung:
Note that I've modified web to explicitly send "no-fallback-environment" (which is an invalid Airplane slug). If the header isn't sent, we use the fallback environment that the server starts with, rather than no fallback environment. This means we can release a new CLI without fallback environments breaking.

pkg/server/state/state.go
```diff
@@ -33,14 +33,10 @@ type ViteContext struct {
...
-	RemoteClient api.APIClient
-	RemoteEnv    libapi.Env
-	StudioURL    url.URL
-	// We need UseFallbackEnv because there will always be a RemoteEnv (to look for a user's team's default resources).
-	// UseFallbackEnv determines whether we should use this remote env for tasks/resources/etc. outside of these
-	// defaults.
-	UseFallbackEnv bool
+	LocalClient   *api.Client
+	RemoteClient  api.APIClient
+	RemoteEnvSlug *string
```
calufornia, line 38: Should this also be `initialRemoteEnvSlug`?

calufornia (approved):
LGTM—double check that things work when you specify `--env ''` (empty string that gets mapped to default). I'm pretty sure we resolve this to the actual default environment (with a non-empty slug) early on but let's just make sure that still works.

pkg/resources/resources.go
francescaleung: No default resources, just local + fallback. Pretty sure this matches what was being done previously (e.g., if you set your fallback environment to something like stage, you wouldn't get default resources from prod).

PR #962, opened 2023-02-14
Report panics

calufornia:
## Description

Captures and reports all panics to Sentry instead of printing them to the terminal. Continues to print the panic if the `--debug` flag is enabled. Adds some nil checks.

Resolves DTS-297

## Test plan
Manually induced panics and verified that we recovered and reported to sentry.


pkg/configs/configs.go
```diff
@@ -71,6 +71,10 @@ func MergeRemoteConfigs(ctx context.Context, state *state.State) (map[string]env
 }
 
 func ListRemoteConfigs(ctx context.Context, state *state.State) ([]api.Config, error) {
+	if state.RemoteClient == nil {
```
colinking, line 74: Why not return an error in this case? Is this expected?

pkg/server/apiint/apiint.go
```diff
@@ -187,20 +187,22 @@ type ListResourcesResponse struct {
...
-				Slug:              slug,
-				Kind:              libapi.ResourceKind(r.Resource.Kind()),
-				ExportResource:    r.Resource,
-				CanUseResource:    true,
-				CanUpdateResource: true,
-			},
-			Remote: false,
-			Env:    env.NewLocalEnv(),
-		})
+	if state.DevConfig != nil {
```
colinking, line 190: Could we initialize these to zero values instead? We're going to end up accidentally adding code that doesn't include these guards again.

colinking (approved):

PR #963, opened 2023-02-14
Add decompression middleware into dev servers

yolken-airplane:
## Description
This change updates the dev server mux to support compressed request bodies. This is needed if any clients are using the updated client logic added in https://github.com/airplanedev/lib/pull/552.

Note that the middleware is very similar to the one added to the API in https://github.com/airplanedev/airport/pull/4069. Unfortunately, we can't move that to lib at the moment and share it between the two repos because it depends on the airport-specific errorlib (which has also been semi-replicated on the CLI side).

cc: @calufornia 

## Test plan
Testing completed via extra unit tests and running locally.


colinking (approved):

colinking:
@yolken-airplane IIUC, no clients are using gzip against the CLI server yet, correct? I believe only `web` and the various task SDKs talk to the CLI. If either is updated to use gzip, we'll need to bump the studio's minimum CLI version (which is defined in `web`).

yolken-airplane:
> @yolken-airplane IIUC, no clients are using gzip against the CLI server yet, correct? I believe only `web` and the various task SDKs talk to the CLI. If either is updated to use gzip, we'll need to bump the studio's minimum CLI version (which is defined in `web`).

Yeah, nothing is using gzip yet, but I was worried that something might in the future, so wanted to be prepared here. :-).

PR #966, opened 2023-02-16
Fix error checking of signal killed

calufornia:
## Description

This was causing non-signal killed errors to redownload the builtins binary.

## Test plan
Tested locally.


pkg/server/apiext/apiext.go
```diff
@@ -247,9 +247,9 @@ func ExecuteTaskHandler(ctx context.Context, state *state.State, r *http.Request
 				}
 
 				// If the error is a signal killed error, the builtins binary is likely corrupt. Manually trigger a
-				// re-download of the builtins binary.
-				builtinErr := dev_errors.SignalKilled
-				if errors.As(err, &builtinErr) {
+				// re-download of the builtins binary. We need to compare the error string because the error is
+				// reported by the builtins binary itself, which doesn't use the error in the `dev_errors` package.
+				if err.Error() == dev_errors.SignalKilled.Error() {
```
francescaleung, line 252: this feels really brittle, is this really the best we can do?

francescaleung (approved):
approving to unblock

pkg/server/apiext/apiext.go
calufornia: switched to using `exec.ExitError` instead

PR #969, opened 2023-02-17
Generate alias -> resource map in headless dev

calufornia:
## Description

Modifies the `GenerateAliasToResourceMap` so that it can be used by headless dev without needing to pass in a state. This was panicking before when attempting to call `state.UseFallbackEnv`, as state was nil for headless dev. Also adds a test for this method.

Also fixes an issue where the default dev config file was not getting loaded in properly in headless dev.

Also makes it so that we capture all panics in the CLI, and displays the panic stacktrace only when the user has enabled the `--debug` flag.

## Test plan
Tested with `airplane dev --studio=false` and verified panics are not displayed to the user anymore unless `--debug` mode is enabled.


cmd/airplane/main.go
```diff
@@ -24,6 +27,22 @@ func main() {
 
 	cmd.Version = version
 
+	defer func() {
+		if r := recover(); r != nil {
+			logger.Log("Panic detected: %+v", r) // This does not print the stack trace.
```
colinking, line 32: This will render in red text, similar to what we have below.

```suggestion
			logger.Error("The CLI unexpectedly crashed: %+v", r) // This does not print the stack trace.
```

Do you also want to report it via `analytics.ReportError`?

pkg/resources/resources_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"context"
+	"testing"
+
+	"github.com/airplanedev/cli/pkg/dev/env"
+	libresources "github.com/airplanedev/lib/pkg/resources"
+	"github.com/airplanedev/lib/pkg/resources/kinds"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGenerateAliasToResourceMap(t *testing.T) {
```
colinking, line 13: awyis

colinking (approved):

cmd/airplane/main.go
calufornia: I could be wrong, but I believe that Sentry already reports it in `Recover`: https://docs.sentry.io/platforms/go/panics/#usage

PR #972, opened 2023-02-21
Clean up feature flag and default to inline configured tasks

leeweisberger:
## Description
This has been out for a while - default to inline configured tasks. Remove feature flag check.


pkg/flags/flagsiface/flags.go
```diff
@@ -6,9 +6,7 @@ import (
 	"github.com/airplanedev/cli/pkg/logger"
 )
 
-const (
-	DefaultInlineConfigTasks = "default-inline-tasks"
-)
+const ()
```
ameyakhare, line 9: remove this block

ameyakhare (approved):

PR #973, opened 2023-02-22
Update default tsconfigs

leeweisberger:
## Description
Update default tsconfigs to add recommended fields.

I looked up a bunch of best practice tsconfigs, and they pretty much all have these settings enabled.

esModuleInterop: This fixes a few issues with how typescript imports commonjs.

skipLibCheck: This makes type checking faster by ignoring .d.ts files which are already compiled. This is turned on in most default tsconfigs


geodn (approved):
Let's change the description about making the other PR unnecessary

PR #982, opened 2023-02-27
Update `/tasks/get` endpoint to return all fields

colinking:
## Description

We need this endpoint to return to all fields so that we can render the new config panel in studio.

This also now uses `GetUpdateTaskRequest` to re-use logic from within the `lib/pkg/definitions`.

## Test plan

Confirmed I could dev on tasks via studio.


colinking:
* main
  * **PR #982** <a href="https://app.graphite.dev/github/pr/airplanedev/cli/982" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/cli/982?utm_source=stack-comment).

pkg/server/apiint/apiint.go
```diff
@@ -620,41 +622,66 @@ func GetTaskInfoHandler(ctx context.Context, state *state.State, r *http.Request
...
+		Description:                utr.Description,
+		Image:                      utr.Image,
+		Command:                    utr.Command,
+		Arguments:                  utr.Arguments,
+		Parameters:                 utr.Parameters,
+		Configs:                    []libapi.ConfigAttachment{},
+		Constraints:                utr.Constraints,
+		Env:                        utr.Env,
+		ResourceRequests:           utr.ResourceRequests,
+		Resources:                  utr.Resources,
```
colinking, line 656: Notably, this will now return resources as `{ALIAS: ID}` rather than `{ALIAS: SLUG}` which matches what the main API does. I've updated `web` in: https://github.com/airplanedev/web/pull/2193 (which will merge before this PR).

pkg/server/apiint/apiint.go
```diff
@@ -620,41 +622,66 @@ func GetTaskInfoHandler(ctx context.Context, state *state.State, r *http.Request
...
 	if err != nil {
-		return libapi.Task{}, errors.Wrap(err, "getting parameters")
+		return libapi.Task{}, libhttp.NewErrBadRequest("task %q is invalid: %s", taskSlug, err.Error())
 	}
-	req.Parameters = parameters
-	kind, options, err := taskConfig.Def.GetKindAndOptions()
-	if err != nil {
-		return libapi.Task{}, errors.Wrap(err, "getting kind and options")
+
+	req := libapi.Task{
```
colinking, line 643: I'm worried that we'll add new task fields and not realize we also need to add them here. I'm tempted to move this into `lib/pkg/definitions` as a `(Definition) GetTask()` method. wdyt @francescaleung?

francescaleung: seems reasonable

PR #984, opened 2023-02-28
Upgrade lib

colinking:
## Description

Upgrades `lib` to bring in: https://github.com/airplanedev/lib/pull/567

## Test plan

Tested a deployment from the CLI locally. Also tested the local dev server (e.g. executing a task, getting a run).


go.mod
```diff
@@ -5,7 +5,7 @@ go 1.20
 require (
 	github.com/AlecAivazis/survey/v2 v2.3.6
 	github.com/MakeNowJust/heredoc v1.0.0
-	github.com/airplanedev/lib v0.5.163
+	github.com/airplanedev/lib v0.5.165-0.20230228171046-772fa7385c0c
```
colinking, line 8: Will update once I merge `lib`.

pkg/server/apiext/apiext.go
```diff
@@ -174,7 +174,7 @@ func ExecuteTaskHandler(ctx context.Context, state *state.State, r *http.Request
 			run.StdAPIRequest = stdapiReq
 			run.TaskName = req.Slug
 			run.ParamValues = req.ParamValues
-		} else if localTaskConfig.Def != nil {
```
colinking, line 177: Since we switched from an interface to a struct, a Definition can no longer be nullable. AFAICT, this was never the case and these statements were only defensive programming.

francescaleung (approved):

PR #985, opened 2023-02-28
Fix AppCondition cache

francescaleung:
## Description

`AppCondition` is an eager cache for errors + warnings & isn't environment-aware. This patch changes it to a lazy cache that _is_ environment-aware. When something on the filesystem changes, the whole cache is thrown out, which is erring on the side of refreshing the cache very often--but because it's now a lazy cache, I don't think this is too problematic.

This patch also removes the `resources` -> `state` dependency + the unused `UnattachedResource` struct.

Resolves DTS-379.

## Test plan

Started a studio server, navigated to a task page that used resources that existed in the fallback environment. Task page now does not indicate that resources are missing.


pkg/server/state/state.go
```diff
@@ -143,6 +146,73 @@ func (s *State) GetEnv(ctx context.Context, envSlug string) (libapi.Env, error)
...
+		}
+		if len(missingResources) > 0 {
+			sort.Strings(missingResources)
+			result.Errors = append(result.Errors, dev_errors.AppError{
+				Level:   dev_errors.LevelWarning,
+				AppName: taskConfig.Def.GetName(),
+				AppKind: "task",
+				Reason:  fmt.Sprintf("Attached resource: %v not found in dev config file or remotely.", missingResources),
+			})
+		}
```
colinking, line 197: Unrelated to your change, but the array brackets in this error message have bothered me before. It seems straightforward to create one error per missing resource. What do you think?

```suggestion
		for _, resource := range missingResources {
			result.Errors = append(result.Errors, dev_errors.AppError{
				Level:   dev_errors.LevelWarning,
				AppName: taskConfig.Def.GetName(),
				AppKind: "task",
				Reason:  fmt.Sprintf("Attached resource %q not found in dev config file or remotely.", resource),
			})
		}
```

pkg/server/server.go
```diff
@@ -411,36 +375,30 @@ type DiscoverOpts struct {
...
+		return nil, errors.Wrap(err, "validating task")
 	}
 	if opts.OverwriteAll {
 		// clear existing tasks, task errors, and views
 		s.state.TaskConfigs.ReplaceItems(map[string]discover.TaskConfig{})
-		s.state.AppCondition.ReplaceItems(map[string]state.AppCondition{})
 		s.state.ViewConfigs.ReplaceItems(map[string]discover.ViewConfig{})
 	}
-	now := time.Now()
+	// Always throw out all AppConditions
```
colinking, line 388: Adjusted wording to clarify it's a cache.

```suggestion
	// Always invalidate the AppConditions cache.
```

pkg/server/server.go
```diff
@@ -411,36 +375,30 @@ type DiscoverOpts struct {
...
+	s.state.AppCondition.ReplaceItems(map[string]state.AppCondition{})
 	for _, cfg := range opts.Tasks {
-		if _, isUnsupported := warnings.UnsupportedApps[cfg.Def.GetSlug()]; !isUnsupported {
+		if _, isUnsupported := unsupportedApps[cfg.Def.GetSlug()]; !isUnsupported {
 			s.state.TaskConfigs.Add(cfg.Def.GetSlug(), cfg)
```
colinking, line 392: Unrelated to your change, but it seems to me that overwriting should be done atomically? Could we build up the `map[string]discover.TaskConfig` and then do a call to `s.state.TaskConfigs.ReplaceItems(...)` at the end?

pkg/server/server.go
```diff
@@ -242,42 +237,11 @@ func ValidateTasks(ctx context.Context, resourcesWithEnv map[string]env.Resource
...
-					AppKind: apidev.EntityKindTask,
-					Reason:  fmt.Sprintf("Attached resource: %v not found in dev config file or remotely.", missingResources),
-				},
-			}
-		}
 	}
 
-	return dev_errors.RegistrationWarnings{
-		UnsupportedApps:     unsupportedApps,
-		UnattachedResources: unattachedResources,
```
colinking, line 278: So this was just never used anywhere, right?

colinking (approved):
Nice. Thanks for debugging!

pkg/server/server.go
francescaleung: yea, nothing was reading this. Didn't dig into the git history to figure out when/where this had previously got used.

PR #990, opened 2023-03-01
Sanitize inputs in studio

francescaleung:
## Description

Resolves DTS-381.

## Test plan

image
image


pkg/server/state/state.go
```diff
@@ -215,6 +219,77 @@ func appConditionKey(slug, envSlug string) string {
...
+	if s.Flagger != nil && s.Flagger.Bool(ctx, s.Logger, flagsiface.SanitizeInputs) {
+		parameters := libapi.Parameters{}
+		if run.Parameters != nil {
+			parameters = *run.Parameters
+		}
+		sanitized, err := libparams.SanitizeParamValues(run.ParamValues, parameters)
+		if err != nil {
+			return dev.LocalRun{}, errors.Wrap(err, "sanitizing param values")
+		}
+		run.ParamValues = sanitized
```
calufornia, line 240: nit: extract this into something like
```
func sanitizeParamValues(run dev.LocalRun) err {
        parameters := libapi.Parameters{}
	if run.Parameters != nil {
		parameters = *run.Parameters
	}
	sanitized, err := libparams.SanitizeParamValues(run.ParamValues, parameters)
	if err != nil {
		return errors.Wrap(err, "sanitizing param values")
	}
	run.ParamValues = sanitized
}
```

which we can use here and below

calufornia (approved):

PR #996, opened 2023-03-07
Add stub endpoints for /runbooks/execute and /runbooks/get

geodn:
## Description
Add stub endpoints returning an error saying that runbooks are not supported.

Fixes PDS-2126

## Test plan
Built cli and ran a test view
Screenshot 2023-03-07 at 12 03 16 AM


calufornia (approved):
nit: Could you put these handlers in a new file `runbooks.go`? I just want to start moving handlers outside of `api*.go` files.

PR #1004, opened 2023-03-09
Pass sandbox token in API_HEADERS

calufornia:
## Description

Include the sandbox token in the bag of API headers that are included amongst requests. There was a [change](https://github.com/airplanedev/lib/pull/560/files#diff-e8f5f1eeb2dbf108a36275c5a5fa914eca498ab36457d39e09599d83318b1a70R38) that caused the airplane sandbox token to not be set in environment variables, which is fine for requests that use the View fetcher, but for requests that use the JS sdk fetcher, we depended on that env var being present. We probably need to include the sandbox token in the execute options like [here](https://github.com/airplanedev/views/blob/924795b6a2a1a35dc9261753e6e6b80c7ac72cc3/lib/src/client/env.ts#L54), but for now just forward these in the API headers.

## Test plan
Verified that header was properly set locally


francescaleung (approved):
this only affects TRS, right? i.e. it's only been broken for TRS in the last week, & not for normal users.

calufornia:
> this only affects TRS, right? i.e. it's only been broken for TRS in the last week, & not for normal users.

Correct, this has only affected the `X-Airplane-Sandbox-Token` header, which is not applicable to normal users

PR #1009, opened 2023-03-10
Use api client interface instead of implementation types

calufornia:
## Description

There are many places in the CLI code where we use the implementation type for the API client (e.g. `api.Client`) instead of the interface type (`api.APIClient`), which makes it difficult to test CLI commands E2E. This PR changes all these instances to instead accept the interface type, which will allow us to utilize a mock API client. 

Also changes the implementation methods for `api.Client` to have pointer receivers since we now have methods that mutate the Client (`Set...`, etc.)

Resolves DTS-250

cc @colinking 

## Test plan

Verified that this is a no-op locally.


calufornia:
* main
  * **PR #1009** <a href="https://app.graphite.dev/github/pr/airplanedev/cli/1009" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/cli/1009?utm_source=stack-comment).

leeweisberger (approved):

PR #1011, opened 2023-03-13
Add prompter interface

calufornia:
## Description

Adds a `Prompter` interface so that we can mock user stdin during tests. This is the [recommended approach](https://github.com/go-survey/survey/issues/448#issuecomment-1253631737) for mocking out survey calls. We have two implementations of the `Prompter` interface: a `Surveyor` that uses the `survey` package, and `Mock` that we can pre-load responses into.

Resolves DTS-250

## Test plan
Verified that this is a no-op locally.


calufornia:
* main
  * **PR #1009** <a href="https://app.graphite.dev/github/pr/airplanedev/cli/1009" target="_blank">Graphite</a> 
    * **PR #1011** <a href="https://app.graphite.dev/github/pr/airplanedev/cli/1011" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/cli/1011?utm_source=stack-comment).

pkg/prompts/survey.go
```diff
@@ -0,0 +1,115 @@
...
+
+import (
+	"os"
+
+	"github.com/AlecAivazis/survey/v2"
+	"github.com/pkg/errors"
+)
+
+// Surveyor is a Prompter that uses the survey package.
+type Surveyor struct{}
```
leeweisberger, line 11: Add

var _ Prompter = Surveyor{}

This should provide a compile time check in this file that surveyor implements prompter

pkg/prompts/prompts.go
```diff
@@ -0,0 +1,89 @@
...
+		o.Default = defaultValue
+	}
+}
+
+func WithDefaultTrue() Opt {
+	return WithDefault(true)
+}
+
+func WithDefaultFalse() Opt {
+	return WithDefault(false)
```
leeweisberger, line 77: Can we just delete these and use WithDefault?

pkg/prompts/prompts.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+}
+
+func WithRequired() Opt {
+	return func(o *opts) {
+		o.Required = true
+	}
+}
+
+func WithSecret() Opt {
```
leeweisberger, line 42: Can you add comments to all these? I'm not sure what all of them do personally

pkg/prompts/prompts.go
```diff
@@ -0,0 +1,89 @@
...
+	Help                string
+	Default             interface{}
+	Select              []string
+	Suggest             func(toComplete string) []string
+}
+
+type Opt func(opts *opts)
+
+// WithRedirectStdOutToErr redirects stdout to stderr.
+func WithRedirectStdOutToErr() Opt {
```
leeweisberger, line 30: What exactly does this do? Why would I set this? it seems like we set this most of the time. Is there are reason we don't set it all of the time?

leeweisberger (changes requested):

pkg/prompts/prompts.go
calufornia: I just wanted to make this PR as much of a no-op as possible. I think we can set it all the time, I will go ahead and do that

leeweisberger: If we want to set it all the time, we can just default it to true and then nobody has to set it.

leeweisberger (approved):

PR #1017, opened 2023-03-16
Add dev server endpoint to generate and write SQL task

calufornia:
## Description

Adds an endpoint, `/dev/autopilot/generate`, that generates the contents for a SQL task and writes two files: the SQL entrypoint and the task definition yaml. 

Resolves DTS-446

## Test plan
Verified locally with frontend changes.


pkg/server/apidev/autopilot.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/airplanedev/cli/pkg/utils"
+	libhttp "github.com/airplanedev/lib/pkg/api/http"
+	"github.com/airplanedev/lib/pkg/build"
+	"github.com/pkg/errors"
+)
+
+type GenerateType string
+
+const (
+	CreateSQLTask       GenerateType = "create-sql-task"
```
ameyakhare, line 21: I feel like it might be better to abstract this as action, subject, subject type instead of these permutations of things

ie action: create/insert
subject type: view/task
subject: sql, etc.

pkg/server/apidev/autopilot.go
```diff
@@ -0,0 +1,110 @@
...
+	// Write content to file
+	if err := os.WriteFile(filepath.Join(s.Dir, sqlEntrypoint), []byte(res.Content), 0644); err != nil {
+		return errors.Wrap(err, "writing sql entrypoint")
+	}
+
+	res, err = s.RemoteClient.AutopilotComplete(ctx, api.AutopilotCompleteRequest{
+		Type: api.TaskYAMLCompletionType,
+		Prompt: fmt.Sprintf(
+			// TODO: Include this in the prompt in our API, not here.
+			"Generate a task definition YAML with slug %s, entrypoint %s, and resource slug %s and only return YAML without any prefixes or postfixes in your answer. Name and description should not be empty.",
```
ameyakhare, line 89: we discussed this offline but this prompt should contain everything necessary except shaping commands (ie name and description should not be empty, remove pre/post fixes) I think these should just be added to the backend prompt generation and this should be as simple as possible.

ameyakhare (changes requested):

pkg/server/apidev/autopilot.go
calufornia: Yep, I'll just remove the TODO for now and move the shaping commands in separate PR.

pkg/server/apidev/autopilot.go
```diff
@@ -0,0 +1,110 @@
...
+			},
+		},
+	})
+	if err != nil {
+		return err
+	}
+
+	// Generate random file name
+	rand := utils.RandomString(3, utils.CharsetAlphaLowercase)
+	slug := fmt.Sprintf("my_sql_task_%s", rand)
```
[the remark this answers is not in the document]
calufornia: Yeah, that's probably better—I'll add a TODO for now

pkg/server/apidev/autopilot.go
```diff
@@ -0,0 +1,119 @@
...
+
+type GenerateSQLContext struct {
+	ResourceID   string `json:"resourceID"`
+	ResourceSlug string `json:"resourceSlug"`
+}
+
+func AutopilotGenerateHandler(ctx context.Context, s *state.State, r *http.Request, req AutopilotGenerateRequest) (struct{}, error) {
+	autopilotContext := req.Context
+	// TODO: Add support for other autopilot contexts.
+	if autopilotContext.Action == Create && autopilotContext.Subject == Task && autopilotContext.SubjectKind == SQL {
```
ameyakhare, line 59: would be nice to put this in some sort of validator class/func instead of just a long conditional, but leaving that up to you

ameyakhare (approved):

PR #1018, opened 2023-03-17
Optimistically update task cache on update

colinking:
## Description

When we update a task via the new config panel, optimistically update the task in the in-memory cache. This change will get re-applied on the next task discovery loop, but this prevents `/i/tasks/get` calls from returning the outdated task in the meantime.

## Test plan

Tested via https://github.com/airplanedev/web/pull/2220


colinking:
* main
  * **PR #1018** <a href="https://app.graphite.dev/github/pr/airplanedev/cli/1018" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/cli/1018?utm_source=stack-comment).

pkg/server/apiint/apiint.go
```diff
@@ -714,7 +715,16 @@ func UpdateTaskHandler(ctx context.Context, state *state.State, r *http.Request,
 	}
 
 	// Update the underlying task file.
-	if err := rt.Edit(ctx, state.Logger, taskConfig.TaskEntrypoint, req.Slug, taskConfig.Def); err != nil {
+	if err := rt.Edit(ctx, state.Logger, taskConfig.Def.GetDefnFilePath(), req.Slug, taskConfig.Def); err != nil {
```
colinking, line 718: This includes a fix for editing YAML files. Had to use the definition file path not the entrypoint.

PR #1020, opened 2023-03-17
Use build tools package.json for views local dev dependencies

calufornia:
## Description

Reads dev dependencies for local views dev from our build tools package.json, which includes an upgrade to Vite. Also changes some of the typing to use `build.PackageJSON` instead, which is cleaner. Adds tests.

Resolves DTS-213

## Test plan
Tested developing a view locally.


pkg/views/dev.go
```diff
@@ -308,8 +308,8 @@ func addDevDepsToPackageJSON(existingPackageJSON map[string]interface{}, package
 		"object-hash":     "3.0.0",
 	}
 	defaultDevDeps := map[string]string{
-		"@vitejs/plugin-react": "2.1.0",
-		"vite":                 "3.1.3",
+		"@vitejs/plugin-react": "3.1.0",
+		"vite":                 "4.2.0",
```
leeweisberger, line 312: I was gonna do this, but instead of this change, can you import `BuildToolsPackageJSON` from lib and then read package versions from there? We should be using the same package versions when building as when devving

leeweisberger (changes requested):

pkg/views/dev.go
```diff
@@ -288,53 +291,41 @@ func createViteConfig(root string, airplaneViewDir string, port int, token *stri
...
-	}
-	deps, ok := (*packageJSON)["dependencies"].(map[string]interface{})
-	if !ok {
-		(*packageJSON)["dependencies"] = map[string]interface{}{}
-		deps = (*packageJSON)["dependencies"].(map[string]interface{})
-	}
-	addMissingDeps(defaultDeps, existingDeps, &deps, false)
+	existingDeps := existingPackageJSON.Dependencies
+	deps := packageJSON.Dependencies
+	addMissingDeps(defaultDeps, existingDeps, deps, false)
```
calufornia, line 324: The logic to not override the user's dependencies still exists.

leeweisberger (approved):

PR #1025, opened 2023-03-21
Load studio env vars during discovery

ameyakhare:
Resolves DTS-460

## Description
Adds basic studio env vars with fixed values as described here: https://docs.airplane.dev/tasks/runtime-env-vars

## Test plan
I loaded a task that had the slug in the name and it appropriately added "studio" to the name.


cmd/airplane/tasks/dev/server.go
```diff
@@ -277,3 +281,21 @@ func runLocalDevServer(ctx context.Context, cfg taskDevConfig) error {
 
 	return nil
 }
+
+func getDiscoveryEnvVars(devConfig *conf.DevConfig) []string {
```
calufornia, line 285: Move this function into `pkg/dev/envvars.go`

cmd/airplane/tasks/dev/server.go
```diff
@@ -277,3 +281,21 @@ func runLocalDevServer(ctx context.Context, cfg taskDevConfig) error {
...
+	envVars := []string{
+		"AIRPLANE_RUNTIME=build", // copies builder behavior
+		"AIRPLANE_ENV_SLUG=studio",
+		"AIRPLANE_ENV_IS_DEFAULT=true",
+		"AIRPLANE_ENV_ID=studio",
+		"AIRPLANE_ENV_NAME=studio",
```
calufornia, line 291: Once this function is moved into `envVars.go`, use `getCommonEnvVars()` here for these last 4 values instead

calufornia (approved):

cmd/airplane/tasks/dev/server.go
calufornia: You might need to resolve some circular dependency issues, but I think it's better to keep env var logic in there

PR #1026, opened 2023-03-21
Add allow list of system env vars

calufornia:
## Description

We made a [change](https://github.com/airplanedev/cli/commit/f8406ee6652a151845f6c5df178edd1c194849da) that stopped system environment variables from propagating to the run execution environment. However, the `go-snowflake` package [uses the `$HOME` environment variable](https://github.com/snowflakedb/gosnowflake/blob/d04e9aad8c13a649b6d347f4ee87d227c15f5e0c/ocsp.go#L904) to determine the caching directory. Because this was empty, the directory was being created in the directory where `airplane dev` was being run.

This PR adds an allow list of environment variables that are forwarded from the user's system into the run environment.

## Test plan
Unit tests and verify that `Library/` folder is not created anymore.


pkg/dev/envvars_test.go
```diff
@@ -58,6 +59,33 @@ func TestSetEnvVars(t *testing.T) {
 	require.True(requireEnvVarExists(envVars, "ENV_VAR_FROM_CONFIG", configVarValue))
 }
 
+func TestSetSystemEnvVars(t *testing.T) {
+	var home string
+	var ok bool
+	if home, ok = os.LookupEnv("HOME"); !ok {
```
calufornia, line 65: `$HOME` should be set in the Github runner, but may not necessarily be set on a developer's machine.

pkg/dev/envvars_test.go
```diff
@@ -58,6 +59,33 @@ func TestSetEnvVars(t *testing.T) {
...
+		t.Skip("$HOME not set")
```
stepzhou, line 66: Why skip instead of fail the test?

stepzhou (approved):

pkg/dev/envvars_test.go
stepzhou: Oh I see

PR #1027, opened 2023-03-22
Add `/i/tasks/canUpdate` endpoint

colinking:
## Description

Adds a new CLI endpoint that can be used by the studio to detect if it can update a task. This endpoint is specific to the studio, as tasks can only be updated locally if they do not contain "computed fields".

Depends on:
- https://github.com/airplanedev/lib/pull/607
- https://github.com/airplanedev/lib/pull/608

Used by:
- https://github.com/airplanedev/web/pull/2282

## Test plan

Ran locally and confirmed that tasks with dynamic logic could not be updated.


pkg/server/apiint/apiint.go
```diff
@@ -788,6 +789,45 @@ func UpdateTaskHandler(ctx context.Context, state *state.State, r *http.Request,
...
+	if err != nil {
+		return CanUpdateTaskResponse{}, err
+	}
+
+	rt, err := runtime.Lookup(taskConfig.TaskEntrypoint, kind)
+	if err != nil {
+		return CanUpdateTaskResponse{}, err
+	}
+
+	canUpdate, err := rt.CanUpdate(ctx, state.Logger, taskConfig.Def.GetDefnFilePath(), slug)
```
ameyakhare, line 821: im assuming this is in another PR

ameyakhare: ah yes its this: https://github.com/airplanedev/lib/pull/607/files

ameyakhare (approved):

colinking:
@yolken-airplane: I made a couple small breaking changes in `lib` that I fix in this PR. The next version of `lib` includes breaking changes from both of us, so we'll need to merge our changes together.

yolken-airplane:
> @yolken-airplane: I made a couple small breaking changes in `lib` that I fix in this PR. The next version of `lib` includes breaking changes from both of us, so we'll need to merge our changes together.

Sounds good, thanks for the head's up.

PR #1028, opened 2023-03-22
Add insertion support to dev server

calufornia:
## Description

Adds support for generating a view component—the dev server takes the user's current cursor position and inserts an `{{INSERT}}` block that we tell the model to replace with a component. Also moves the `autopilot` code into its own package, reorganizes some code, and adds some missing types.

Resolves DTS-447

## Test plan
Verified in conjunction with web + api changes


pkg/server/autopilot/insert.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+}
+
+func insertComponentIntoView(ctx context.Context, s *state.State, prompt string, context *GenerateViewContext) (string, error) {
+	if context == nil {
+		return "", libhttp.NewErrBadRequest("missing view context")
+	}
+
+	var builder strings.Builder
+	builder.WriteString(fmt.Sprintf("Replace %s with a %s component that %s. ONLY return the generated component.", insertBlock, context.Kind, prompt))
```
ameyakhare, line 40: I feel like this prompt insertion may not make sense depending on what the user writes?

pkg/server/autopilot/insert.go
```diff
@@ -0,0 +1,113 @@
...
+	// The last line is the line that the cursor is on. Track how many spaces are in front of the cursor so that we
+	// can properly indent this line.
+	insertLineLeadingSpaces := getNumLeadingSpaces(lines[len(lines)-1])
+
+	return indentationSpaces, insertLineLeadingSpaces
+}
+
+// getNumLeadingSpaces returns the number of leading spaces in the given line.
+func getNumLeadingSpaces(line string) int {
+	for i, c := range line {
```
ameyakhare, line 90: what if they use tabs?

ameyakhare (changes requested):
just have some questions about the prompting style and indentation strategy

pkg/server/autopilot/insert.go
calufornia: CodeMirror by default seems to convert tabs into spaces—so we should be able to use this for now. I'll add a TODO

PR #1037, opened 2023-04-03
Simplify default view

stepzhou:
## Description

* Simplify default view by removing a bunch of fluff
* Make the View component name based off of the view name rather than being `ExampleView`

image

## Test plan

Build CLI locally and went through `airplane views init` flow

Fix tests


stepzhou:
@geodn Do you think you could get to this today? I have a couple of PRs stacked behind this

cmd/airplane/root/initcmd/fixtures/view/my_view.airplane.tsx
```diff
@@ -1,120 +1,44 @@
...
-      {selectedCustomer && (
-        <Table
-          title={`Users for ${selectedCustomer.name}`}
-          data={usersData.filter((u) => u.customer_id == selectedCustomer.id)}
-          columns={usersCols}
-          hiddenColumns={["customer_id"]}
-        />
-      )}
+    <Stack spacing="lg">
+      <Stack spacing={0}>
```
geodn, line 7: I like the change, my one comment is that I think the spacing=0 stack is ugly, especially for our default view. I'd rather have the spacing be slightly off (but still good) than have these extra Stacks. Can we get rid of it?

geodn (approved):

cmd/airplane/root/initcmd/fixtures/view/my_view.airplane.tsx
stepzhou: No, it just looks too awful without the zero spacing stacks.

image

I feel like this is a weakness of our Stack and something we should try to improve

PR #1039, opened 2023-04-03
Change init views filename

stepzhou:
## Description

Change the init views filename from `my_view.airplane.tsx` to a more React-idiomatic `MyView.airplane.tsx`

## Test plan

Tests pass


cmd/airplane/views/initcmd/init_test.go
```diff
@@ -11,8 +11,7 @@ import (
 )
 
 func TestInit(t *testing.T) {
-	// TODO(justin, 04152023): re-enable tests once we resolve the flaky yarn install collisions
-	t.Skip()
+	t.Setenv("YARN_CACHE_FOLDER", t.TempDir())
```
stepzhou, line 14: From the test failure (https://github.com/airplanedev/cli/actions/runs/4600787985/jobs/8127835501), I think the issue was some global Yarn cache collision when the tests are running in parallel or something.

I did some searching and this is a way to avoid cache collisions for each test

See
* https://github.com/yarnpkg/yarn/issues/986#issuecomment-279147561
* https://classic.yarnpkg.com/en/docs/cli/cache/

cc: @colinking

colinking: Good find. Can you re-enable the tests in the other `initcmd` packages? https://github.com/search?q=org%3Aairplanedev+NOT+is%3Aarchived+TODO%28justin%2C+04152023%29%3A+re-enable+tests+once+we+resolve+the+flaky+yarn+install+collisions&type=code

calufornia (approved):

cmd/airplane/views/initcmd/init_test.go
stepzhou: I'll do this in a quick follow-up PR!

PR #1044, opened 2023-04-06
Fix update logic for SQL tasks

colinking:
## Description

Fixes a bug where the entrypoint would get overwritten when editing a SQL task.

## Test plan

Edited a SQL task via the studio config panel.


colinking:
* main
  * **PR #1044** <a href="https://app.graphite.dev/github/pr/airplanedev/cli/1044" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/cli/1044?utm_source=stack-comment).

pkg/deploy/taskdir/definitions/task_def_update.go
```diff
@@ -119,22 +119,34 @@ func (d *Definition) updateResources(resources api.Resources, kind buildtypes.Ta
...
 	case buildtypes.TaskKindShell:
-		d.Shell = &ShellDefinition{}
+		if d.Shell == nil {
+			d.Shell = &ShellDefinition{}
+		}
 		return d.Shell.update(t, availableResources)
 	case buildtypes.TaskKindSQL:
-		d.SQL = &SQLDefinition{}
+		if d.SQL == nil {
+			d.SQL = &SQLDefinition{}
```
colinking, line 143: The issue is that `d.SQL` has fields, such as the entrypoint path, that aren't inferred from the task. In general, we should be doing an update not an overwrite, to match the update semantics used elsewhere when updating a task definition.

francescaleung (approved):

PR #1049, opened 2023-04-11
Shard golang tests and cache by date

yolken-airplane:
## Description
This change updates the cli repo to use the golang test performance optimizations that we've been using in airport for a while:

1. Shard the tests so the very expensive ones run independently
2. Include the date in the cache so that the caches are regularly refreshed even if `go.mod` doesn't change

## Test plan
Testing to be completed via running through CI.


.github/workflows/golang.yml
```diff
@@ -61,15 +77,30 @@ jobs:
...
       - name: go test
         run: |
-          go test -race -timeout=30m -p 4 ./...
+          go test -race -timeout=30m -p 4 $(./.github/workflows/split_tests.py ${{ matrix.shard }})
+
+  # Create no-op test job that depends on sharded tests so we don't need to
+  # switch around our required checks.
+  test:
+    needs:
+      - sharded-test
```
colinking, line 102: This would presumably add some overhead (idk, maybe 10s or so) since it would need to get scheduled. Why not just change the required checks?

colinking (approved):

.github/workflows/golang.yml
yolken-airplane: It's pretty fast, like 3-4 seconds? :-) I can look into changing the required checks instead.

colinking: If you don't have permission to change, LMK, and I can update them!

yolken-airplane: Oh, actually, I just looked into this and it's a bit of a pain because you have to add each shard to the required checks individually, and if we add more shards in the future, we'd need to remember to manually update that. Let me just update the comment here and leave it as-is for now.

Screenshot 2023-04-12 at 3 04 32 PM

PR #1051, opened 2023-04-12
Extract init commands to pkg

francescaleung:
## Description

Functionality change: I removed the `initWithExample` path from tasks (currently in [`example.go`](https://github.com/airplanedev/cli/blob/e1f83d6466fefa38a6a3d35168c7ef4961d0ae35/cmd/airplane/tasks/initcmd/example.go#L19)). This is not documented in `--help` & I'm pretty sure the examples repo doesn't exist anymore (which is what it was initially for, although I think it'll take any github path with an Airplane task def file in it).

Aside from that this is mostly a fairly straight move, with some renaming so as to avoid name collisions.

Supports DTS-303.

## Test plan

Unit tests, sanity tests.


pkg/initcmd/initcmd.go
```diff
@@ -0,0 +1,88 @@
+package initcmd
```
colinking, line 1: nit: rename file to `suggest.go`? Naming it `initcmd.go` (which matches the package name) implies it contains the "main" logic for the package.

pkg/initcmd/task.go
```diff
@@ -0,0 +1,722 @@
...
+				def = definitions.NewBuiltinDefinition(
+					req.TaskName,
+					slug,
+					&definitions.GraphQLDefinition{
+						Operation: `query GetUser($id: ID) {
+  user(id: $id) {
+    id
+	name
+	email
+  }
```
colinking, line 97: query spacing is a mix of tab + space

cmd/airplane/views/initcmd/init.go
```diff
@@ -58,124 +47,26 @@ func GetConfig(c *cli.Config) config {
 
 func Run(ctx context.Context, cfg config) error {
 	if cfg.from != "" {
-		if err := utils.CopyFromGithubPath(cfg.from, cfg.root.Prompter); err != nil {
+		if err := initcmd.InitViewFromExample(ctx, initcmd.InitViewFromExampleRequest{
+			Prompter:    cfg.root.Prompter,
+			ExamplePath: cfg.from,
```
colinking, line 52: Is initializing views from an "example" used? I'm surprised this got called `--from` since it differs from `airplane tasks init` 😞

pkg/initcmd/view.go
```diff
@@ -0,0 +1,105 @@
...
+	Name     string
+}
+
+func InitView(ctx context.Context, req InitViewRequest) error {
+	if req.Name == "" {
+		return errors.New("missing new view name")
+	}
+
+	slug := utils.MakeSlug(req.Name)
+	viewDir := ""
```
colinking, line 32: Don't forget to port this from `cfg.viewDir`

cmd/airplane/views/initcmd/init.go
colinking: You'll need the `(&cfg).viewDir = filepath.Base(cfg.from)` logic here.

colinking (approved):
Make sure to test the various init flags to make sure they still work.

cmd/airplane/views/initcmd/init.go
francescaleung: It's documented, still works.

`(&cfg).viewDir = filepath.Base(cfg.from)` isn't needed here because `cfg` never gets read again. `suggestNextSteps` has been moved into `InitView{,FromExample}`.

pkg/initcmd/view.go
francescaleung: There's nothing to port, `cfg.viewDir` is always empty string.

colinking: Wasn't it previously set to a non-empty value via `(&cfg).viewDir = filepath.Base(cfg.from)`?

PR #1052, opened 2023-04-13
Add schedule validation endpoint

colinking:
## Description

Adds an API endpoint for viewing the next runs of a schedule in studio.

## Test plan

Ran studio locally (with [this PR](https://github.com/airplanedev/web/pull/2371)) and confirmed I could view upcoming schedule runs.


colinking:
* main
  * **PR #1052** <a href="https://app.graphite.dev/github/pr/airplanedev/cli/1052" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/cli/1052?utm_source=stack-comment).

pkg/schedules/cron_expr.go
```diff
@@ -0,0 +1,98 @@
+package schedules
```
colinking, line 1: This is ported from `airport` (without changes). I'll remove it from `airport` in a follow-up change.

francescaleung (approved):
optionally we could pass it on to the API server. I guess that's kind of a weird thing to do though, especially if there's not any Airplane-specific logic involved.

PR #1055, opened 2023-04-13
Add hidden --dry-run option to (some) init options

francescaleung:
## Description

This is restricted to InitTask + InitView for now. Does not include: InitWorkflowFromRunbook, InitFromTemplate, InitViewFromExample.

Resolves DTS-592.

## Test plan

Unit tests.


pkg/initcmd/task.go
```diff
@@ -242,8 +248,20 @@ func InitTask(ctx context.Context, req InitTaskRequest) error {
...
+		Prompter: req.Prompter,
+		DryRun:   req.DryRun,
+		Inline:   req.Inline,
+		Kind:     kind,
+		Def:      def,
+	})
+	if err != nil {
+		return nil, err
+	}
+	filesCreated = append(filesCreated, kindSpecificFilesCreated...)
```
colinking, line 261: Do we want to distinguish between which files are created vs. edited?

What about previewing the changes that we're going to make? For example, it might be useful to know how it's going to change your `tsconfig.json` before you save it.

pkg/initcmd/task.go
```diff
@@ -575,82 +596,111 @@ func apiTaskToRuntimeTask(task *libapi.Task) *runtime.Task {
...
+				Dependencies: node.NodeDependencies{
+					Dependencies:    []string{"airplane"},
+					DevDependencies: []string{"@types/node"},
+				},
+			}, req.Prompter)
+			if err != nil {
+				return nil, err
+			}
 		}
+		filesCreated = append(filesCreated, filepath.Join(packageJSONDir, "package.json"))
```
colinking, line 627: Are all files absolute paths? Or are they relative to some directory? (if so, what directory makes sense?)

Speaking of which, should the UI clarify where certain files are going to be created?

colinking (approved):
Looking back at the mocks, there's not much thought put into showing the files that are going to be created. I added a comment in the figma about this, so we can discuss further there. Not blocking.

pkg/initcmd/task.go
francescaleung: I thought about that, & if we really want to go ahead & do this I'll add in the bits for it, but otherwise I'm willing to throw "updates + creates" into the same pile & call it a day.

pkg/initcmd/task.go
francescaleung: I think they're kind of a mix right now. I'm adding a WorkingDirectory to InitTask which should probably help. I'll address this in a later PR, since this isn't being used right now.

PR #1056, opened 2023-04-13
Ensure that logs in the task don't affect config parsing

leeweisberger:
## Description
We find the entity config log to determine the entity configs discovered by the parser. If the task itself produces logs, this can mess up the logic used to find the line.

I updated the local discover code to use the same regex logic as the server-side discoverer so it shouldn't be tripped up by additional logs.

## Test plan
I added a test that failed before and now passes.


ameyakhare (approved):
lgtm

PR #1057, opened 2023-04-13
Add file patching endpoint to dev server

calufornia:
## Description

Adds a new endpoint, `/dev/files/patch`, to the dev server that will be used to patch files in the user's development root. It takes in a set of [unified diffs](https://en.wikipedia.org/wiki/Diff#:~:text=syntax%20highlighting.-,Unified%20format,-%5Bedit%5D) (as strings), and uses the gitdiff package to parse the diffs and apply them. Also moves all the files handlers to their own file.

## Test plan
Verified locally with a test patch.


pkg/server/files/patch.go
```diff
@@ -0,0 +1,65 @@
...
+	"strings"
+
+	"github.com/airplanedev/cli/pkg/server/state"
+	"github.com/bluekeyes/go-gitdiff/gitdiff"
+	"github.com/pkg/errors"
+)
+
+func BatchPatch(s *state.State, patches []string) error {
+	for _, patch := range patches {
+		if err := Patch(s, patch); err != nil {
```
ameyakhare, line 16: is each patch on a per-file basis? you can't apply multiple patches to the same file right? or does that work properly, even if applying one patch changes the state of the file and may invalidate the other patches?

pkg/server/files/patch.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// Patch applies a unified diff patch. The patch should contain a preamble that describe the file we're patching, and so
+// we can infer exactly the changes that the patch will cause.
+func Patch(s *state.State, patch string) error {
+	patchReader := strings.NewReader(patch)
+
+	// diffs is a slice of *gitdiff.File describing the diffs included in the patch. This will error if the patch is
+	// malformed.
+	diffs, _, err := gitdiff.Parse(patchReader)
```
ameyakhare, line 30: does this require that the user has git installed? is this making any git sys calls?

pkg/server/files/patch.go
```diff
@@ -0,0 +1,65 @@
...
+		path := filepath.Join(s.Dir, diff.OldName)
+
+		f, err := os.OpenFile(path, os.O_RDWR, 0755)
+		if err != nil {
+			return errors.Wrap(err, "opening file")
+		}
+
+		// apply the changes in the patch to a source diff and store it in a buffer.
+		var output bytes.Buffer
+		err = gitdiff.Apply(&output, f, diff)
```
ameyakhare, line 49: ditto

ameyakhare (changes requested):
have some questions

pkg/server/files/patch.go
calufornia: Nope it doesn't—it's basically a reimplementation of what git diff is doing: https://github.com/bluekeyes/go-gitdiff/blob/master/go.mod

pkg/server/files/patch.go
calufornia: A file should be referenced by at most one patch (OTOH, a patch may reference multiple files), so nope it shouldn't invalidate other patches.

ameyakhare (approved):

PR #1064, opened 2023-04-14
Account for nil prompter in InitTask

francescaleung:
## Description

Moves all of `cmd/airplane/testutils` to `pkg/testutils` so it can be reused outside of `cmd`. Checks for a nil prompter in places where a prompter would be invoked via `InitTask` and either picks a sane default or returns an error. Also adds an optional WorkingDirectory to `InitTaskRequest` so invoking `InitTask` does not have to rely on the actual working directory.

Resolves DTS-593.

## Test plan

Unit tests!


pkg/initcmd/task.go
```diff
@@ -142,47 +163,82 @@ func InitTask(ctx context.Context, req InitTaskRequest) ([]string, error) {
 		}
 
 		if filepath.Ext(entrypoint) == "tsx" || filepath.Ext(entrypoint) == "jsx" {
-			logger.Log("You are trying to deploy a React file. Use `airplane views init` if you'd like to initialize a view.")
-			if ok, err := req.Prompter.ConfirmWithAssumptions("Are you sure you'd like to continue?", req.AssumeYes, req.AssumeNo); err != nil {
+			errorMsg := "You are trying to deploy a React file. Use `airplane views init` if you'd like to initialize a view."
```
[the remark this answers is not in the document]
colinking: Separately, @leeweisberger do we even need to ask this if the file exists? It's perfectly valid to create a task inside of a tsx file.

pkg/initcmd/task.go
```diff
@@ -142,47 +163,82 @@ func InitTask(ctx context.Context, req InitTaskRequest) ([]string, error) {
...
 
-			if fsx.Exists(entrypoint) {
-				shouldOverwrite, shouldPrintToStdOut, err := shouldOverwriteTaskEntrypoint(req, entrypoint, kind)
+		if req.Prompter == nil {
+			// Add a suffix to it.
+			absEntrypoint, err = tryRandomSuffix(absEntrypoint, addEntrypointSuffixFunc(req), 3, req.suffixCharset)
+			if err != nil {
+				return nil, errors.Wrap(err, "finding entrypoint")
+			}
+			entrypoint = filepath.Base(absEntrypoint)
```
colinking, line 207: Won't this break if the user specifies a folder (e.g. `support/entrypoint.ts`)?

colinking (approved):

pkg/initcmd/task.go
francescaleung: Current behaviour when user specifies a folder:
* entrypoint, airplane.yaml, tsconfig.json, requirements.txt, defn file get placed in the folder (some of these you can change during init but the default is to put it all in the folder)
* package.json gets placed in the wd

Seems like package.json should also be in the folder? Hilariously the value returned from `CreatePackageJSON` ("returns the path to the directory where the package.json is created/found") doesn't match where we actually put the package.json, because the package.json is always in the wd regardless...

I'll add a test for this.

francescaleung: to answer the question: This doesn't break, because `absEntrypoint` is either (a) just the given entrypoint, if it's already absolute, or (b) the given entrypoint relative to the working directory, if it isn't. After this point `entrypoint` is mostly used for display purposes; `absEntrypoint` gets used for most important things.

PR #1069, opened 2023-04-17
Improve slug/value parsing for shell

leeweisberger:
## Description
Improve the way we parse slugs and values for shell tasks. See https://airplanedev.slack.com/archives/C045CGB1B6H/p1681757490838859 for an example of parameters that throw an error and don't parse correctly.

The old way had issues with newlines if the separator (`=`) existed after the first line because of how `cut` operates.

Instead of using `cut` which has interactions with newlines, use a different splitting algorithm from unix stackexchange.

## Test plan
See https://replit.com/join/jpwemhwmge-leeweisberger1 for a comparison of the old vs new algorithm.

I wrote a unit test that covers a case that failed before with error and succeeds now.

I also tested the params in the slack thread and they work as intended.


colinking (approved):
🔥

PR #1072, opened 2023-04-18
Add Python task updater

colinking:
## Description

This PR adds support for editing inline Python tasks. This is implemented similar to JavaScript, except that:
1. It uses the built-in Python library for parsing code as an AST.
2. The AST library doesn't support converting ASTs to formatted code, so we use black for formatting.
3. Since we need a 3rd-party dependency, we package our Python code and black as a zipapp using Pex. Unfortunately, this does not support Windows (see [Slack](https://airplanedev.slack.com/archives/C04HC759C2E/p1682538754510169)).
4. Python task definitions sometimes require imports (e.g. datetime, typing) so we automatically add those as necessary.
5. Python task parameters are managed as function parameters, so we have additional logic for serializing those.

cc @francescaleung 

## Test plan

See unit tests. Ideally those are exhaustive -- please review them and point out any test cases that we should add!

You can also test it from the Studio using https://github.com/airplanedev/web/pull/2446. Just create an inline Python task and edit it with the config panel.


.editorconfig
```diff
@@ -8,6 +8,6 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
-[{*.go,go.mod,go.sum}]
```
colinking, line 11: TIL Makefiles require tabs.

pkg/deploy/discover/code_task_discover.go
```diff
@@ -16,12 +16,6 @@ import (
 	"github.com/pkg/errors"
 )
 
-//go:embed parser/node/parser.js
-var nodeParserScript []byte
-
-//go:embed parser/python/parser.py
-var pythonParserScript string
```
colinking, line 23: These scripts were both embedded in two places. I removed this one.

pkg/runtime/javascript/javascript.go
```diff
@@ -635,48 +623,6 @@ func (r Runtime) CanUpdate(ctx context.Context, logger logger.Logger, path strin
 	return updaters.CanUpdateYAML(path)
 }
 
-func runNodeCommand(ctx context.Context, logger logger.Logger, script []byte, args ...string) (string, error) {
```
colinking, line 638: I extracted this logic so we could reuse part of it for running the Python updater.

pkg/runtime/python/fixtures/update/all_cleared_out_airplane.py
```diff
@@ -0,0 +1,11 @@
+import airplane
+from datetime import datetime, timezone
+from typing import Annotated, Optional
```
colinking, line 3: note: we don't automatically remove these imports. We could add some logic to detect if they are used, but that doesn't seem worth the effort.

pkg/runtime/python/fixtures/update/all_out_airplane.py
```diff
@@ -0,0 +1,41 @@
...
+    ],
+)
+def my_task_2(
+    dry: Annotated[
+        Optional[bool],
+        airplane.ParamConfig(
+            name="Dry run?", description="Whether or not to run in dry-run mode."
+        ),
+    ] = True,
+    datetime: Optional[datetime] = None,
```
colinking, line 39: This requires a `= None` since `dry` has a default and because we match the parameter order from the task definition.

pkg/runtime/python/fixtures/update/all_out_airplane.py
```diff
@@ -0,0 +1,41 @@
+from datetime import datetime, timezone
+from typing import Annotated, Optional
+import airplane
+
+
+@airplane.task(
```
colinking, line 6: @ericliuche Take a look at these fixtures and confirm the syntax looks correct. Ideally we generate idiomatic code, so LMK if we could make anything better (e.g. skip kwargs).

pkg/runtime/python/fixtures/update/all_out_airplane.py
```diff
@@ -0,0 +1,41 @@
...
+    allow_self_approvals=False,
+    restrict_callers=["view", "task"],
+    schedules=[
+        airplane.Schedule(
+            slug="all",
+            name="All fields",
+            description="A description",
+            cron="0 12 * * *",
+            param_values={
+                "datetime": datetime(2006, 1, 2, 15, 4, 5, tzinfo=timezone.utc)
```
colinking, line 26: Should we do `datetime.datetime`? Or just `datetime`?

pkg/runtime/python/fixtures/update/imports_out_airplane.py
```diff
@@ -0,0 +1,16 @@
+"""
+This module has a docstring!
+"""
+from datetime import datetime
```
colinking, line 4: AD: we just dump new imports at the top of the file (but there's some logic for detecting comments).

pkg/runtime/python/fixtures/update/modern_syntax_airplane.py
```diff
@@ -0,0 +1,6 @@
+import airplane
+
+
+@airplane.task()
+async def my_task() -> int:
```
colinking, line 5: We don't support `async def`, right? Added anyway.

pkg/runtime/python/python.go
```diff
@@ -114,56 +119,41 @@ func (r Runtime) PrepareRun(ctx context.Context, logger logger.Logger, opts runt
...
 }
 
-// pythonBin returns the first of python3 or python found on PATH, if any.
+// getPythonBinary returns the first of python3 or python found on PATH, if any.
+//
 // We expect most systems to have python3 if Python 3 is installed, as per PEP 0394:
 // https://www.python.org/dev/peps/pep-0394/#recommendation
 // However, Python on Windows (whether through Python or Anaconda) does not seem to install python3.exe.
-func pythonBin(logger logger.Logger) string {
+func (r Runtime) getPythonBinary(ctx context.Context, logger logger.Logger) (string, error) {
```
colinking, line 131: I needed to re-use this. Refactored into a single function since we always called both functions together.

pkg/runtime/python/updater/pyproject.toml
```diff
@@ -0,0 +1,20 @@
+[tool.poetry]
+name = "updater"
+version = "1.0.0"
+description = ""
+authors = ["Airplane <support@airplane.dev>"]
+
+[tool.poetry.dependencies]
+python = ">=3.8,<3.12"
+black = "23.3.0"
+typing-extensions = "4.5.0"
```
colinking, line 10: I had to include typing-extensions so the code could be run in earlier versions of Python. Should I be wrapping the typing imports with `if` checks so it's only included when type checking? And then I could make this a dev dep so we don't have to include it in the pex file?

pkg/runtime/python/updater/src/main.py
```diff
@@ -0,0 +1,26 @@
+import json
+import sys
+import traceback
+
+if __name__ == "__main__":
+    try:
+        # We don't support 3.7 because the ast package does not include
+        # `end_lineno / end_col_offset` values on nodes.
+        # As of the time of writing this code, 3.7 hits EOL in 2 months.
```
colinking, line 9: ^ note: we don't support 3.7. Since it hits EOL soon...I think that's fine? I'm curious what the adoption on Airplane is.

pkg/runtime/python/updater/src/update.py
```diff
@@ -0,0 +1,127 @@
+# TODO: add can_update tests, including ones that error on the slug matching
+# TODO: add docstrings and module walkthrough
+# TODO: add tests that use modern Python syntax (async, return types)
+# TODO: explicitly set slug
+# TODO: test with unicode / multi-byte chars
+# TODO: make offsets 0-indexed?
+# TODO: generate code for min py version between sys.version_info and airplane.yaml
+# TODO: test other kinds of args, e.g. kwonlyargs
+# TODO: ignore name if it is equivalent to the slug?
+# TODO: generate a nicer error if using Windows
```
colinking, line 10: I will address these TODOs before merging this PR. However, I wanted to get this PR up for a first pass review.

colinking: I went through today and updated the rest of these!

cmd/airplane/tasks/dev/config/config.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/airplanedev/cli/cmd/airplane/tasks/dev/config/set_envvar"
 	"github.com/airplanedev/cli/cmd/airplane/tasks/dev/config/set_resource"
 	"github.com/airplanedev/cli/pkg/cli"
-	"github.com/airplanedev/cli/pkg/conf"
+	"github.com/airplanedev/cli/pkg/devconf"
```
colinking, line 15: I split out the `airplane.dev.yaml` logic out of the `conf` package into a new `devconf` package because of a circular dependency.

pkg/runtime/javascript/javascript_test.go
```diff
@@ -23,6 +23,7 @@ import (
 )
 
 func TestFormatComment(t *testing.T) {
+	t.Parallel()
```
colinking, line 26: We should do this more often. Added here because I was already doing this for the equivalent Python tests.

pkg/runtime/python/fixtures/update/py310_out_airplane.py
```diff
@@ -0,0 +1,7 @@
+from typing import Annotated
+import airplane
+
+
+@airplane.task()
+def my_task(name: Annotated[str | None, airplane.ParamConfig(name="User name")]):
```
colinking, line 6: Added support for looking at the Python version (as discussed) and using the latest possible version. e.g. we'll use the union operator if possible!

pkg/runtime/python/fixtures/update/unicode_airplane.py
```diff
@@ -0,0 +1,11 @@
+"""This is a test that uses multi-byte characters such as 你好世界, ߷, or 👨‍👩‍👧‍👧."""
```
colinking, line 1: This uncovered some bugs lol. Glad I tested it.

pkg/runtime/python/updater/src/serialize.py
```diff
@@ -0,0 +1,331 @@
...
+                ]
+            if self._is_py_version_gte("3.9"):
+                self.expected_imports.add("typing", "Annotated")
+            else:
+                self.expected_imports.add("typing_extensions", "Annotated")
+            kwargs = self._serialize_kwargs(annotated_param)
+            py_type = f"Annotated[{py_type}, airplane.ParamConfig({kwargs})]"
+
+        default = param.get("default", None)
+        if default is not None:
```
ericliuche, line 143: a better check might be `if "default" in param:` so we can allow for `foo: Optional[str] = None`. right now, there's no behavior difference to airplane between supplying the `= None` and not

pkg/runtime/python/updater/src/main.py
colinking: The AST doesn't include whitespace or comments, so the `end_lineno` of node N won't line up with `lineno` of node N+1.

pkg/runtime/python/fixtures/update/can_update_airplane.py
```diff
@@ -0,0 +1,43 @@
...
+
+# This is not valid and should not get matched on.
+@airplane.task()
+class MyTask:
+    @airplane.task()
+    def run(self):
+        pass
+
+
+@airplane.task(slug="deploy" if os.getenv("AIRPLANE_ENV") == "prod" else "deploy_dev")
```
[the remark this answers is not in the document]
colinking: Added a test that covers shared parameters:

```
parameters = []

@airplane.task(parameters=parameters)
def shared_parameters():
    pass
```

I'm less worried about computed fields in the parameters...so gonna leave that as is.

pkg/runtime/python/python.go
```diff
@@ -508,17 +502,94 @@ func (r Runtime) SupportsLocalExecution() bool {
...
+			pyVersion = cfg.Python.Version
+		}
+
+		envVars := []string{
+			// Store the pex venv under the .airplane directory instead of the user's home directory.
+			"PEX_ROOT=" + filepath.Join(conf.Dir(), "pex"),
+		}
+		if r.testingOnlyIgnoreCurrentPythonVersion {
+			envVars = append(envVars, "TESTING_ONLY_IGNORE_CURRENT_PYTHON_VERSION=true")
+		}
```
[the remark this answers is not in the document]
colinking: This is useful for our tests. If you were to run the tests with the wrong Python interpreter, the tests would fail.

pkg/runtime/python/updater/src/location.py
```diff
@@ -0,0 +1,59 @@
...
+        offset = 0
+        for i in range(lineno):
+            # Add one to account for the newline at the end of the line.
+            offset += len(lines[i]) + 1
+        return offset + col_offset
+
+    @classmethod
+    def from_node(cls, node: ast.AST, content: str) -> Self:
+        """Generates a Location from an AST node."""
+        end_lineno = node.end_lineno if hasattr(node, "end_lineno") else None
```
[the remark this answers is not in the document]
colinking: This is because the type is `Optional[int]`. Looks like I can get rid of these hasattr checks due to the assert, though.

pkg/runtime/python/updater/src/utils.py
```diff
@@ -0,0 +1,73 @@
...
+
+def camel_to_snake(s: str) -> str:
+    return re.sub(r"(?<!^)(?=[A-Z])", "_", s).lower()
+
+
+def apply_updates(content: str, updates: List[Tuple[Location, str]]) -> str:
+    # Update the locations to account for earlier updates. Assumes that updates do not overlap.
+    updates.sort(key=lambda a: a[0].start)
+    for i, update in enumerate(updates):
+        print(f"Will update {update[0]} with:\n{update[1]}")
```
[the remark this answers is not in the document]
colinking: We print this as debug output. It's also logged by the tests.

pkg/runtime/python/updater/src/serialize.py
```diff
@@ -0,0 +1,331 @@
...
+
+        # Extract python-specific fields.
+        if "python" in task_def:
+            if "envVars" in task_def["python"]:
+                # Rewrite env vars as a list instead of an object.
+                env_vars = [
+                    SerializedValue(self._serialize_env_var(key, value))
+                    for key, value in task_def["python"]["envVars"].items()
+                ]
+                task_def = insert_after(task_def, "python", "envVars", env_vars)
```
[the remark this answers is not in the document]
colinking: This retains the ordering, where `envVars` is written out wherever `python` was serialized, which is defined by the ordering of fields in the Go `Definition` struct.

pkg/runtime/python/updater/src/utils.py
```diff
@@ -0,0 +1,73 @@
...
+    return dict(items)
+
+
+def camel_to_snake(s: str) -> str:
+    return re.sub(r"(?<!^)(?=[A-Z])", "_", s).lower()
+
+
+def apply_updates(content: str, updates: List[Tuple[Location, str]]) -> str:
+    # Update the locations to account for earlier updates. Assumes that updates do not overlap.
+    updates.sort(key=lambda a: a[0].start)
```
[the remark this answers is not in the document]
colinking: Good call. Updated.

pkg/runtime/python/updater/src/serialize.py
colinking: How would you configure a default of `None`? If `default` is `nil`, the default is not set. Should we treat that as `None` if the parameter is optional>?

colinking:
@ericliuche updated to include the latest main + all your feedback. Only OQ I have left is https://github.com/airplanedev/cli/pull/1072#discussion_r1183029886

cmd/airplane/apikeys/create/create.go
```diff
@@ -9,7 +9,7 @@ import (
 
 	"github.com/airplanedev/cli/pkg/api/cliapi"
 	"github.com/airplanedev/cli/pkg/cli"
-	"github.com/airplanedev/cli/pkg/logger"
+	"github.com/airplanedev/cli/pkg/utils/logger"
```
colinking, line 12: While merging main, I noticed we have two logger packages (due to the lib migration). I merged them together.

pkg/runtime/python/updater/src/serialize.py
ericliuche: hmm yeah i think we want to set it to `None`, imagine the case of
```
def f(foo: Optional[str], bar: Optional[str] = None):
    pass
```
from Airplane's perspective, you can call `f` from our app without setting `foo` or `bar`. both will have `None` passed in.

you can also call this directly from another python task, via `f(...)`, but if you dont set `= None` for `foo`, then your linter will complain that a non defaulted argument has no value, e.g. you'll be forced to set `f(None)`

pkg/runtime/python/fixtures/update/can_update_airplane.py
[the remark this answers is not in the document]
ericliuche: a little confused, i don't think `parameters` is a valid field on `airplane.task`. i also don't think that computed fields in parameters is that far of a stretch but i'm also fine to leave it out if it's difficult, e.g.

```
MY_DEFAULT_CONSTANT = "foo"

@task()
def my_task(val: str = MY_DEFAULT_CONSTANT):
    ...

@task()
def my_other_task(val: str = MY_DEFAULT_CONSTANT):
    ...
```

ericliuche (approved):

pkg/runtime/python/fixtures/update/can_update_airplane.py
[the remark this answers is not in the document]
colinking: Whoops, thanks! Will fix.

What would you want to support in that example?
- is it that if they rename `val`, that we retain the `= MY_DEFAULT_CONSTANT` instead of swapping it for `= "foo"`? That gets tricky because we'd need to resolve the value of `MY_DEFAULT_CONSTANT` and ensure it is still `"foo"`.
- is it that changing the default value changes the value of `MY_DEFAULT_CONSTANT`? If so, we purposefully don't want to do that, otherwise it affects `my_other_task` which may not be intentional.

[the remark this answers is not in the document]
ericliuche: what's the current behavior? i agree with both of your points and shouldn't do that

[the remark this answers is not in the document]
colinking: The current behavior is that we say we can't update that kind of task (via the `can_update` check). Well, I suppose we only check the decorator rn since I didn't consider default values as a way to introduce conditional logic -- I'll add support for detecting this.

PR #1073, opened 2023-04-18
Add permissions to json schema and task definition

rose-huang:
## Description
Adding permissions field to the task def and task schema.json

## Test plan
task def tests pass
also printed out the task def we parse for these cases:
1. no permissions field set => permissions is nil (should not lock to code)
2. permissions: team_access => permissions is pointer to struct with `RequireExplicitPermission = false`
3. permissions: {} => permissions is pointer to an struct with `RequireExplicitPermission = true` (lock to code with explicit permissions but none assigned)
4. permissions: viewers:... => permissions is pointer to a struct with the defined permissions and `RequireExplicitPermission = true` (lock to code with explicit permissions)


rose-huang:
@leeweisberger I added support for "team_access." Didn't split up the schema changes and the task definition as we discussed earlier because it'll slow down development

pkg/deploy/taskdir/definitions/task_def.go
```diff
@@ -140,6 +141,69 @@ type ScheduleDefinition struct {
 	ParamValues map[string]interface{} `json:"paramValues,omitempty"`
 }
 
+type PermissionsDefinition struct {
+	Viewers                    PermissionRecipients `json:"viewers,omitempty"`
+	Requesters                 PermissionRecipients `json:"requesters,omitempty"`
+	Executers                  PermissionRecipients `json:"executers,omitempty"`
+	Admins                     PermissionRecipients `json:"admins,omitempty"`
+	RequireExplicitPermissions bool                 `json:"-"`
```
leeweisberger, line 149: Can you add a comment to `RequireExplicitPermissions` explaining how this interacts with the other fields? i.e. this should always be set if the other fields are set, and if just it is set that means we require explicit permissions but no additional permissions are set. Something like that

pkg/deploy/taskdir/definitions/task_def.go
```diff
@@ -140,6 +141,69 @@ type ScheduleDefinition struct {
...
+	if len(p.Requesters.Groups) > 0 || len(p.Requesters.Users) > 0 {
+		m["requesters"] = p.Requesters
+	}
+	if len(p.Executers.Groups) > 0 || len(p.Executers.Users) > 0 {
+		m["executers"] = p.Executers
+	}
+	if len(p.Admins.Groups) > 0 || len(p.Admins.Users) > 0 {
+		m["admins"] = p.Admins
+	}
+	return m, nil
```
leeweisberger, line 204: Why can't you do the same thing you did for json? Like yaml.Marshal or something?

leeweisberger (approved):

pkg/deploy/taskdir/definitions/task_def.go
rose-huang: Hmm I got a stack overflow error with something like `return yaml.Marshal(m)` if it wasn't team_access

leeweisberger: And you don't get this for json? I would think you would since it would recursively call the `MarshalJSON` function. Did you test this with json? 

Why can't you just use the same solution you used with unmarshal - the type alias bit https://medium.com/picus-security-engineering/custom-json-marshaller-in-go-and-common-pitfalls-c43fa774db05

rose-huang: I added tests for marshaling permissions for both YAML and JSON. I don't think we ever hit the MarshalJSON function with the current way we are marshaling permissions so I'm pretty sure that's why it didn't stack overflow 🤔 

I removed the MarshalJSON and updated the MarshalYAML function and tests for both are still passing 🤔 
https://github.com/airplanedev/cli/pull/1077

PR #1076, opened 2023-04-19
Speed up studio discovery reloads by 1s

colinking:
## Description

This PR updates the debounce logic used by the file watcher in studio. Previously, the watcher would wait at least one second before reloading. This is unnecessary in many cases (e.g. if you have auto save off) so this ends up saving one second on re-discovery. The new implementation will fire immediately before performing debouncing logic.

I couldn't find a good debounce implementation that implemented "leading" debounces, so I implemented my own.

cc @kellymtrinh @calufornia 

## Test plan

Test cases pass.

I ran studio locally and simulated frequent saving by spamming "CMD-S" in VSCode while printing out file events + reloads. As expected, a file event would be logged every time I saved the editor, but a reload would only happen (1) on the first save, (2) whenever I stopped saving, and (3) every 5s.


colinking:
* main
  * **PR #1076** <a href="https://app.graphite.dev/github/pr/airplanedev/cli/1076" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/cli/1076?utm_source=stack-comment).

yolken-airplane (approved):

PR #1077, opened 2023-04-19
Fix MarshalYAML for permissions definition

rose-huang:
## Description

follow up for https://github.com/airplanedev/cli/pull/1073#discussion_r1171786778

> I added tests for marshaling permissions for both YAML and JSON. I don't think we ever hit the MarshalJSON function with the current way we are marshaling permissions so I'm pretty sure that's why it didn't stack overflow 🤔 

> I removed the MarshalJSON and updated the MarshalYAML function and tests for both are still passing 🤔 
https://github.com/airplanedev/cli/pull/1077

## Test plan
Tests in task_def_test.go pass


pkg/deploy/taskdir/definitions/task_def.go
```diff
@@ -180,31 +180,15 @@ func (p *PermissionsDefinition) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
-func (p PermissionsDefinition) MarshalJSON() ([]byte, error) {
```
rose-huang, line 183: I can add a MarshalJSON that does something like 
```
	if !p.RequireExplicitPermissions {
		return "team_access", nil
	}
	// Note we need a new type, otherwise we recursively call this
	// method and end up stack overflowing.
	type permissions PermissionsDefinition
	perm := permissions(p)
	return json.Marshal(perm)
```
for completeness but I don't think it gets used at all

leeweisberger (approved):

PR #1085, opened 2023-04-20
Add task + view init endpoint

francescaleung:
## Description

Supports ASM-284.

## Test plan

Tested locally with web changes


pkg/initcmd/initcmd.go
```diff
@@ -79,6 +87,23 @@ func (ir InitResponse) GetModifiedFiles() []string {
...
 }
 
+func (ir InitResponse) GetFilenamesWithStatus() []FilenameWithStatus {
+	filenames := []string{}
+	for fn := range ir.filenames {
+		filenames = append(filenames, fn)
+	}
+	slices.SortFunc(filenames, sortFilenames)
+
+	filenamesWithStatus := make([]FilenameWithStatus, len(filenames))
```
colinking, line 97: Minor nit, but I'm tempted to lint against pre-initializing slice lengths (or at least lint against not specifying the length). Would you be against that? haha

It's one of those pre-optimizations that very rarely makes a meaningful difference and can easily be wrong (e.g. calling append on `filenamesWithStatus` does not work unless you switch to `make(..., 0, len(...))`)

colinking: Just found https://github.com/ashanbrown/makezero -- what would you think if I enabled that with `-always`?

pkg/server/apidev/apidev.go
```diff
@@ -60,6 +60,9 @@ func AttachDevRoutes(r *mux.Router, s *state.State) {
 	r.Handle("/logs/{run_id}", handlers.SSE(s, LogsHandler)).Methods("GET", "OPTIONS")
 	r.Handle("/tasks/errors", handlers.New(s, GetTaskErrorsHandler)).Methods("GET", "OPTIONS")
 
+	r.Handle("/tasks/init", handlers.WithBody(s, InitTaskHandler)).Methods("POST", "OPTIONS")
+	r.Handle("/views/init", handlers.WithBody(s, InitViewHandler)).Methods("POST", "OPTIONS")
```
colinking, line 64: Can you call these `/tasks|views/create` to align with airport?

pkg/server/apidev/init.go
```diff
@@ -0,0 +1,138 @@
...
+		DryRun:           req.DryRun,
+		WorkingDirectory: s.Dir,
+		Inline:           true,
+		TaskName:         req.Name,
+		TaskKind:         req.Kind,
+		TaskKindName:     req.KindName,
+		TaskDescription:  req.Description,
+	})
+	if err != nil {
+		return InitTaskResponse{}, errors.Wrap(err, "error initializing task")
```
colinking, line 52:
```suggestion
		return InitTaskResponse{}, errors.Wrap(err, "initializing task")
```

pkg/server/apidev/init.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	var newTask *api.Task
+	if !req.DryRun && resp.NewTaskDefinition != nil {
```
colinking, line 56: Is `NewTaskDefinition` not set during dry run? Any reason not? Seems nice if dry run returns the task.

pkg/server/apidev/init.go
```diff
@@ -0,0 +1,138 @@
...
+		t, err := resp.NewTaskDefinition.GetTask(definitions.GetTaskOpts{
+			Bundle:        true,
+			IgnoreInvalid: true,
+		})
+		if err != nil {
+			return InitTaskResponse{}, errors.Wrap(err, "error converting definition to task")
+		}
+		newTask = &t
+
+		// Wait a maximum of five seconds for the task to be discovered.
```
colinking, line 66: Can you kick off rediscovery for the path? That way we don't need to wait on the file watcher to poll for changes. Specifically, call the debounced method so that it doesn't run twice when the file watcher detects the changes. See: https://github.com/airplanedev/cli/blob/06d66ddc4e611c1441c45df0c6e044ff49fd9299/pkg/server/server.go#L348-L355

pkg/server/apidev/init.go
```diff
@@ -0,0 +1,138 @@
...
+		max := time.Second * 5
+		wg := sync.WaitGroup{}
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			interval := time.Millisecond * 500
```
colinking, line 72: Let's poll much more often, since this is all in-memory. Like every 10ms.

pkg/server/apidev/init.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/airplanedev/cli/pkg/initcmd"
+	"github.com/airplanedev/cli/pkg/server/state"
+	"github.com/airplanedev/cli/pkg/utils"
+	"github.com/pkg/errors"
+)
+
+type InitTaskRequest struct {
+	DryRun bool `json:"dryRun"`
+
+	Name        string              `json:"name"`
```
colinking, line 22: We also need to pass the slug through. We wanted that to be editable in the modal. We can then use that as a the basis for generating a unique slug, in the off chance it isn't unique.

colinking (changes requested):

pkg/initcmd/initcmd.go
francescaleung: no strong feelings about this one, you do you :)

pkg/server/apidev/init.go
francescaleung: In a dry run we don't actually create any files, so a dry run SQL initialization fails when we go to convert the def into a task, even with IgnoreInvalid=true (it tries to read the entrypoint file so it can put the query into KindOptions). That's fixable, but seemed like a lot to do here. Also, we don't even need to really return the whole task; returning just the slug here would be sufficient, & would have the bonus effect of matching what we do for views (because view initialization is a lot dumber & we don't even have a view def there). Thoughts?

pkg/server/apidev/init.go
francescaleung: It's still async, we'd still want to wait. Reason: if you return right away & it hasn't been picked up yet by discovery, we redirect the user to a page that 404s.

pkg/server/apidev/init.go
francescaleung: we'll get there, I'm considering that out of scope for this particular PR

pkg/server/apidev/init.go
```diff
@@ -0,0 +1,146 @@
...
+		// Kick off discovery for the task.
+		path := resp.NewTaskDefinition.GetDefnFilePath()
+		if path == "" {
+			entrypoint, err := resp.NewTaskDefinition.GetAbsoluteEntrypoint()
+			if err != nil {
+				return InitTaskResponse{}, errors.Wrap(err, "getting absolute entrypoint")
+			}
+			path = entrypoint
+		}
+		dfn, ok := s.Debouncers.Get(path)
```
colinking, line 79: The path would never be in `Debouncers` since it's a new path (unless the watcher happened to already pick it up). Add logic to initialize the debouncer if it doesn't exist (maybe just add a state function so that both codepaths can share the logic?).

colinking (approved):

francescaleung:
rerequesting for NodeFlavor + task/view discovery logic moving into state

PR #1086, opened 2023-04-21
Create any necessary directories or files when patching

calufornia:
## Description

The patch instructions may require us to create new folders and files in the local filesystem. Add logic to do so.

## Test plan
Verified locally


pkg/server/files/patch.go
```diff
@@ -39,7 +39,11 @@ func Patch(s *state.State, patch string) error {
 		// TODO: Handle the case when old name and new name differ (i.e. we're renaming files).
 		path := filepath.Join(s.Dir, diff.OldName)
 
-		f, err := os.OpenFile(path, os.O_RDWR, 0755)
+		if err := os.MkdirAll(filepath.Dir(path), 0770); err != nil {
```
ameyakhare, line 42: this will always be in the dev server files root right? since we're appending the "oldname" to s.dir?

ameyakhare (approved):

pkg/server/files/patch.go
calufornia: yep, precisely

PR #1089, opened 2023-04-24
Fix cached builds not discovering entities

ameyakhare:
## Description
Introduce a build arg BUILD_ID into the bundle build dockerfile. We'll set this buildarg to the build id so that all the steps that occur underneath the ARG command cannot be cached, since the build id would be unique: https://stackoverflow.com/questions/35134713/disable-cache-for-specific-run-commands

The next step will be adding this build arg in the bundler (airport) when it executes the build.

## Test plan
Builds still work properly.


pkg/build/views/view.go
```diff
@@ -367,6 +367,9 @@ func ViewBundle(root string, buildContext buildtypes.BuildContext, options build
 		{{end}}
 
 		{{if .FilesToDiscover}}
+		ARG BUILD_ID
+		RUN echo "$BUILD_ID"
```
leeweisberger, line 371: This is going to kill the build cache which is below this line... but I guess we have to do that to discover the view every time...

pkg/build/views/view.go
```diff
@@ -367,6 +367,9 @@ func ViewBundle(root string, buildContext buildtypes.BuildContext, options build
 		{{end}}
 
 		{{if .FilesToDiscover}}
+		ARG BUILD_ID
```
leeweisberger, line 370: This search seems to indicate that this will produce an error if the arg is not supplied...

Is this true? if so this should be step 2.
image

leeweisberger (changes requested):

pkg/build/views/view.go
ameyakhare: yeah we kinda have to do this...

pkg/build/views/view.go
ameyakhare: take a look at the link. all the blocks before the arg statement are not cached, since the intermediate layer is cached. we'll have to be careful when we bump this in airport, but it is a one time check to pass in the deployment id in the build args in the builder/bundler.

leeweisberger (approved):
Let's pass the arg in from airport first before we merge this

PR #1091, opened 2023-04-25
Resource interpolation in studio

kellymtrinh:


pkg/dev/expressions.go
```diff
@@ -0,0 +1,93 @@
...
+		return nil, err
+	}
+	updatedRes := map[string]resources.Resource{}
+	for slug, r := range interpolatedAliasMap {
+		mapRes, ok := r.(map[string]interface{})
+		if !ok {
+			return nil, errors.New("expected resource to be a map")
+		}
+		resourceKind, err := getResourceKind(rawRes, slug)
+		if !ok {
```
calufornia, line 80:
```suggestion
		if err != nil {
```

pkg/server/apiext/apiext.go
```diff
@@ -9,7 +9,7 @@ import (
 	libapi "github.com/airplanedev/cli/pkg/api"
 	api "github.com/airplanedev/cli/pkg/api/cliapi"
 	libhttp "github.com/airplanedev/cli/pkg/api/http"
-	"github.com/airplanedev/cli/pkg/dev"
+	dev "github.com/airplanedev/cli/pkg/dev"
```
calufornia, line 12: Same here

cmd/airplane/root/initcmd/download.go
```diff
@@ -7,7 +7,7 @@ import (
 	"os"
 	"path/filepath"
 
-	"github.com/airplanedev/cli/pkg/build/clibuild"
+	build "github.com/airplanedev/cli/pkg/build/clibuild"
```
calufornia, line 10: Why this change?

pkg/dev/dev.go
```diff
@@ -174,9 +173,14 @@ func (l *LocalExecutor) Execute(ctx context.Context, config LocalRunConfig) (api
 	}
 
 	baseInterpolateRequest := baseEvaluateTemplateRequest(config, configVars)
-
+	interpolatedRes, err := interpolateResource(ctx, config.RemoteClient, baseInterpolateRequest, config.AliasToResource)
+	if err != nil {
+		return api.Outputs{}, errors.Wrap(err, "failed to interpolate resources")
+	}
+	config.AliasToResource = interpolatedRes
```
calufornia, line 180: How come we need this line? I see that we're already setting the resources in the base interpolation request to the interpolated resources below, so at this point we're "done using" `config.AliastoResource`

cmd/airplane/root/initcmd/download.go
kellymtrinh: I had trouble compiling this - it was automatic when I saved and vscode ran lint?

pkg/dev/dev.go
kellymtrinh: The cfg is used further down- in `getEnvVars` and later `getBuiltInTaskEnvVars` or in the future, so I think keeping resources up to date makes sense otherwise downstream users will need to know that cfg. AliastoResource isn't interpolated

cmd/airplane/root/initcmd/download.go
kellymtrinh: This was the error I kept seeing: https://github.com/airplanedev/cli/pull/1091/commits/1e1ca8058f84365a8add8e8e2391d78d37b3fa28
But let me revert this and see if it still complains

calufornia (approved):

PR #1094, opened 2023-04-26
Handle concurrency key and limits when parsing tasks

zhan-airplane:
## Description
Allows defining concurrency key and limits inline in node/python tasks

## Test plan
Tested deploying a task locally while setting concurrency key/limit (as well as no concurrency key/limit)

---
cc @yolken-airplane


pkg/deploy/taskdir/definitions/task_def.go
```diff
@@ -39,6 +39,8 @@ type Definition struct {
 	RequireRequests    bool                  `json:"requireRequests,omitempty"`
 	AllowSelfApprovals DefaultTrueDefinition `json:"allowSelfApprovals,omitempty"`
 	RestrictCallers    []string              `json:"restrictCallers,omitempty"`
+	ConcurrencyKey     string                `json:"concurrencyKey,omitempty"`
```
calufornia, line 42: How come this isn't a pointer?

calufornia (approved):

pkg/deploy/taskdir/definitions/task_def.go
zhan-airplane: It's also `string` on the airport side - I don't want to have to support too much logic for sorting between empty strings and null, except in update where we need to do patch semantics

PR #1096, opened 2023-04-26
Add endpoints to check for slug availability

francescaleung:
## Description

Also adds checks for unique slug in task/view create endpoints. Note that the endpoints for slug availability currently do not take an entity id as the API endpoints do. Discussed somewhat in ASM-413.

Supports ASM-413. Depends on https://github.com/airplanedev/airport/pull/4487.

## Test plan

Tested locally with corresponding web changes.


pkg/server/apidev/init.go
```diff
@@ -43,12 +44,20 @@ func InitTaskHandler(ctx context.Context, s *state.State, r *http.Request, req I
 		return InitTaskResponse{}, libhttp.NewErrBadRequest("missing kind")
 	}
 
+	// Check for slug uniqueness
+	if ok, err := IsTaskSlugAvailable(ctx, s, req.Slug); err != nil {
+		return InitTaskResponse{}, errors.Wrap(err, "checking if slug is available")
+	} else if !ok {
+		return InitTaskResponse{}, libhttp.NewErrBadRequest("slug taken")
```
colinking, line 51: Let's provide more user-friendly error messages. Same for Views, below.

```suggestion
		return InitTaskResponse{}, libhttp.NewErrBadRequest("This slug is already used by another task. Please choose  a unique slug.")
```

colinking (approved):

PR #1097, opened 2023-04-26
Add permissions to JS and python parsers

rose-huang:
## Description
Adding permissions to parsers. Will add python tests later

## Test plan
discover tests with permissions pass


leeweisberger (approved):
I'm less confident with the syntax in the python parser, but I think it lgtm

PR #1100, opened 2023-04-27
Add default run permissions to task def

leeweisberger:
## Description

Add default run permissions to task definition.

Add default run permissions to json schema.

Parse default run permissions from inline tasks (todo to add to sdk)

Add default run permissions when initializing a task

## Test plan
Added unit tests for each of the steps. have not yet tested e2e


pkg/deploy/taskdir/definitions/yaml_templates.go
```diff
@@ -51,6 +52,9 @@ name: {{.name}}
 # this task. Default: true.
 # allowSelfApprovals: false
 
+# Manage who can view new runs of this task. Default: "task-viewers".
```
rose-huang, line 55: do we add everything to this template? just wondering since schedules / permissions / restrict. task callers arent here

pkg/deploy/discover/parser/python/parser.py
```diff
@@ -62,17 +62,24 @@ class Def:
     restrictCallers: Optional[List[Literal["task", "view"]]]
     timeout: int
     runtime: Literal["standard", "workflow"]
-    concurrencyKey: str
-    concurrencyLimit: int
 
-    schedules: Dict[str, Schedule]
+
+<< << << < HEAD
```
rose-huang, line 67: can you fix this merge conflict

pkg/runtime/python/python.go
```diff
@@ -342,6 +342,9 @@ import airplane
         ),
     {{- end}}
     ],
+    {{- end}}
```
rose-huang, line 345: what is this doing?

rose-huang (changes requested):

pkg/deploy/discover/parser/python/parser.py
leeweisberger: woah

pkg/deploy/taskdir/definitions/yaml_templates.go
leeweisberger: Historically we've added everything, but I think we forgot some things. This field is pretty advanced... I'm going to go ahead and delete it. i also started a thread in slack

pkg/runtime/javascript/javascript.go
```diff
@@ -54,23 +54,23 @@ var code = template.Must(template.New("js").Parse(`{{with .Comment -}}
...
-
-	// Sort the data in ascending order by name.
-	data.sort((u1, u2) => {
-		return u1.name.localeCompare(u2.name);
-	});
-
-	// You can return data to show output to users.
-	// Output documentation: https://docs.airplane.dev/tasks/output
-	return data;
+  const data = [
```
leeweisberger, line 57: Calling out that this is a real change. It replaces tabs with spaces which is standard for js. Without this, our js tasks are ill formatted.

pkg/runtime/python/python.go
leeweisberger: This is a bad diff. I added the end after the if. The diff is showing the one before

rose-huang (approved):

pkg/runtime/javascript/javascript.go
```diff
@@ -97,142 +97,145 @@ var inlineCode = template.Must(template.New("ts").Funcs(template.FuncMap{
...
+          {{- end}}
+        }
+        {{- end}}
+      }
+    {{- end}}
+    },
+    {{- end}}
+    {{- if and (ne .DefaultRunPermissions "task-viewers") (.DefaultRunPermissions) }}
+    defaultRunPermissions: "{{.DefaultRunPermissions}}",
+    {{- end}}
```
colinking, line 198: If it wasn't for the generated comments, we could get rid of this `Generate` logic and replace it with the new updaters. Though we could probably get comment generation working to make that happen... 🤔

pkg/runtime/javascript/javascript_test.go
```diff
@@ -135,3 +138,56 @@ func TestVersion(t *testing.T) {
 		})
 	}
 }
+func TestGenerateInline(t *testing.T) {
```
colinking, line 141: 🔥

colinking (approved):
Sheesh, you weren't kidding! So many places to update.

pkg/runtime/javascript/javascript.go
leeweisberger: We could also just add the comments to the jsdoc and be good with that

PR #1103, opened 2023-04-28
[RFC] Add setting to airplane.yaml for not updating tsconfig

zhan-airplane:
## Description
Context: https://linear.app/airplane/issue/ASM-532/pave-seeing-undesired-tsconfigjson-changes

Somewhat unsure how opinionated we need to be about `tsconfig.json` files in general, but a thought I had is that we could just add a flag to `airplane.yaml` that skips the `tsconfig.json` update, since even if we decide we need to be opinionated about some of them, people might still continue complaining down the road. Thus, opening this PR as a proposal

## Test plan
Ran locally


pkg/initcmd/task.go
```diff
@@ -787,13 +787,15 @@ func runKindSpecificInstallation(ctx context.Context, req runKindSpecificInstall
...
 
 		if filepath.Ext(entrypoint) == ".ts" || filepath.Ext(entrypoint) == ".tsx" {
-			// Create/update tsconfig in the same directory as the package.json file
-			tsConfigCreated, err := node.CreateTaskTSConfig(packageJSONDir, req.Prompter, req.Logger, req.DryRun)
-			if err != nil {
-				return err
-			}
-			if req.InitResponse != nil {
-				req.InitResponse.AddFile(tsConfigCreated, filepath.Join(packageJSONDir, "tsconfig.json"))
+			if !finalAirplaneConfig.Javascript.UseExistingTSConfig {
```
calufornia, line 790: What if the user has this flag set to true but doesn't have a tsconfig? Do we just consider that user error?

calufornia (approved):
This approach makes sense to me.

leeweisberger (changes requested):
I actually don't think this is necessary and think we should just change our logic to be less opinionated (if there is a tsconfig, only update the things that must be updated).

 While this change is not wrong per se, I don't think this config option really makes sense and I don't think we should expect users to have to set this. It's also pretty different from the other options in the airplane.yaml file

PR #1105, opened 2023-05-01
Fix omitempty behavior for concurrencyLimit/timeout

colinking:
## Description

When updating a task via studio, `concurrencyLimit` was getting set to `1` in the updated task definition. This generates errors for JS/Py tasks if the SDK hasn't been updated to latest version. This is also unnecessary because `1` is the default value. The same issue existed for `timeout`, but we had special logic in the updaters for not serializing `timeout` if the value was `3600`.

This PR fixes the omitempty behavior for both fields so that they are omitted when serialized if set to their default values. It also reworks the definition serialization test to cover these cases.

## Test plan

Unit tests pass.


colinking:
* main
  * **PR #1105** <a href="https://app.graphite.dev/github/pr/airplanedev/cli/1105" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/cli/1105?utm_source=stack-comment).

pkg/runtime/javascript/updater/update.ts
```diff
@@ -109,10 +109,6 @@ const buildTaskConfig = (def: any): ExpressionKind => {
 
     delete def.node;
   }
-  // Apply a default value to the timeout field.
-  if (def.timeout === 0 || (def.runtime !== "workflow" && def.timeout === 3600)) {
-    delete def.timeout;
-  }
```
colinking, line 115: This is no longer necessary since the definition will properly remove `timeout` if it's set to the default value.

pkg/runtime/updaters/fixtures/all_defaults.out.task.yaml
```diff
@@ -2,3 +2,4 @@ slug: my_task
 node:
   entrypoint: ./entrypoint.js
   nodeVersion: "18"
+permissions: team_access
```
colinking, line 5: FYI @rose-huang I added tests that cover the new permissions fields while I was here. Other fields (like concurrencyLimit) don't get written if they are set to their default value, but it seems permissions should get written even if it's the default value since that "locks" permission editing to code. LMK if that doesn't seem right.

Also FYI @leeweisberger, you'll want to add similar tests for default run permissions.

pkg/deploy/taskdir/definitions/task_def_test.go
```diff
@@ -10,130 +10,157 @@ import (
...
-parameters:
-- name: Name
-  slug: name
-  type: shorttext
-  description: Someone's name.
-  default: World
-  required: true
-resources:
-  db: demo_db
+func TestDefinitionMarshal(t *testing.T) {
```
colinking, line 13: Rewrote these tests to be more extensive / easier to read.

colinking:
Just noticed @zhan-airplane is OOO. @leeweisberger mind if I send by you since you wanted to get context? We can chat about this tomorrow and you can stamp based on that.

leeweisberger:
> Just noticed @zhan-airplane is OOO. @leeweisberger mind if I send by you since you wanted to get context? We can chat about this tomorrow and you can stamp based on that.

sounds good!

leeweisberger (approved):

PR #1108, opened 2023-05-01
Move updaters into definitions subdir

colinking:
## Description

The Views definition updater will use the same setup as the JavaScript task updater, which currently lives in the (task-specific) runtime. I extracted the updaters into `pkg/definitions/updaters` which is not task-specific. I'll add Views here in a follow-up PR.

There are no functional code changes beyond moving and renaming.

## Test plan

Tests should pass.

PR #1107, opened 2023-05-01
Move definitions folder out of taskdir

colinking:
## Description

The definitions folder includes the View definition. Therefore, it doesn't make sense for it to live in the `taskdir` package.

This just moves the code, it makes no other changes.

## Test plan

Tests should still pass.

PR #1110, opened 2023-05-02
Add marshal method to view definition

colinking:
## Description

This PR adds a `Marshal` method to `ViewDefinition` which aligns with what we have on the task `Definition` struct. This will be used in a follow-up PR where we add logic for updating View definitions.

This PR also makes the `name` field optional, which matches the behavior of tasks. The JS SDK will automatically convert slugs into humanized strings and therefore already doesn't require a `name` field. I tested out what happens when you deploy a YAML task without a name and noticed that we don't handle it well (e.g. in the deploy logs or in the deploy UI it shows an empty string). Since this is an issue with tasks and only applies to YAML definitions, I'm punting on that for now (see https://linear.app/airplane/issue/CAP-2533/generate-names-for-tasksviews-without-names).

## Test plan

Tests pass.


colinking:
* main
  * **PR #1107** <a href="https://app.graphite.dev/github/pr/airplanedev/cli/1107" target="_blank">Graphite</a> 
    * **PR #1108** <a href="https://app.graphite.dev/github/pr/airplanedev/cli/1108" target="_blank">Graphite</a> 
      * **PR #1110** <a href="https://app.graphite.dev/github/pr/airplanedev/cli/1110" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/cli/1110?utm_source=stack-comment).

pkg/definitions/view_def.go
```diff
@@ -62,6 +73,39 @@ func (d *ViewDefinition) Unmarshal(format DefFormat, buf []byte) error {
 	return nil
 }
 
+func (d ViewDefinition) Marshal(format DefFormat) ([]byte, error) {
```
colinking, line 76: This is the exact same logic as for task definitions.

pkg/definitions/view_def.go
```diff
@@ -13,13 +13,24 @@ import (
...
-	Entrypoint   string               `json:"entrypoint"`
-	EnvVars      api.EnvVars          `json:"envVars,omitempty"`
-	DefnFilePath string               `json:"defnFilePath,omitempty"`
+	Slug        string      `json:"slug"`
+	Name        string      `json:"name,omitempty"`
+	Description string      `json:"description,omitempty"`
+	Entrypoint  string      `json:"entrypoint"`
+	EnvVars     api.EnvVars `json:"envVars,omitempty"`
+	// DefnFilePath is the absolute path to this View definition, if one exists.
+	DefnFilePath string               `json:"-"`
```
colinking, line 33: This will no longer be serialized. AFAICT, this field is only read within the CLI for Studio and we never relied on it being serialized as JSON.

pkg/definitions/view_def.go
```diff
@@ -13,13 +13,24 @@ import (
...
+// `pkg/definitions/updaters` when adding new fields).
 type ViewDefinition struct {
-	Name         string               `json:"name"`
-	Slug         string               `json:"slug"`
-	Description  string               `json:"description,omitempty"`
-	Entrypoint   string               `json:"entrypoint"`
-	EnvVars      api.EnvVars          `json:"envVars,omitempty"`
-	DefnFilePath string               `json:"defnFilePath,omitempty"`
+	Slug        string      `json:"slug"`
+	Name        string      `json:"name,omitempty"`
```
colinking, line 28: Switched the order so that slug is serialized before name, matching tasks.

colinking, line 28: Made name `omitempty`, so it no longer is serialized if empty. I also updated the JSON schema to not require it.

leeweisberger (approved):
The name thing does rely on you using a version of the js sdk that supports the auto-naming. That's been around for a bit though so it's probably safe

PR #1114, opened 2023-05-02
Don't locally execute tasks with errors

francescaleung:
## Description

We prevent users from executing tasks directly, but we don't do anything to redirect child runs to the remote client.

## Test plan

Tested locally with a task that called a Docker task that exists locally.


pkg/server/apiext/tasks.go
```diff
@@ -75,7 +75,13 @@ func ExecuteTaskHandler(ctx context.Context, state *state.State, r *http.Request
 
 	localTaskConfig, ok := state.TaskConfigs.Get(req.Slug)
 	isBuiltin := builtins.IsBuiltinTaskSlug(req.Slug)
-	if !isBuiltin && !ok {
+	taskErrors, err := state.GetTaskErrors(ctx, req.Slug, pointers.ToString(envSlug))
+	// If the task isn't local, GetTaskErrors will always fail. Check the error only if the task is
+	// local.
```
colinking, line 80: What about built-ins? Maybe nest this `GetTaskErrors` call in `if ok {}` to make this more clear.

PR #1115, opened 2023-05-02
Remove dependency on jsdom

leeweisberger:
## Description

We initially added JSDOM to all tasks, both in the parser and build stage, to cover the case where a task and view are in the same file. When this happened (in the past), the task would fail because the view was not node-safe (It referenced things like document which could not be run in Node).

Since views 2.0, views is now node-safe so we no longer need JSDOM. We do still need to keep it around (unfortunately) for users using an old version of views. In this pr, we check to see what version of views is being used (if any) and then conditionally add JSDOM.

JSDOM has a few issues:
1. It doesn't work with `type: "module"`. This pr does not get us to `type: "module"` compatibility, but it is necessary to get there.
2. It's a big dependency that we're adding to every task image
3. It causes bugs in some cases. e.g. a user's library was tricked into thinking it was in a browser and failed because of our jsdom inclusion.

**Note**
There is a potential breaking case here - if someone defined a task and view in the same file and their view (or a library it imports) is not node-safe, it would have worked in the past but now will not work. They'll either have to split up their task and view, or add something like jsdom themselves. I think this becomes the user's burdon and I'm pretty confident we shouldn't just be doing this every time.

## Test plan
I ran `airplane dev` for a task/view combo using the latest version of views
I ran `airplane dev` for a task/view combo using views 1.5 and it still worked
I ran airplane deploy with the same 2 cases ^
We have unit tests that cover a task and a view together.
We also have integration tests that cover a task/view together which will be triggered when this merges.

CAP-2411


pkg/build/node/node.go
```diff
@@ -709,7 +711,10 @@ func GenShimPackageJSON(opts GenShimPackageJSONOpts) ([]byte, error) {
 		shimDeps = append(shimDeps, workflowRuntimePkg)
 	}
 	if opts.IsBundle {
-		buildDeps = append(buildDeps, []string{"esbuild", "esbuild-plugin-tsc", "typescript", "jsdom"}...)
+		buildDeps = append(buildDeps, []string{"esbuild", "esbuild-plugin-tsc", "typescript"}...)
+		if opts.NeedJSDOM {
```
leeweisberger, line 715: we're now conditionally installing jsdom

pkg/build/node/node.go
```diff
@@ -796,6 +801,38 @@ func TemplatedNodeShim(params NodeShimParams) (string, error) {
...
+	}
+	ver, err := semver.ParseTolerant(viewsVersion)
+	if err != nil {
+		return false
+	}
+	// Add JSDOM if views is < 2.0.0. Before this version, views was not node-friendly so we need to add JSDOM to ensure
+	// that tasks and views work in the same file.
+	return ver.LT(semver.Version{Major: 2})
+}
+func TemplatedUniversalNodeShim(params UniversalNodeShimParams) (string, error) {
```
leeweisberger, line 823: The universal node shim only includes jsdom if it's necessary.

pkg/build/node/node.go
```diff
@@ -1148,8 +1195,14 @@ func NodeBundle(
 		if err := os.MkdirAll(path.Dir(parserPath), 0755); err != nil {
 			return "", errors.Wrapf(err, "creating parser file")
 		}
-		if err := os.WriteFile(parserPath, []byte(parser.NodeParserScript), 0755); err != nil {
-			return "", errors.Wrap(err, "writing parser script")
+		if needJSDOM {
```
leeweisberger, line 1198: The parser includes jsdom, only if it's necessary

pkg/build/node/node_externals.go
```diff
@@ -471,7 +471,8 @@ func getYarnLockPackageVersion(dir string, packageName string) (string, error) {
 
 	for _, tree := range results.Data.Trees {
 		if strings.HasPrefix(tree.Name, fmt.Sprintf("%s@", packageName)) {
-			components := strings.SplitN(tree.Name, "@", 2)
+			name := strings.TrimPrefix(tree.Name, "@") // If the package starts with an @, remove it.
+			components := strings.SplitN(name, "@", 2)
```
leeweisberger, line 475: This was a bug. If the package started with `@` like `@airplane/views` then the incorrect version was returned. I added a unit test for this

pkg/build/node/nodetest/node_test.go
```diff
@@ -1053,6 +1054,22 @@ func TestGenShimPackageJSON(t *testing.T) {
...
+			desc:        "Yarn workspace with no bundle shim dependencies, don't need jsdom",
+			packageJSON: "typescript/yarnworkspacesnoairplane/package.json",
+			isWorkflow:  true,
+			isBundle:    true,
+			needJSDOM:   false,
+			expectedShimPackageJSON: node.ShimPackageJSON{
+				Dependencies: map[string]string{
+					"airplane":                   buildToolsPackageJSON.Dependencies["airplane"],
+					"@airplane/workflow-runtime": buildToolsPackageJSON.Dependencies["@airplane/workflow-runtime"],
+					"esbuild":                    buildToolsPackageJSON.Dependencies["esbuild"],
```
leeweisberger, line 1067: We no longer install jsdom

pkg/build/views/viewstest/view_test.go
```diff
@@ -129,6 +129,27 @@ func TestViewBundleBuilder(t *testing.T) {
 				"myView.airplane.tsx",
 			},
 		},
+
+		// Uncomment when the inline parser supports type module.
```
leeweisberger, line 133: This test still doesn't pass. I left it around so we can uncomment it in the future.

pkg/deploy/discover/build.go
```diff
@@ -138,7 +148,7 @@ func extractJSConfigs(file string, env []string) (ParsedJSConfigs, error) {
 }
 
 func extractPythonConfigs(file string, env []string) ([]map[string]interface{}, error) {
-	parserCmd := exec.Command("python3", "-c", string(pythonParserScript), file)
+	parserCmd := exec.Command("python3", "-c", parser.PythonParserScript, file)
```
leeweisberger, line 151: Unrelated - just reusing an `embed` point

pkg/deploy/discover/parser/node/main.go
```diff
@@ -0,0 +1,90 @@
...
+	"strings"
+
+	_ "embed"
+
+	"github.com/airplanedev/cli/pkg/build/utils"
+	"github.com/airplanedev/cli/pkg/utils/pointers"
+	esbuild "github.com/evanw/esbuild/pkg/api"
+)
+
+func build(file string, out string) {
```
leeweisberger, line 15: I had to reimplement this in go from javascript so we could use go's templating system. We create two versions of parser.js - one with and one without jsdom

pkg/deploy/discover/parser/node/parser-with-jsdom.ts
```diff
@@ -0,0 +1,142 @@
+// Only `parser.ts` should be changed. After changing it, run `go run main.go` to build the other parser files.
```
leeweisberger, line 1: This file is autogenerated

pkg/deploy/discover/parser/node/parser-without-jsdom.ts
```diff
@@ -0,0 +1,138 @@
+// Only `parser.ts` should be changed. After changing it, run `go run main.go` to build the other parser files.
```
leeweisberger, line 1: This file is autogenerated

pkg/runtime/javascript/javascript.go
```diff
@@ -404,7 +404,13 @@ func (r Runtime) PrepareRun(
 		return nil, nil, err
 	}
 
-	shim := node.UniversalNodeShim
+	needJSDOM := node.NeedJSDOM(root)
```
leeweisberger, line 407: We use the same logic to build the node shim template when devving as when deploying

.github/workflows/javascript.yml
```diff
@@ -19,7 +19,6 @@ jobs:
       matrix:
         dir:
           - { name: "updater", value: "pkg/definitions/updaters/javascript" }
-          - { name: "discovery", value: "pkg/deploy/discover/parser/node" }
```
ameyakhare, line 22: why was this removed?

pkg/build/node/universal-node-shim.js
```diff
@@ -1,10 +1,14 @@
...
 import airplane from "airplane";
+{{ if .AddJSDOM -}}
 import { JSDOM } from "jsdom";
+{{ end -}}
 
+{{ if .AddJSDOM -}}
 // Replace the global "document" with JSDOM to mock out browser APIs.
 const dom = new JSDOM(`<!DOCTYPE html><body></div></body>`);
 global.document = dom.window.document;
+{{ end -}}
```
ameyakhare, line 11: put this in the same "if" block

pkg/deploy/discover/build.go
```diff
@@ -101,13 +102,22 @@ type ParsedJSConfigs struct {
...
 	defer os.Remove(tempFile.Name())
-	_, err = tempFile.Write(nodeParserScript)
+
+	needJSDOM := node.NeedJSDOM(rootDir)
+	var nodeParserScript string
+	if needJSDOM {
+		nodeParserScript = parser.NodeParserScript
+	} else {
+		nodeParserScript = parser.NodeParserScriptWithoutJSDOM
+	}
```
ameyakhare, line 118: can we make this a function getParserScript(needsJSDOM). this if condition is duplicated in a couple of places

ameyakhare (commented):
im a little confused - why can we remove jsdom if we still run discovery for the view configuration itself, wont some of the packages people use potentially not be compatible with node? is that really their fault? do we make this an explicit requirement anywhere?

PR #1116, opened 2023-05-03
Rename CI step for updaters to avoid collision

colinking:
## Description

The JS and Py updaters currently have the same name (since branch protection rules do not look at the workflow name, only the job name), so we can't require they both pass.

## Test plan

Confirm the names are updated in CI below. I'll update the protection rules before+after merging.


colinking:
* main
  * **PR #1116** <a href="https://app.graphite.dev/github/pr/airplanedev/cli/1116" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/cli/1116?utm_source=stack-comment).

francescaleung (approved):

PR #1117, opened 2023-05-03
Fix waiting for user logic

leeweisberger:
## Description
Waiting for user always returns true... I think we messed up the boolean check?

## Test plan
Manually tested that when I complete a prompt, wait for user returns false.


pkg/server/apiint/apiint.go
```diff
@@ -439,7 +439,7 @@ func SubmitPromptHandler(ctx context.Context, state *state.State, r *http.Reques
 				// Check if the run is still waiting for user input.
 				run.IsWaitingForUser = false
 				for _, prompt := range run.Prompts {
-					run.IsWaitingForUser = run.IsWaitingForUser || (prompt.SubmittedAt == nil || prompt.CancelledAt == nil)
+					run.IsWaitingForUser = run.IsWaitingForUser || (prompt.SubmittedAt == nil && prompt.CancelledAt == nil)
```
leeweisberger, line 442: Once of submitted at and cancelled at will always be null causing this to always be true right?

ericliuche: ah yeah, good catch

ericliuche (approved):

PR #1119, opened 2023-05-03
Add .airplane to gitignore

calufornia:
## Description

Adds .airplane to the default gitignore, since this should also not be checked in. I also realized we add a gitignore in the view creation path, but not the task creation path, so adding it to the latter.

Resolves ASM-536

## Test plan
Verified locally


pkg/testutils/compare.go
```diff
@@ -26,9 +26,19 @@ func includeFunc(dir string) func(string, os.FileInfo) (bool, error) {
 	}
 }
 
+func compareDirectories(
```
calufornia, line 29: I fixed this comparison so that it tests strict equality between directories.

francescaleung (approved):

PR #1120, opened 2023-05-03
Prevent browser from opening on ctrl+c on windows

calufornia:
## Description

Exiting `airplane dev` currently causes the browser to open in Windows. This is because the interrupt signal is getting captured by `fmt.Scanln()`. Furthermore, we always advance past `fmt.Scanln()` _before_ the stop channel receives a notification of an interrupt signal. This PR adds an artificial timeout that allows the stop channel to get notified of the interrupt signal before continuing on to open the browser.

cc @yolken-airplane 

Resolves ASM-20

## Test plan
Verified that the browser-opening issue no longer occurs on windows.


cmd/airplane/tasks/dev/server.go
```diff
@@ -264,6 +265,17 @@ func runLocalDevServer(ctx context.Context, cfg taskDevConfig) error {
...
 			fmt.Scanln()
+			if runtime.GOOS == "windows" {
+				// This is a hack so that Ctrl+C on Windows doesn't cause the browser to open. Ctrl+C is being captured
+				// by the fmt.Scanln() above, and furthermore it is being captured _before_ the stop channel gets
+				// notified of a termination signal. We add an artificial timeout here to give the stop channel a chance
+				// to get notified first.
+				select {
+				case <-stop:
+					return
+				case <-time.After(100 * time.Millisecond):
```
calufornia, line 276: This is quite hacky, but consistently prevents the browser from opening on exit and also does not noticeably increase the time it takes between the user pressing enter and the browser opening.

calufornia:
I'm not entirely sure why `fmt.Scanln` is capturing ctrl+c—I think somewhere, the windows interrupt signal is getting converted into a new line. Using bufio.NewScanner doesn't help here either.

cmd/airplane/tasks/dev/server.go
```diff
@@ -264,6 +265,17 @@ func runLocalDevServer(ctx context.Context, cfg taskDevConfig) error {
 		logger.Log("")
 		go func() {
 			fmt.Scanln()
```
colinking, line 267: Can you inspect the output of `fmt.Scanln` to check if it's a ctrl-c? 

Based on https://pkg.go.dev/fmt#Scanln -- you might be able to tell this based on one of the return values, or by passing in an arg to get whatever they typed in.

colinking (changes requested):
Not against this, but curious if there's a better approach.

Also your PR is still in draft, did you mean that?

cmd/airplane/tasks/dev/server.go
calufornia: Yep, I've done that—it's nothing. If you print out how many bytes are written, it's 0—no error either

colinking: You mentioned `bufio.NewScanner`, but have you tried `bufio.NewReader().ReadBytes('\n')`? Based on the docs https://pkg.go.dev/bufio#Reader.ReadBytes, it sounds like you'll get an error back if the bytes do not end in a newline (which I assume will be the case when ctrl-c).

calufornia: 😮 I think you might be on to something—`ReadBytes()` seems to return an error when it encounter ctrl+c

calufornia: Updated with the new approach—thanks!

cmd/airplane/tasks/dev/server.go
```diff
@@ -263,7 +264,14 @@ func runLocalDevServer(ctx context.Context, cfg taskDevConfig) error {
 		logger.Log("Press ENTER to open the studio in the browser.")
 		logger.Log("")
 		go func() {
-			fmt.Scanln()
+			// Use bufio.NewReader to properly handle interrupts on Windows. fmt.Scanln() does not return an error
+			// when interrupted on Windows, and so there's no way to distinguish between a user pressing enter vs
+			// Ctrl+C here.
```
colinking, line 269: FYI there's a comment above about `fmt.Scanln` that should be tweaked.

colinking (approved):
Nice!

PR #1121, opened 2023-05-03
Remove views version calculation

leeweisberger:
## Description
At some point we attempted to get the version of views and save it to a file to be read in later.

1. Not everybody uses yarn, so this calculation is incorrect which causes errors.
2. We are never using the result of this calculation.
3. We're getting some errors downstream because of this line.

Since we aren't using it and it causes issues, I'm removing this calculation.


rose-huang (changes requested):
should we remove this line from ViewBundle as well?

leeweisberger:
> should we remove this line from ViewBundle as well?

yes... yes

rose-huang (approved):

PR #1126, opened 2023-05-05
Remove old tasks/views when slug or file is changed

ameyakhare:
Resolves ASM-404

## Description
Fixes the issue of duplicate slugs on save by deleting all entities that existed in the old file path. Now if you change a task or view slug, it doesn't duplicate. This also works for file renames and deletions.

## Test plan
https://user-images.githubusercontent.com/1912084/236525313-a507f600-bbed-41c0-a5b6-15df7066d211.mov


calufornia (approved):

gustavopch:
Related to this, when you delete a task and update the Studio, it starts to show notification errors because the URL is pointing to a task that doesn't exist anymore. Would be better to just redirect to some other valid URL.

ameyakhare:
> Related to this, when you delete a task and update the Studio, it starts to show notification errors because the URL is pointing to a task that doesn't exist anymore. Would be better to just redirect to some other valid URL.

we'll look into how to fix this

PR #1129, opened 2023-05-05
Only prompt on required tsconfig changes

calufornia:
## Description

Makes it so that the user is prompted only to modify their existing tsconfig only if they don't have certain required fields. This helps reduce noise for teams that don't want their tsconfig to be updated, and in general makes us less opinionated about what fields need to be inside the tsconfig. For now, the only configs that we require (for views) are "jsx" and "lib". cc @zhan-airplane 

If no tsconfig.json exists, the logic to write the default one remains the same.

Note that if a user has an invalid value in their tsconfig (let's say `"jsx": "blah"`), this doesn't fix it for the user as we're only checking that required properties _exist_.

I removed the logic to prefer the default TS config vs the existing TS config since we now always prefer the existing TS config no matter what.

I also added a test since this was previously untested. 

Resolves ASM-532

## Test plan
I developed + deployed a TS task, view, and template with an empty tsconfig and verified that everything worked. I verified that I'm only prompted to make a tsconfig update when there's a missing required field.


pkg/node/node.go
```diff
@@ -351,12 +345,34 @@ func mergeMapsRecursively(dest, src map[string]interface{}) {
 				dest[key] = map[string]interface{}{}
 			}
 			mergeMapsRecursively(dest[key].(map[string]interface{}), subMap)
+		} else if list, isList := value.([]interface{}); isList {
```
calufornia, line 348: I added logic here to merge lists in the tsconfig.

pkg/node/node.go
```diff
@@ -351,12 +345,34 @@ func mergeMapsRecursively(dest, src map[string]interface{}) {
...
 }
 
+func contains(list []interface{}, item interface{}) bool {
+	for _, listItem := range list {
+		if listItem == item {
+			return true
+		}
+	}
+	return false
+}
```
leeweisberger, line 374: image
https://stackoverflow.com/a/10486196

leeweisberger (approved):
I think this looks great!

leeweisberger:
Just making sure, if I have additional fields in my `lib` array, I'm good right? No diff? If I'm missing then I get a diff?

calufornia:
> Just making sure, if I have additional fields in my `lib` array, I'm good right? No diff? If I'm missing then I get a diff?

Correct, if I'm understanding the question. You only get a diff if you don't have the values in `requiredViewTSConfigJSON`. So for example, if you were creating a view and you currently had `"lib": ["DOM", "DOM.Iterable", "ESNext", "ESNextBlah"]` you wouldn't get a diff, but if you only had `"lib": ["DOM", "ESNext", "ESNextBlah"]` then you would get a diff

calufornia:
> > Just making sure, if I have additional fields in my `lib` array, I'm good right? No diff? If I'm missing then I get a diff?
> 
> Correct, if I'm understanding the question. You only get a diff if you don't have the values in `requiredViewTSConfigJSON`. So for example, if you were creating a view and you currently had `"lib": ["DOM", "DOM.Iterable", "ESNext", "ESNextBlah"]` you wouldn't get a diff, but if you only had `"lib": ["DOM", "ESNext", "ESNextBlah"]` then you would get a diff

This should also be covered by the unit test

PR #1134, opened 2023-05-09
Support permissions as code

leeweisberger:
## Description

Support permissions as code in the studio editor. We're basically adding permissions to all of our structs and transforming them correctly.

## Test plan
Added some unit tests, but also did a good amount of manual testing
See loom https://www.loom.com/share/e533afd9cc654bef860fc30df23cc67d


pkg/definitions/task_def_update.go
```diff
@@ -82,6 +82,52 @@ func (d *Definition) Update(req api.UpdateTaskRequest, opts UpdateOptions) error
...
+		if *req.RequireExplicitPermissions == false {
+			d.Permissions = nil
+		} else {
+			if d.Permissions == nil {
+				d.Permissions = &PermissionsDefinition{}
+			}
+			d.Permissions.RequireExplicitPermissions = true
+		}
+	}
+	if d.Permissions == nil && req.PermissionsSource != nil && *req.PermissionsSource == api.PermissionsSourceCode {
```
leeweisberger, line 126: Here's the other place we're using the permissions source. If the permissions source is set to code, that means, for the task definition, set the permissions but don't set anything else

pkg/deploy/discover/discover_test.go
```diff
@@ -711,6 +711,9 @@ func TestDiscover(t *testing.T) {
 						},
 						Node:                  &definitions.NodeDefinition{},
 						DefaultRunPermissions: definitions.NewDefaultTaskViewersDefinition(api.DefaultRunPermissionTaskParticipants),
+						Permissions: &definitions.PermissionsDefinition{
```
leeweisberger, line 714: Test is unrelated to the rest of the changes, just added to add a test

pkg/server/tasks/tasks.go
```diff
@@ -205,9 +205,6 @@ func taskConfigToAPITask(
 	if task.ResourceRequests == nil {
 		task.ResourceRequests = libapi.ResourceRequests{}
 	}
-	if task.Permissions == nil {
-		task.Permissions = libapi.Permissions{}
-	}
```
leeweisberger, line 210: It doesn't seem like deleting this thing does anything?? We could add a feature flag, but I couldn't tell what this would do. idk if you have any other ideas @colinking

pkg/initcmd/task.go
```diff
@@ -108,6 +110,12 @@ func InitTask(ctx context.Context, req InitTaskRequest) (InitResponse, error) {
 			return InitResponse{}, err
 		}
 
+		if req.Flagger == nil || !req.Flagger.Bool(ctx, req.Logger, flagsiface.PermissionsAsCode) {
```
leeweisberger, line 113: This feature flag is needed to ensure that when initializing a task, we don't set the permissions

pkg/definitions/updaters/javascript_test.go
```diff
@@ -270,6 +270,33 @@ func TestUpdateJavaScriptTask(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "empty_explicit_permissions",
```
leeweisberger, line 274: This test should also be unrelated to the changes here... just thought it was a good test

pkg/api/client.go
```diff
@@ -142,7 +142,11 @@ func (t Task) AsUpdateTaskRequest() UpdateTaskRequest {
 	if req.DefaultRunPermissions == nil {
 		req.DefaultRunPermissions = (*DefaultRunPermissions)(pointers.String(string(DefaultRunPermissionTaskViewers)))
 	}
-
+	if len(t.Permissions) == 0 {
+		req.Permissions = &Permissions{}
+	} else {
+		req.Permissions = &t.Permissions
+	}
```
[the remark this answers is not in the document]
colinking: If `*PermissionsSource != "code"`, should `req.Permissions` be `nil`?

pkg/definitions/task_def.go
```diff
@@ -638,8 +638,9 @@ func (d Definition) GetTask(opts GetTaskOpts) (api.Task, error) {
 		Constraints: api.RunConstraints{
 			Labels: []api.AgentLabel{},
 		},
-		DefaultRunPermissions: api.DefaultRunPermissions(d.DefaultRunPermissions.Value()),
-		SDKVersion:            d.SDKVersion,
+		DefaultRunPermissions:      api.DefaultRunPermissions(d.DefaultRunPermissions.Value()),
+		SDKVersion:                 d.SDKVersion,
+		RequireExplicitPermissions: d.Permissions != nil && d.Permissions.RequireExplicitPermissions,
```
colinking, line 643: Is it not valid to set `requireExplicitPermissions: true` and not set `permissions`? Users with admin permissions would still have access.

pkg/definitions/task_def.go
```diff
@@ -677,6 +678,72 @@ func (d Definition) GetTask(opts GetTaskOpts) (api.Task, error) {
 	}
 	task.Triggers = triggers
 
+	if d.Permissions != nil {
+		if d.Permissions.RequireExplicitPermissions {
```
colinking, line 682: Nit, I'd switch this `if` case. Seems easier to read.
```suggestion
		if !d.Permissions.RequireExplicitPermissions {
		// If we don't require explicit permissions but permissions have been set. This means that
		// ...
```

pkg/definitions/task_def.go
```diff
@@ -677,6 +678,72 @@ func (d Definition) GetTask(opts GetTaskOpts) (api.Task, error) {
...
+				task.Permissions = append(task.Permissions, api.Permission{
+					RoleID:     api.RoleTaskAdmin,
+					SubGroupID: &p,
+				})
+			}
+		} else {
+			// If we don't require explicit permissions but permissions have been set. This means that
+			// the user wants to set team permissions and only allow permissions to be managed as code.
+			// We represent this in the task by setting the permissions source to "code".
+			task.PermissionsSource = (*api.PermissionsSource)(pointers.String(string(api.PermissionsSourceCode)))
```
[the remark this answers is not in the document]
colinking: Shouldn't `PermissionsSource` also get set to "code" whenever `RequireExplicitPermissions == true`?

pkg/definitions/task_def_test.go
```diff
@@ -2079,6 +2079,113 @@ func TestDefinitionToUpdateTaskRequest(t *testing.T) {
...
+					ConcurrencyKey:      &emptyStr,
+					ConcurrencyLimit:    pointers.Int64(1),
+				},
+				Timeout: 0,
+				Env:     api.EnvVars{},
+				Constraints: api.RunConstraints{
+					Labels: []api.AgentLabel{},
+				},
+				RequireExplicitPermissions: pointers.Bool(true),
+				DefaultRunPermissions:      (*api.DefaultRunPermissions)(pointers.String(string(api.DefaultRunPermissionTaskViewers))),
```
colinking, line 2119: Nit: could we make a generic version of `pointers` that works for these enums? e.g. https://go.dev/play/p/Xsy1HmZb490

pkg/definitions/task_def.go
```diff
@@ -677,6 +678,72 @@ func (d Definition) GetTask(opts GetTaskOpts) (api.Task, error) {
 	}
 	task.Triggers = triggers
 
+	if d.Permissions != nil {
+		if d.Permissions.RequireExplicitPermissions {
+			for _, p := range d.Permissions.Viewers.Users {
+				p := p
+				task.Permissions = append(task.Permissions, api.Permission{
+					RoleID:    api.RoleTaskViewer,
+					SubUserID: &p,
```
colinking, line 687: Aren't we providing email addresses and group slugs? Not IDs?

Assuming we add logic for resolving emails/group slugs, don't add the API logic into the `definitions` package. Take a look at what we do for resources, where we have to resolve IDs to slugs and vice versa. Let the callsite decide how to do the lookups, as this behavior will vary between e.g. the builder and the studio.

pkg/print/types.go
```diff
@@ -30,6 +30,7 @@ type printTask struct {
 	Repo                       string                       `json:"repo" yaml:"repo"`
 	RequireExplicitPermissions bool                         `json:"requireExplicitPermissions" yaml:"-"`
 	Permissions                libapi.Permissions           `json:"permissions" yaml:"-"`
+	PermissionsSource          *libapi.PermissionsSource    `json:"permissionsSource" yaml:"-"`
```
colinking, line 33: This struct is used for printing out tasks, e.g. if you do `airplane tasks get`. You can change how the task is printed, either JSON or YAML. All fields should have both json and yaml struct tags set. Can you add them to all three permission fields here and confirm that these fields are printed correctly via `airplane tasks get`?

pkg/server/tasks/tasks.go
colinking: The web client does some sanitization which IIRC sets `permissions: {}`. So this is fine.

colinking (changes requested):
Requesting changes to discuss the email/slug <> ID issue.

pkg/definitions/updaters/javascript_test.go
colinking: Can you also add tests where real permissions are set? Same with the Python and YAML updaters.

leeweisberger: This is already tested in the "all" test

pkg/definitions/task_def.go
leeweisberger: It is valid.

`d.Permissions` is a pointer so we have to check if it's null before we check for requireExplicitPermissions. For the case you outlined, d.Permissions will be non-nil, d.Permissions.RequireExplicitPermissions will be true, and d.Permissions will contain no additional permissions.

Does this make sense?

pkg/definitions/task_def.go
leeweisberger: Yeah man there's a lot to do! Missed this in testing because I validated that the field was set, but it wasn't set correctly. Updated and manualyl tested this (plus added unit tests for the conversion)

leeweisberger:
GH actions is having issues, but I made changes, wrote unit tests and manually tested

pkg/cli/initcmd/task.go
```diff
@@ -112,14 +109,31 @@ func InitTask(ctx context.Context, req InitTaskRequest) (InitResponse, error) {
...
 			task.Permissions = nil
 			task.RequireExplicitPermissions = false
 			task.PermissionsSource = nil
 		}
 
-		def, err = definitions.NewDefinitionFromTask(task, resp.Resources)
+		def, err = definitions.NewDefinitionFromTask(task, definitions.NewDefinitionOptions{
+			AvailableResources: resp.Resources,
+			Users:              users,
+			Groups:             groupsResp.Groups,
```
colinking, line 135: These are only necessary if the task has explicit permissions set, correct? Maybe this should be a function instead so that we only do the HTTP requests if necessary?

```go
definitions.NewDefinitionOptions{
// ...
LookupUsersGroups: func() ([]Users, []Groups, error) {
		usersResp, err := client.ListTeamUsers(ctx, client.TeamID())
		if err != nil {
			return nil, nil, err
		}
		var users []libapi.User
		for _, user := range usersResp.Users {
			users = append(users, user.User)
		}

		groupsResp, err := client.ListGroups(ctx)
		if err != nil {
			return nil, nil, err
		}
		return users, groupsResp.Groups, nil
}
}
```

pkg/definitions/task_def.go
```diff
@@ -680,73 +690,101 @@ func (d Definition) GetTask(opts GetTaskOpts) (api.Task, error) {
...
-			task.PermissionsSource = (*api.PermissionsSource)(pointers.String(string(api.PermissionsSourceCode)))
 		}
+		// Permissions have been defined which indicates that the user wants to manage permissions as code.
+		task.PermissionsSource = (*api.PermissionsSource)(pointers.String(string(api.PermissionsSourceCode)))
 	}
 
 	return task, nil
 }
 
+func (d Definition) addPermissionsToTask(task *api.Task, opts GetTaskOpts) error {
```
colinking, line 704: Nice

colinking (approved):

pkg/cli/initcmd/task.go
colinking: FWIW, in retrospect I would have done the same for AvailableResources.

leeweisberger: I did this for all of them, both in the init flow and the update flow.

If there are no explicit permissions, don't fetch users and groups
If there are no resources, don't fetch resources

PR #1136, opened 2023-05-10
Update lint to match airport

colinking:
## Description

This PR updates CLI's lint setup to match airport and fixes the issues that I encountered. This avoids lint issues when merging the repos.

It also configures the required CI steps to run on merge queue, which we enabled on airport.

## Test plan

Tests pass.


cmd/cli/main.go
```diff
@@ -50,7 +50,8 @@ func main() {
 
 		logger.Debug("Error: %+v", err)
 		logger.Log("")
-		if exerr, ok := errors.Cause(err).(utils.ErrorExplained); ok {
+		var exerr utils.ErrorExplained
+		if errors.As(err, &exerr) {
```
colinking, line 54: I converted the `errors.Cause` usage to the standard library's `errors.As`. I think we should do the same in airport, but this at least avoids the lint issues. Since there were only a handful, it was simpler to do this conversion then to port over `errorsx` from airport.

cmd/cli/apikeys/create/create.go
```diff
@@ -8,7 +8,7 @@ import (
 	"time"
 
 	"github.com/airplanedev/cli/pkg/cli"
-	"github.com/airplanedev/cli/pkg/cli/apiclient/cliapi"
+	api "github.com/airplanedev/cli/pkg/cli/apiclient/cliapi"
```
yolken-airplane, line 11: Oh, so the package name in the file doesn't match the path? That's kind of annoying. Out-of-scope here, but can we eventually fix that?

yolken-airplane (approved):

cmd/cli/apikeys/create/create.go
colinking: Yep, I figure we can do that as part of cleaning up the API client. There are going to be three implementations in airport for a short bit 😓

yolken-airplane: Haha, ok.

PR #1139, opened 2023-05-10
Use string raw in vite config

leeweisberger:
## Description

We're seeing the following error for a windows user
```
[debug] starting Vite server on port 5173
[debug] Started vite with process id: 4636
⣻[debug] X [ERROR] Syntax error "i"
[debug]
[debug]     vite.config.ts:22:20:
[debug]       22 │         "C:\Users\usuf\projects\company\backend",
[debug]          ╵                     ^
[debug]
[debug] failed to load config from C:\Users\usuf\projects\company\backend\.airplane-view\vite.config.ts
[debug] error when starting dev server:
```

It appears that if `\u` is in a string, javascript tries to interpret it as unicode which fails.

Add `String.raw` to interpret this as a raw string.

## Test plan
Manually dev'ed a view and confirmed a) it works and b) it adds `String.raw` to the vite config
image (1)


kellymtrinh (approved):
REALLY nice find!

PR #1142, opened 2023-05-11
Refactor template cloning for slug uniqueness

francescaleung:
## Description

Instead of cloning directly in the working directory, we clone in a temp directory instead. If a discoverer is provided we run discovery on the temp directory and compare it to any existing task/view slugs that were passed in, rewriting files as necessary. The directory that gets added to the working directory will also get renamed if necessary.

Notable change in CLI flow: we now clone the directory, install dependencies, _then_ ask the user if they want to overwrite any collisions.

Supports ASM-304.

## Test plan

The whole template cloning flow is woefully undertested. Manually tested for now. Plans to add unit tests for `utils.Sed`, but if you have ideas for better testing around template cloning I am all ears. Separately, I need to also ensure that all of our templates will work with the sed approach--I mostly don't anticipate this being problematic, but I'll want to double-check.


francescaleung:
Migrated to https://github.com/airplanedev/airport/pull/4660

PR #1145, opened 2023-05-11
Fix incorrect resources check

leeweisberger:
## Description
Fix two bugs:
1. We are doing an incorrect check to see if the task has resources. We need to check the kind-specific fields, e.g. sql.resource. We could do a better check, but I removed this optimization for now. We should add a test to cover this.
2. We added a new field to task definition (sdkVersion) which is not available in the js sdk. When you update a js task from the UI, this field gets set in your code which causes an error.

## Test plan
Manually tested both changes. This fixes two bugs - one happens when you update a js task from the studio UI. The other happens when you dev on a task that has a resource - the task is continuously updated in a loop.


pkg/cli/server/tasks/tasks.go
```diff
@@ -202,13 +198,9 @@ func taskStateToAPITask(
 	taskState state.TaskState,
 	envSlug *string,
 ) (libapi.Task, error) {
-	var resourceMetadata []libapi.ResourceMetadata
-	if len(taskState.Def.Resources) > 0 {
```
colinking, line 206: This could be replaced with `taskState.Def.GetResourceAttachments()`. That handles built-in resources.

pkg/definitions/updaters/javascript/update.ts
```diff
@@ -163,6 +163,9 @@ const buildTaskConfig = (def: any): ExpressionKind => {
     }
   }
 
+  // We don't want to include the SDK version in the definition.
+  delete def.sdkVersion;
```
colinking, line 167: This is the second time (after concurrency keys) where a change to `Task -> Definition` logic has caused issues here, but hasn't been detected. We might want to add some logic for discovering a task (to get something real) and then throwing that through the updater?

colinking (approved):

